PR #49, opened 2015-02-14
Avoid MSVC fseek()'s 2GiB barrier

t-mat:
First of all, this is not a critical.  Absolutely minor thing.

Since MSVC `fseek()`'s `SEEK_CUR` mode has 2GiB barrier, our `fseek()` calling in `program/lz4io.c/selectDecoder()` will fail for large (>2GiB) .lz4 file.

There are two solution

 1. Use `fread()` for skippable frame.
 2. Use `_fseeki64()` which has been supported since MSVC2005.

1. is obviously slow. But it has few benefits

 - As [Neil pointed out](https://code.google.com/p/lz4/issues/detail?id=155#c5), we can't use `fseek()` for stdin, etc. So using `fread()` is good fallback method for such devices.
 - Even in old MSVC, "`fread()` seeking" has no 2GiB barrier.

Since I'm not sure about demand of skippable frame for large lz4, it seems over engineering for me to use `fread()` as fallback method.

So I employ `_fseeki64()` for my patch. It just replace `fseek` with `_fseeki64` by macro. Second argument is automatically casted to `__int64` (signed 64bit integer). Other things are completely same.


Cyan4973:
Thanks Takayuki,
this seems to save the day for Visual.

I do however have a few questions : 
- What about Linux and non-Windows platforms ?

t-mat:
> - What about Linux and non-Windows platforms ? Is the normal fseek() free of such limitation ?

Modern POSIX compatible systems does not have such limitations for `fseek()`.

 - POSIX compatible [LFS](http://en.wikipedia.org/wiki/Large_file_support) systems has [`fseeko()`](http://pubs.opengroup.org/onlinepubs/009695399/functions/fseek.html).
 - `fseek()` calls `fseeko()` compatible routines internally.

```
// pseudo code
int internal_fseeko(FILE *stream, off_t offset, int whence);

int fseek(FILE *stream, long offset, int whence) {
    return internal_fseeko(stream, (off_t) offset, whence);
}

int fseeko(FILE *stream, off_t offset, int whence) {
    return internal_fseeko(stream, offset, whence);
}
```

 - In this case, if sizeof(long) >= 8 (64bit), there is no problem at all.
 - Even if sizeof(long) < 8 (32bit, etc), `fseek()`'s `offset` bit width is just a limitation for single `fseek()` call. It does not limit current offset positon of the file.
 - So the following code will skip 4GiB in POSIX LFS systems :

```
void test(FILE *stream) {
    const uint64_t n = 1ULL << 32; // 4Gi
    uint64_t i = 0;
    for(i = 0; i < n; ++i) {
        fseek(stream, 1, SEEK_CUR); // Forward 1 byte per call
    }
}
```


> How is handled stdout ? fseek is fine too ?

If stdout is seekable (eg. file), `fseek()` works fine. But if stdout is not seekable (eg. terminal), `fseek()` will fail.  In latter case, we should use `fwrite()` or something similar function.

Cyan4973:
Thanks Takayuki, that's clear

In such case, maybe we have to think about a dynamic fseek/fwrite selector, since lz4 used in pipe mode is a relatively common scenario.
`./generator | lz4 | ./consumer`
The most common target scenario is writing to stdout (with stdout not a file).

I guess the selector itself is relatively simple to do.
It's the "non-seekable" property which might be more difficult to test.
Or maybe fseek() function result (errorCode) might be enough to help ?

t-mat:
> It's the "non-seekable" property which might be more difficult to test.

Agreed.  Standard C library say nothing about "seekable" things.
I'm not sure about details, but POSIX `fstat()` seems to help us.


> Or maybe fseek() function result (errorCode) might be enough to help ?

I hope so.  But I'm afraid to continue to manipulate stream after error occured :

 1. Is stream still safe to use ?
 2. Is file pointer guaranteed to recover from error ?
 3. Does it point exactly same position before the error ?

I'm realized I don't know anything about these questions and this kind of situation. I'll investigate this, but perhaps it takes time.

So Yann, could we focus to this PR itself - MSVC's `fseek()` ?
I think many `fseek()` problems are deeply related to googlecode issue 155 and current lz4io, but this PR is relatively apart from these `fseek()` problems.
I'd like to continue discussing this problem as issue 155 or another new issue.

Cyan4973:
Hi Takayuki

Sorry, I know it's a bit long wait, but i haven't forgotten, 
it's just I'm completely overwhelmed these days.
Your topic is nonetheless my next priority, so I should take care of it within the next few days.

Regards

t-mat:
No problem at all ! :smiley:
Since there's no report about this issue, I will never rush to solve it.

I belive, in this project everything must be done on your pace.
So please use your time for your own problem first.

Cyan4973:
I think I indeed confused your patch with issue 155.
It's very simple and straightforward.

PR #52, opened 2015-03-02
Adding support for multiple file processing.  Fixes Google code issue 151.

KyleJHarper:
Continuation of my previous PR, which accidentally had all changes in master and the PR targeted master.  I'm not spectacular with git, but I think this PR is correct.  If I did something wrong or you simply dislike the feature branch, you can just copy the lz4cli.c file out and add it to your local, then push; it's the only file changed.

Added support for multiple input files to act more like other compressors. For example: gzip file1 file2 file3. You can now do: lz4 [args] -m file1 file2 file3. Fixes 151.


Cyan4973:
Thanks Kyle, your patch is correctly directed.

Cyan4973:
I'm surprised that this line of code works : 

`char *input_filenames[argc];`

argc is not a constant, so it looks like a kind of "dynamic size allocation on stack" .
I suspect that Visual will not like it.
Let's find out....

Cyan4973:
Quick question about error case behavior : 

Let's suppose we want to compress a list of 6 filenames.
In the middle of the list, say name 3, compression fails (typically because the file does not exist or is not valid).
How should the program behave ?
=> Exit immediatly with an error code ?
=> Proceed with the rest of list, trying to compress as much valid files as possible ?

t-mat:
@Cyan4973, it's C99 feature "VLA" (Variable Length Array). Since our `CFLAGS` has `-std=c99`, it's strictly valid code. But VLA has some problems :

 - As you suppose, VC++ and old C89 compilers could not compile it.
 - VLA becomes optional feature in C11 and greater (see [`__STDC_NO_VLA__`](http://en.wikipedia.org/wiki/C11_%28C_standard_revision%29#Optional_features)). :anguished:

If we will support C89 compilers, [intermingled declarations](https://github.com/KyleJHarper/lz4/commit/fa27d236b4f3fc9938478e4464b63953d6429d4b#diff-638ed824ff46570e8635f871191331edR291) also should be removed.

> How should the program behave ?

Since this PR's intention is mimicing gzip, gzip-like behaviour will be expected. In this case, latter is more "gzip-like" behaviour. The following log is my experiment :

```
# Initial state
$ ls -1
lz4.1
lz4c.1
lz4cat.1

# Normal compression
$ gzip -k lz4.1 lz4c.1 lz4cat.1
$ ls -1
lz4.1
lz4.1.gz
lz4c.1
lz4c.1.gz
lz4cat.1
lz4cat.1.gz

$ rm *.gz

# Compression with errors
$ gzip -k lz4.1 dummy1 lz4c.1 dummy2 lz4cat.1 dummy3
gzip: dummy1: No such file or directory
gzip: dummy2: No such file or directory
gzip: dummy3: No such file or directory
$ ls -1
lz4.1
lz4.1.gz
lz4c.1
lz4c.1.gz
lz4cat.1
lz4cat.1.gz
```

Cyan4973:
Thanks Takayuki
It's an excellent reference.
Let's make it behave this way.

I'll also update Kyle contribution to make it C89 compatible.

PR #59, opened 2015-03-15
Removed checking of CMAKE_SYSTEM_PROCESSOR when adding -fPIC, breaks when that var is '64bit'.

Cyberunner23:


Cyan4973:
ok, but what's the point of removing the check ?
-fPIC should always be used, in all circumstances ?

Cyberunner23:
I don't see an issue with using -fPIC all the time except that maybe the compiler can do a little more optimisation without it.

Cyan4973:
OK
The only problem is that I don't accept commit directly into master, they must first go through "dev" or their own feature branch.
But don't worry, I've created #60 to redirect your patch.

PR #60, opened 2015-03-16
Removed checking of CMAKE_SYSTEM_PROCESSOR when adding -fPIC, breaks whe...

Cyan4973:
...n that var is '64bit'.


Cyan4973:
Redirect #59 to "dev" branch

PR #71, opened 2015-04-01
Cygwin has fileno, so there's no need to use _fileno.

fzort:
I had to change _fileno() to fileno() to build lz4 on cygwin. Please review.

### DIFF
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -65,7 +65,7 @@
 /****************************
 *  OS-specific Includes
 *****************************/
-#if defined(MSDOS) || defined(OS2) || defined(WIN32) || defined(_WIN32) || defined(__CYGWIN__)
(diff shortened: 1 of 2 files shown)


Cyan4973:
Thanks Mauro.
The code modification looks simple enough.

I won't merge directly into master, it goes against branch management policy.
Alternatively, I can make these small modifications directly by hand into current active branch (FastMode).

Another minor detail is that it would be better if cygwin tests could be automatically performed on a CI environment. This would avoid such problems to reproduce in the future.
Currently, we have gcc, g++ & clang, thanks to Travis, and then Visual on Windows, thanks to Appveyor. Cygwin environment unfortunately is not in the list. If you have any idea how this could be achieved, you are welcomed.

Cyan4973:
Superceded by #72

PR #75, opened 2015-04-05
Fixed a decompression bug in lz4io.c

funcodeio:
- If an lz4 file does not have end mark, lz4c didn't decompress everything. But
  it said "Successfully decoded".
- Even if we consume all input data, we need to call LZ4F_decompress more times
  to pull all decoded data within the lib because dest buffer size might be
  smaller than decoded data.

### DIFF
--- a/programs/lz4io.c
+++ b/programs/lz4io.c
@@ -624,13 +624,14 @@ static unsigned long long decodeLZ4S(FILE* finput, FILE* foutput)
         readSize = fread(inBuff, 1, inBuffSize, finput);
         if (!readSize) break;   /* empty file or stream */
 
-        while (pos < readSize)
+        while (1)
         {
             /* Decode Input (at least partially) */
             size_t remaining = readSize - pos;
             size_t decodedBytes = outBuffSize;
             errorCode = LZ4F_decompress(ctx, outBuff, &decodedBytes, (char*)inBuff+pos, &remaining, NULL);
             if (LZ4F_isError(errorCode)) EXM_THROW(66, "Decompression error : %s", LZ4F_getErrorName(errorCode));
+            if (decodedBytes == 0 && remaining == 0) break;
             pos += remaining;
 
             if (decodedBytes)


funcodeio:
I sent a python script reproducing this bug to Mr. Collet' email.
Please refer to the email. 
Thanks!

t-mat:
Hi, I think @funcodeio found out good point.

It's lz4io's defect but it's origin is lz4f.

Terminator function such as `LZ4F_freeDecompressionContext()` should check its own state (`LZ4F_dctx_internal_t.dStage`), and if it does not indicate proper state to exit, it should return error code.
For example, zlib's [`deflateEnd()`](https://github.com/madler/zlib/blob/50893291621658f355bc5b4d450a8d06a563053d/deflate.c#L979) has error checking code.

Also, second point suggests "explicit flush" function for decompression context. Since framing format specification said "End of compressed stream is implicit", if users know end of stream, explicit flush function might be useful.

Cyan4973:
Thanks for notifying Yongwoon

And yes, as usual you are right Takayuki.
The problem is within lz4frame, and shall be best fixed there.

funcodeio:
Yes, it would be better if we have a function to check context status in lz4frame.h. For flush, I think we don't need to introduce "explicit flush" because LZ4F_decompress with zero input has the same effect as flush function.

Here is the python code that you can reproduce the bug I pointed.
http://pastebin.com/2tBe5GY9
Note that
 - The lz4c binary shipped with Ubuntu 14.04 does not have this problem.
 - The latest version of lz4c has this problem.

Cyan4973:
I've been studying this issue in more detail, and I no longer believes it is related to lz4frame.

lz4frame seems to behaves as expected, 
that is, it clearly states that the frame is unfinished.
This is achieved by the return value of LZ4F_decompress(), which is not zero
(Specifically, for the test case of @funcodeio, it is 4).

The issue did not happened before r128 because up to r127, there was enough space to decode a full block in a single block. No longer, with r128 I introduced a new method within LZ4IO, which basically expects lz4frame library to do all the job. So it just provides input by segments of 64KB, and flush outputs by segments of 64 KB. This is much less than the default block size, which is 4 MB.

The method itself could be reverted back to previously, that is, provide a buffer size large enough to accommodate a full block. But that would be like hiding the issue.

In latest "dev" version (cbcdd88ccb97632015cf3732b46f8800e62e337b),
LZ4IO has been updated.
It now continues flushing data to output as long as it receives some,
in contrast with previous version, which would flush only once per pass.

It also controls proper frame termination signal, and output an error code when a frame is unfinished, such as in your example, where both the EndMark and Checksum are missing.

I believe it properly covers @funcodeio use case.
My only minor issue is that I've no been able to integrate the test case into the Continuous Integration test suite.

Regards

Cyan4973:
In answer to @t-mat : 

> Terminator function such as LZ4F_freeDecompressionContext() should check its own state (LZ4F_dctx_internal_t.dStage), and if it does not indicate proper state for termination, it should return error code.

There is a need to be able to free `decompressionContext` memory, even when its internal state is not "properly finished". After all, this is the case after any kind of error.

An error code would be too strong, as it would imply that the `free` operation has failed, which is not the case.

Eventually, an "information" message could prove helpful. `free` would still succeed, but the caller would have a chance to check that its decompression context was properly finished.

Note that the proper way to check that a frame is properly terminated is to look for `LZ4F_decompress()` result and verify if it is `== 0`.

Regards

Cyan4973:
I've updated LZ4F_freeDecompressionContext() within latest version in the "dev" branch 72e679438f9ead31e671e0738b7730f846ebde70, 
to provide an hint of current stage when decompression context is getting released.

That is, function result should be equal to zero if decompression has been completed fully and correctly.

t-mat:
> An error code would be too strong,

I see.  And I feel new information/hint code is also useful even it just indicates "something wrong."  I'll try to write new idiomatic LZ4F (de)compressor later.

PR #82, opened 2015-04-12
Add LZ4 prefix to deprecation macros

t-mat:
Trivial fix. Since `DEPRECATED` is too generic name to public header, this PR replace the following macros:

 - `_DEPRECATE_WARNING_DEFBLOCK` to `LZ4_DEPRECATE_WARNING_DEFBLOCK`
 - `DEPRECATED` to `LZ4_DEPRECATED`


Cyan4973:
Good point !

PR #86, opened 2015-04-12
Origin/r129/multiple inputs patch

KyleJHarper:
Hello,

Pardon my mix-up if I've made this PR backward or something silly.  I'm not an expert in C either so please forgive oddities.

Google code issue 151 resulted in the addition of the -m switch which allows the lz4 cli compressor to act like others (gzip/bzip2/xz).  However, it does not work when decompressing with -d.  I made a sub-branch of an r129 branch to create a function which will perform the decompression of multiple files (in short:  ./lz4 -m -d file1.lz4 file2.lz4 now works).  This is the first commit.

The second change was to support missing input files.  When a compressor (gzip et al) encounters a missing input file specified on command line it simply prints a warning to STDERR and continues operating on the remaining files.  My second commit accomplishes this by attempting an fopen() on each input file when -m is specified.  This leaves other file opening/reading/writing errors to the actual compression function (which is fatal, and probably should be).  fopen() may not be ideal, but it's what I knew.  Oh, and missing_files is now tracked and will change the return code to '1' when missing inputs are discovered, which also mimics other compressors' default behavior.

If you need me to repackage these as different branches or whatever just let me know.  Or if you want you can just rip the three files out and make commits on your account; I'm not really concerned with credit or whatever.

Thanks!
Kyle

### DIFF
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -265,7 +265,8 @@ int main(int argc, char** argv)
         forceStdout=0,
         forceCompress=0,
         main_pause=0,
-        multiple_inputs=0;
+        multiple_inputs=0,
+        multiple_rv=0;
     const char* input_filename=0;
     const char* output_filename=0;
     char* dynNameSpace=0;
@@ -518,7 +519,7 @@ int main(int argc, char** argv)
     if (decode)
     {
       if (multiple_inputs)
-        LZ4IO_decompressMultipleFilenames(inFileNames, ifnIdx, LZ4_EXTENSION);
+        multiple_rv = LZ4IO_decompressMultipleFilenames(inFileNames, ifnIdx, LZ4_EXTENSION);
       else
         DEFAULT_DECOMPRESSOR(input_filename, output_filename);
     }
@@ -533,7 +534,7 @@ int main(int argc, char** argv)
       else
       {
         if (multiple_inputs)
-          LZ4IO_compressMultipleFilenames(inFileNames, ifnIdx, LZ4_EXTENSION, cLevel);
+          multiple_rv = LZ4IO_compressMultipleFilenames(inFileNames, ifnIdx, LZ4_EXTENSION, cLevel);
         else
           DEFAULT_COMPRESSOR(input_filename, output_filename, cLevel);
       }
@@ -542,5 +543,6 @@ int main(int argc, char** argv)
     if (main_pause) waitEnter();
     free(dynNameSpace);
     free((void*)inFileNames);
+    if (multiple_rv != 0) return multiple_rv;
     return 0;
 }
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -515,22 +515,28 @@ int main(int argc, char** argv)
 
     /* IO Stream/File */
     LZ4IO_setNotificationLevel(displayLevel);
-    if (decode) DEFAULT_DECOMPRESSOR(input_filename, output_filename);
+    if (decode)
+    {
+      if (multiple_inputs)
+        LZ4IO_decompressMultipleFilenames(inFileNames, ifnIdx, LZ4_EXTENSION);
+      else
+        DEFAULT_DECOMPRESSOR(input_filename, output_filename);
+    }
     else
     {
-        /* compression is default action */
-        if (legacy_format)
-        {
-            DISPLAYLEVEL(3, "! Generating compressed LZ4 using Legacy format (deprecated) ! \n");
-            LZ4IO_compressFilename_Legacy(input_filename, output_filename, cLevel);
-        }
+      /* compression is default action */
+      if (legacy_format)
+      {
+        DISPLAYLEVEL(3, "! Generating compressed LZ4 using Legacy format (deprecated) ! \n");
+        LZ4IO_compressFilename_Legacy(input_filename, output_filename, cLevel);
+      }
+      else
+      {
+        if (multiple_inputs)
+          LZ4IO_compressMultipleFilenames(inFileNames, ifnIdx, LZ4_EXTENSION, cLevel);
         else
-        {
-            if (multiple_inputs)
-                LZ4IO_compressMultipleFilenames(inFileNames, ifnIdx, LZ4_EXTENSION, cLevel);
-            else
-                DEFAULT_COMPRESSOR(input_filename, output_filename, cLevel);
-        }
+          DEFAULT_COMPRESSOR(input_filename, output_filename, cLevel);
+      }
     }
 
     if (main_pause) waitEnter();
--- a/programs/lz4io.c
+++ b/programs/lz4io.c
@@ -501,12 +501,21 @@ int LZ4IO_compressFilename(const char* input_filename, const char* output_filena
 int LZ4IO_compressMultipleFilenames(const char** inFileNamesTable, int ifntSize, const char* suffix, int compressionlevel)
 {
     int i;
+    int missing_files = 0;
+    FILE *ifp;
     char* outFileName = (char*)malloc(FNSPACE);
     size_t ofnSize = FNSPACE;
     const size_t suffixSize = strlen(suffix);
 
     for (i=0; i<ifntSize; i++)
     {
+        ifp = fopen(inFileNamesTable[i], "r");
+        if (ifp == NULL) {
+            DISPLAYLEVEL(2, "Unable to access file for processing: %s\n", inFileNamesTable[i]);
+            missing_files++;
+            continue;
+        }
+        fclose(ifp);
         size_t ifnSize = strlen(inFileNamesTable[i]);
         if (ofnSize <= ifnSize+suffixSize+1) { free(outFileName); ofnSize = ifnSize + 20; outFileName = (char*)malloc(ofnSize); }
         strcpy(outFileName, inFileNamesTable[i]);
@@ -514,6 +523,7 @@ int LZ4IO_compressMultipleFilenames(const char** inFileNamesTable, int ifntSize,
         LZ4IO_compressFilename(inFileNamesTable[i], outFileName, compressionlevel);
     }
     free(outFileName);
+    if (missing_files > 0) return 1;
     return 0;
 }
 
@@ -521,6 +531,8 @@ int LZ4IO_decompressMultipleFilenames(const char** inFileNamesTable, int ifntSiz
 {
     int i;
     int skipped_files = 0;
+    int missing_files = 0;
+    FILE *ifp;
     char* outFileName = (char*)malloc(FNSPACE);
     size_t ofnSize = FNSPACE;
     size_t suffixSize = strlen(suffix);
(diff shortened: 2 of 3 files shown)


Cyan4973:
Thanks Kyle.
The proposed commits look great, in line with the rest of the code.

There are a couple a minor portability issues to handle, but nothing important. I'll take care of that.

I'll have a closer look at the proposed commit and probably merge it later today.

KyleJHarper:
Sounds good.

By the way, I did some testing with ~575,000 small files (8k) for a project I'm working on.  I'm testing compressor performance on a single thread (1 core VM using Intel i7 cpu):
find /tmp/ram_drive -type f | xargs -r $compressor $switches
for lz4 (r129 patch) it is:
find /tmp/ram_drive -type f | xargs -r lz4 -m [-d]

The performance is poor compared to say lzop or gzip, particularly for decompression where lz4 usually shines.  This has, I believe, nothing to do with the lz4 compress/decompress logic.  We seem to be suffering performance degradation in either the LZ4IO_[de]compressMultipleFilenames() functions or in the construction/invocation of the subsequent LZ4IO_[de]compressFilename() functions.  The performance hit is microscopic on a per-file basis: for ~575,000 files it added 60 seconds of cpu time.  However, the whole process only takes ~22 seconds with lzop cli so clearly there's room for improvement.

Possibilities:
1.  I introduced this performance hit with the fopen() check to continue when missing files are encountered.  I considered stat() and access() but wasn't sure they were any faster/better.  All file-checking options are going to have TOCTOU issues so if another function is faster we could switch.
2.  I introduced this performance hit with memcpy() in LZ4IO_decompressMultipleFilenames.  A proper use of strcpy/strncpy() might help avoid this.  My entire approach to finding and comparing the files' suffixes to the LZ4_EXTENSION macro (passed as const char* suffix by the caller) might need adjusted.
3.  The start-up cost of the LZ4IO_[de]compressFilename() functions might be showing up now that we have the ability to process multiple files (-m).  Previous performance testing would have masked this because the cost of spawning a new system process massively outweighs the minor performance hit I'm observing.

Let me know what you think.  I can provide you with the sample data too (it's not sensitive data).  I'm also happy to run the tests for you since I have an existing environment/VM already.

Thanks,
Kyle

Cyan4973:
I suspect it might be related to default block size.

Try `./lz4 -B4D` to select 64 KB linked blocks during compression stage
(or `B4` only if all your files are <= 64 KB),
decompression will automatically adapt.

KyleJHarper:
Hey Yann,

I used -B4 and it helped, but there's still a big difference between lz4cli and other compressors.  I fed a little over 750,000 files (all 8k each) to xargs with the following lz4 and lzop invocations and got the following times.  I also wrote a perl script using Compress::LZ4 to compare against.

(compression)
lz4 -1 -m  == 54s
lz4 -1 -B4 -m  == 40s
./lz4_c.pl == 34s
lzop -1  == 33s
(decompression)
lz4 -1 -m -d  == 114s     !!!
lz4 -1 -B4 -m -d  == 63s   !!!
./lz4_d.pl == 26s
lzop -1 -d  == 29s

My perl implementation is more in line with what I expected vs lzop (based on other tests I've done).  I'm not sure what to test, but if you have ideas let me know.  Maybe I should comment out the extension checking in the decompressMulitple function... as decompression seems to suffer more drastically than compression in my example above.

Thanks,
Kyle

Cyan4973:
These are very good tests.
Thanks for the reference.

Your perl script looks good too.

I think I've got a few ideas of what could cause this slowdown.
I'm going to create a new version that should solve some or most of these problems.
I'll propose it to you for testing later tomorrow.

Regards

Cyan4973:
I uploaded a new version of lz4io
within Dev branch : c64200dd85a85d9cc69a6fda682362d851be863b

It's targeting
improved performance when compressing lot of small files.

I'm not able to test it, so it's pure guess for now.
The way it works is by keeping resources shared between files.
It also reduces the amount of `fopen()` to 1 per file.
(Not sure if it is significant, since I don't know what's the amount of work behind `fopen()`,
but as it trigger some interaction with file system, I figured it's nonetheless better to reduce that).

It would be better if I could test something, even if it is not as good as yours.
My current development VM is very small (~10 GB) though,
so I can't create such a large corpus of small files for tests.

Decompression side is not done yet.
It's quite a bit more complex, but I'll take care of it next.

t-mat:
Yann, I've tested c64200d, but `LZ4IO_compressMultipleFilenames()` is not work properly. Number of compression files are limited to `ulimit -n`/2 files. I think you may miss `fclose()`s for `ress` in the loop.

```
int LZ4IO_compressMultipleFilenames(const char** inFileNamesTable, int ifntSize, const char* suffix, int compressionLevel)
{
    ...
    /* loop on each file */
    for (i=0; i<ifntSize; i++)
    {
        ...
        issueWithSrcFile = LZ4IO_getFiles(inFileNamesTable[i], outFileName, &(ress.sFile), &(ress.dFile));
        missing_files += issueWithSrcFile;

        if (!issueWithSrcFile) LZ4IO_compressJob(ress);
/* + */ if (ress.sFile) { fclose(ress.sFile); ress.sFile = NULL; }
/* + */ if (ress.dFile) { fclose(ress.dFile); ress.dFile = NULL; }
    }

    /* Close & Free */
    free(ress.srcBuffer);
    free(ress.dstBuffer);
//- fclose(ress.sFile); // remove
//- fclose(ress.dFile); // remove
```

Cyan4973:
Very good point, thanks Takayuki !
I'll correct that for next update.

Side-comment : I'm surprised this issue evaded the surveillance of static analyzer (scan-build)

KyleJHarper:
For what it's worth:  I'm happy to provide the performance testing as the code is updated, it's trivial to do.  I can also provide the source (it's just a postgres DB with all the internal pages broken out into 8k individual files).  I make a 6.5GB tmpfs directory to help alleviate disk IO skewing results.

If you have the RAM and resources and you'd like to create a performance test case for regression in your build env (Jenkins/Travis) I can help provide the raw data and testing script I've been using.

t-mat:
I've tested 8KiBytes * 30Ki Files @ Ramdisk.  It seems good.  Result is

|program                |seconds|time factor	|
|----                   |----   |----		|
|lzop 1.03		|4.273s	|1.00		|
|lz4 old (da11725)	|4.962s	|1.16		|
|lz4 dev		|3.946s	|0.92		|

 - Log at https://gist.github.com/t-mat/2d0dcfd454272fe135c2

> I'm surprised this issue evaded the surveillance of static analyzer (scan-build)

Hmm, maybe valgrind could found this kind of error.

Cyan4973:
Indeed, thanks Takayuki.
Valgrind test does indeed catch the leak issue.

The only minor limitation is that
valgrind properly "report" the issue,
but doesn't exit with an error code.

That means it's necessary to read the test results
and actively look for potential issue.
This is a weaker signal than an error code,
which would automatically show up on Travis.

Cyan4973:
@KyleJHarper Thanks for the proposal.

The issue is that I've got a very small VM to work with, and I can't load such a large corpus into it.

Integrating into Travis looks initially desirable,
but Travis is not a correct environment for benchmarking.
You never know the system specs nor the parallel load on the same hardware.
So bench results are not indicative on this environment.

I guess I'll have to rely on your measurements ....

t-mat:
> valgrind properly "report" the issue, but doesn't exit with an error code.

The following Q&A resolves this problem:

 - [How to get in script whether valgrind found memory leaks?](http://stackoverflow.com/questions/19246470/)

Valgrind also offers [`--track-fds` option](http://valgrind.org/docs/manual/manual-core.html#opt.track-fds), but it's [just a bit tricky](http://stackoverflow.com/questions/13279749/).

Cyan4973:
Thanks Takayuki

The `--error-exitcode=<number>` solution proposed in your link worked perfectly,
for genuine memory leaks.
That is, when a `free()` is missing and pointer is lost.

Unfortunately, for the specific issue of this thread,
lost file descriptor do not count as "genuine leak",
presumably because there is still a pointer within the file descriptor
pointing to the base memory address.
So valgrind counts them as "still reachable", which is not an error.

I see your other links, proposing `--track-fds` option.
But I'm not very fond of the requirement to manually close `stdin`, `stdout` & `stderr` to avoid false positive.

Maybe we'll just have to live with it....

Cyan4973:
minor note : 
I also suspect that, while having a "compression status message" at the end
when compressing a single file, or stream,

it makes much less sense when there bunch of files to compress.

So, in such case, I could implement the same "trick" as when `lz4` is used in pure pipe mode : 

automatically reduce "displayLevel" so that only error messages get printed.

Cyan4973:
Latest update in the "dev" branch a430b859cf552aa43215fe082e15a95c8bda5d6f
seems to solve also the issue regarding decompression speed.

I was slightly skeptical initially if the changes would prove enough,
but using @t-mat [benchmark methodology](https://gist.github.com/t-mat/2d0dcfd454272fe135c2) (using ramdisk)
the new lz4 version looks faster than tested alternatives.

The speed is also stable, no longer tied to block size (for both compression & decompression).

KyleJHarper:
Thanks Yann.  I'll try to pull a copy down tonight and test just for additional benchmark info.  I'm out for a bike ride for a few hours until then :)

KyleJHarper:
I built the dev branch lz4cli and observed the following using the same test from above:
(compression)
lz4 -1 -m == 54s (original test)
lz4 -1 -m == 33s (dev branch)
lzop -1 == 33s
(decompression)
lz4 -1 -m -d == 114s (original test)
lz4 -1 -m -d == 29s (dev branch)
lzop -1 -d == 29s

The changes you made have put lz4 CLI's file-handling overhead on par with other compressors.  Nicely done... now if only I had a clue what you actually did heh :)

That should close out this PR/issue I think.  The only other thing you mentioned was changing DISPLAYLEVEL to not send messages to stderr when processing multiple files.  I think that's a worthwhile effort but we can make that a new issue if you want.  Without it I have to send 2>/dev/null which hides valid errors.  (If I don't send it to a file or /dev/null the program actually becomes bottlenecked by my console write speed haha).

Cyan4973:
> now if only I had a clue what you actually did heh :)

A bunch of resources are required to compress/decompress a job.
Instead of creating/freeing them with each file, they are made common for all files.
This saves a bunch of `malloc()` or `free()`, and apparently, it's enough to make such a differences.

Since you now have identical results with `lz4` and alternatives,
I suspect the new limiting factor might be the I/O system itself, i.e. the storage.


> The only other thing you mentioned was changing DISPLAYLEVEL to not send messages to stderr when processing multiple files. 

Done : 13c6e16333a63319ef0def9da6595a3bc7d2915d

KyleJHarper:
Thanks for the explanation.  I figured it would come down to something like that.

The new DISPLAYLEVEL suppression is working for multiple files, but it also suppresses valid error messages like so:
kyle@hca:/tmp$ ./lz4 -d -m *.lz4
kyle@hca:/tmp$ echo $?
1
kyle@hca:/tmp$ ll
-rw-rw-r--  1 kyle kyle      0 Apr 16 09:54 1.txt
-rw-rw-r--  1 kyle kyle     15 Apr 16 09:54 1.txt.lz4
-rw-rw-r--  1 kyle kyle      0 Apr 16 09:54 2.txt
-rw-rw-r--  1 kyle kyle     15 Apr 16 09:54 2.txt.lz4
-rw-rw-r--  1 kyle kyle      0 Apr 16 09:54 3.txt
-rw-rw-r--  1 kyle kyle     15 Apr 16 09:54 3.txt.lz4
-rwxrwxr-x  1 kyle kyle 119747 Apr 16 09:53 lz4*

In the above example my wildcard pattern matched the lz4 binary itself (on purpose).  I received no output indicating an error occurred, I simply received a non-zero status code.  Flipping the display level value repeatedly might be overkill, but we might want to buffer display messages for failures in the multiple-files mode and flush them after setting display level higher at the end of the [de]compressMultiple...() functions.

Cyan4973:
Right.
It's probably more logical to classify these messages as "errors" instead of notifications, as they are today.
This way, they will be printed into `stderr`, even in multiple-files mode.

Cyan4973:
Done : 7644bee6435f05e0954b3da5f457550614a199b7

PR #88, opened 2015-04-13
Add more descriptive frame errors

drcrallen:
I was working on validating a different library's results and found this patch very helpful in pinpointing problems. Especially since not all implementations of lz4 support all features of lz4 frames.


Cyan4973:
Indeed, thanks Charles, that's a good idea.

Note that commits are never merged directly into master.
They must go through "dev" or their own feature branch first.

drcrallen:
thnks, closing to move to dev

PR #92, opened 2015-04-14
Revert "Add more descriptive frame errors"

Cyan4973:
Reverts Cyan4973/lz4#89


Cyan4973:
Some issues with import; will try again

PR #95, opened 2015-04-15
Modified lz4frame enum, by @t-mat

Cyan4973:


Cyan4973:
I'm wondering if using macro for old enum is good enough.
This question looks useful mostly in the context of DLL.

t-mat:
Here is experimental implementation which eliminate macro constants :

 - https://github.com/t-mat/lz4/blob/issue/90-exp/lib/lz4frame.h#L65

Perhaps, this notation is better for deprecation attributes.

Cyan4973:
Looks clean :)

Cyan4973:
Do you intend to make it a pull request ?

t-mat:
Yes. I'll send PR.

PR #98, opened 2015-04-23
cmake support for AIX, HPUX, Solaris and Windows

eberge:
The following changes allow for builds on AIX and HPUX
with the native (non-gcc) compilers, as well as
Visual Studio 2008 and Visual Studio 2012.

Also work around a build error with gcc on Solaris
which fails due to the system detecting an attempt
to use C99 mode with an XPG mode less than XPG6.


Cyan4973:
Looks great

PR #99, opened 2015-04-28
Install the lz4frame.h header in the cmake build

eberge:
I believe the intention here was to also install lz4frame.h in the cmake file similarly to the Makefile-based build.


Cyan4973:
Thanks Eric. Yes, it's the correct intention.

Issue #103, opened 2015-05-13
Add compatibility test between releases


t-mat:
Recently, [Joffrey LABORDE (implicitly) pointed out](https://groups.google.com/d/msg/lz4c/D_hSD19ykqU/DE2eecLIrYcJ), there is no clear evidence for interoperability between releases.  Also there is a similar FAQ : "Are 64bit (lz4c) and 32bit (lz4c32) interoperable ?"
The following Python3 script is initial answer for these questions :

 - https://gist.github.com/t-mat/3c5e1e4df3b250691caf

This automated test compress and decompress between

 - All git tagged releases
 - `lz4c` and `lz4c32`
 - `-1` and `-9`

I hope this test will guarantee basic backward and forward compatibility for all LZ4 releases.

### Usage

```sh
cd
mkdir lz4-test
cd lz4-test
curl -O https://gist.githubusercontent.com/t-mat/3c5e1e4df3b250691caf/raw/test-lz4-releases.py
chmod +x ./test-lz4-releases.py
./test-lz4-releases.py
```
[edit] : Add `chmod` 

### Details

 - Checkout all tagged release code (pattern `r[0-9][0-9][0-9]`)
    - Compiles `lz4c` and `lz4c32`
    - Rename these executables. Add tag as suffix to these executables (eg `lz4c32.r129`)
    - Copy all executables to grand parent directory
 - Copy `README.md` as test file (`test.dat`)
 - Compress `test.dat` by all executables
   - Use `lz4c.<TAGS>` and `lz4c32.<TAGS>`
   - Specify `-1zf` and `-9zf`
 - Remove dupilcated `.lz4` files
 - Decompress all `.lz4` files by all executables
   - Compare decompressed data with `test.dat`
   - Remove the file if there is no difference
   - Report the error if difference is found

### Remained Problems

 - More test data is needed for serious test.
 - Test must be running on TravisCI.
    - Also VC++ and appveyor ?
 - Dependency to python is not fit for LZ4 project.
 - Is `make cc=clang` needed ?

Cyan4973:
Hi Takayuki

Testing 32-bits <-> 64-bits compatibility is already part of the CI test suite. This is probably not clear enough if you have missed it.
[...]

PR #104, opened 2015-05-15
Add compatibility test between releases (resolves issue #103 )

t-mat:
This PR resolves issue #103.
It adds compatibility test between all released (git tagged) versions of LZ4.

 - Add Makefile target "[`versionstest`](https://github.com/Cyan4973/lz4/compare/dev...t-mat:issue/103pr2?expand=1#diff-b67911656ef5d18c4ae36cb6741b7965R117)".
   - Standard test procedure like "`make versionstest && make clean`" works properly.
 - Add `test` directory and its [Makefile](https://github.com/Cyan4973/lz4/compare/dev...t-mat:issue/103pr2?expand=1#diff-0d8f5a336dfc7f93d0f8ebb4026e4d46R1) and Python3 [script](https://github.com/Cyan4973/lz4/compare/dev...t-mat:issue/103pr2?expand=1#diff-90c37dff7749a6f2e2baa839336e439eR1).
 - Add Travis build matrix entry [`env: LZ4_TRAVIS_CI_ENV=versionstest`](https://github.com/Cyan4973/lz4/compare/dev...t-mat:issue/103pr2?expand=1#diff-354f30a63fb0907d4ad57269548329e3R22)

Here is an example TravisCI result :

 - https://travis-ci.org/t-mat/lz4/jobs/62636561

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ before_install:
   - sudo apt-get install -qq g++-multilib
   - sudo apt-get install -qq gcc-multilib
   - sudo apt-get install -qq valgrind
+  - sudo apt-get install -qq python3
 
 env:
   - LZ4_TRAVIS_CI_ENV=travis-install
@@ -18,6 +19,7 @@ env:
   - LZ4_TRAVIS_CI_ENV=staticAnalyze
   - LZ4_TRAVIS_CI_ENV=gpptest
   - LZ4_TRAVIS_CI_ENV=armtest
+  - LZ4_TRAVIS_CI_ENV=versionstest
   - LZ4_TRAVIS_CI_ENV=test-lz4
   - LZ4_TRAVIS_CI_ENV=test-lz4c
   - LZ4_TRAVIS_CI_ENV=test-lz4c32
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,7 @@ clean:
 	@cd $(PRGDIR); $(MAKE) clean > $(VOID)
 	@cd $(LZ4DIR); $(MAKE) clean > $(VOID)
 	@cd examples;  $(MAKE) clean > $(VOID)
+	@cd test;      $(MAKE) clean > $(VOID)
 	@echo Cleaning completed
 
 
@@ -112,6 +113,9 @@ staticAnalyze: clean
 armtest: clean
 	cd lib; $(MAKE) -e all CC=arm-linux-gnueabi-gcc CPPFLAGS="-Werror"
 	cd programs; $(MAKE) -e bins CC=arm-linux-gnueabi-gcc CPPFLAGS="-Werror"
+
+versionstest: clean
+	@cd test; $(MAKE) -e versionstest
 
 streaming-examples:
 	cd examples; $(MAKE) -e test
--- a/test/Makefile
+++ b/test/Makefile
@@ -1,0 +1,40 @@
+# ##########################################################################
+# LZ4 tests - Makefile
+# Copyright (C) Yann Collet 2011-2015
+#
+# GPL v2 License
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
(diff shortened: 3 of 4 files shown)


Cyan4973:
Thanks Takayuki, it's a great capability your just added.

A quick question on the test result as seen on [Travis](https://travis-ci.org/Cyan4973/lz4/jobs/62639212) :

All versions from r116 to r129 are being compiled. This part of the test is clear.

Afterwards, the output from the test is less obvious.
````
test_dat_1_32_r116.lz4
test_dat_1_32_r129.lz4
test_dat_1_64_r129.lz4
test_dat_9_32_r116.lz4
test_dat_9_32_r124.lz4
test_dat_9_32_r129.lz4
````

If I do understand out latest discussion, it means only 6 different files are generated.
For `-1` "fast" setting, version r116 32-bits remains unchanged up to r129.
For "fast 64 bits", only r129 is mentioned, do I have to understand that 64-bits is identical to 32-bits up to r129 ?

For `-9` "HC" setting, 3 versions too. I'll supposed that 32-bits & 64-bits versions are identical, which is normal. So there is a change at r124, which is indeed a drop in compression ratio that remained unnoticed. It was corrected on r128, but I don't see this version on the list. Maybe because it regenerated the same file as r116 ? Then there is r129, which is a bit more surprising.

Finally, seeing the generated file is not the same as testing the compatibility. I suspect it's done, it's just not displayed.

Anyway, these are just cosmetic changes. I can probably try to learn a bit of Python to apply such modifications directly.

**Comment by Cyan4973** [comment]
Made some pretty minor cosmetic changes to testVersions into c779c8004c890b61791fd8c96afa62dccd7ed507.

I think the behavior of the test is very good.

And indeed, having an immediate glance at which compressed file is different is a great helper.

**Comment by t-mat** [comment]
Sorry !  Yes, I've omitted useful log messages too much.  Because I've been so afraid Travis's stdout limit :sweat_smile:

> For "fast 64 bits", only r129 is mentioned, do I have to understand that 64-bits is identical to 32-bits up to r129 ?

Yes.  [This Travis log](https://travis-ci.org/t-mat/lz4/jobs/62713625#L336) shows how duplicate recognition works.  Since [.lz4 file names are sorted](https://github.com/Cyan4973/lz4/commit/e3b5bf33f9c6f0e1bc9ca9405e0b683c590e3ccd#diff-90c37dff7749a6f2e2baa839336e439eR87), it always keep oldest file and remove all newer duplicated files.

> It was corrected on r128, but I don't see this version on the list.
> Maybe because it regenerated the same file as r116 ?

Yes.  Output of lz4c.r128 (-1) and lz4c32.r128 (-1) and lz4c32.r116 (-1) are identical.

> Then there is r129, which is a bit more surprising.

Both output of lz4c32.r129 (-1) and lz4c.r129 (-1) are unique.

There are r116, r128, r129 related lines in Travis log :

```
# (output of) lz4c32.r116 -1 == lz4c32.r128 -1 == lz4c.r128 -1
test_dat_1_32_r116.lz4 == test_dat_1_32_r128.lz4
test_dat_1_32_r116.lz4 == test_dat_1_64_r128.lz4

# (output of) lz4c32.r129 -9 == lz4c.r129 -9
test_dat_9_32_r129.lz4 == test_dat_9_64_r129.lz4
```

---
Anyway, just an FYI, here is a small patch which outputs duplicated .lz4 files, OK/ERR message for decompression test, and SHA1 of unique .lz4 files.

 - https://github.com/t-mat/lz4/commit/311823f3effdd2ce6999f2a5f41f8aef12057627
 - https://github.com/t-mat/lz4/commit/581a87d5a3e77db3320370af21a333e17f9c7c08
 - Travis log : https://travis-ci.org/t-mat/lz4/jobs/62713625
    - `r999` means Travis's `head` (last commit from Travis's point of view).  If you will run `make versionstest` in your local system, `r999` means your current working copy.

Also, since `test/Makefile`'s `versionstest` target does not have dependency to `clean`, you could use this target when you use `versionstest` frequently.  This feature is useful for local editing.

```
cd /path/to/lz4
cat README.md
cd test
make versionstest
# edit test-lz4-versions.py or current working copy
make versionstest
```

Cyan4973:
I really like your latest log from Travis. 
It helps diagnosis.

I believe I fixed the minor issue in latest "dev" commit, so you can push your patch.

PR #105, opened 2015-05-18
Disabled sparse file mode when output file is terminal.

funcodeio:
After --sparse option is enabled by default,  the command,"lz4 -d -c <lz4 file>"  always crashes because the lz4 tries to fseek on a console file pointer (stdout).

I made --sparse mode be disabled when output file is console.  Just a simple change.

### DIFF
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -263,6 +263,7 @@ int main(int argc, char** argv)
         bench=0,
         legacy_format=0,
         forceStdout=0,
+        forceFileOut=0,
         forceCompress=0,
         main_pause=0,
         multiple_inputs=0,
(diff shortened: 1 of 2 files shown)


Cyan4973:
Thanks for notification and for the patch.

I would also like the relevant test to be added of the CI suite, to ensure reproduction and non-regression during future evolution.
I'll take care of it.

funcodeio:
I realized that just checking isatty is not enough.

Currently, if I run the following command, it crashes, which is not covered by this patch.
"./lz4 -d -c test.lz4 | cat"
We need to use fstat in this case.

There will be many other cases, and it will be platform dependent. 
It is a complicated problem than I initially thought.

Cyan4973:
> It is a complicated problem than I initially thought.

That's the conclusion most of the time :)
Hence there is some merit in properly solving even minor items.

Don't worry, we'll find a solution.
Worst case is, we can disable sparse support for stdout.
It would be a pity, but if there is nothing better...

Cyan4973:
Latest "dev" update is supposed to fix this issue : 60d657ac86ce2db86a5392d26452e9c63230f0a3
It relies on automatic fseek capability test.
Some new tests are now added in the test suite, including the more complex `| cat` you describe.

Cyan4973:
Fixed within "dev" branch

t-mat:
note : 60d657a in @Cyann4973's [comment](https://github.com/Cyan4973/lz4/pull/105#issuecomment-103220555) seems wrong hash.  It is 58b5aad.

PR #114, opened 2015-05-26
Djgpp

lpsantil:
Minimal changes to the lz4's Makefiles, lz4cli.c, and lz4io.c to allow it to compile with the build-djgpp cross compilers.  FYI, most significant change is switch from `-std=c99` to `-std=gnu99` in CFLAGS.


lib/Makefile
```diff
@@ -60,14 +60,20 @@ else
 	SHARED_EXT_VER = $(SHARED_EXT).$(LIBVER)
 endif
 
+.PHONY: default all liblz4 liblz4 liblz4-static liblz4 clean install uninstall
```
Cyan4973: Why is `liblz4` written 3 times on this line ?

lib/Makefile
```diff
@@ -40,7 +40,7 @@ LIBVER=$(LIBVER_MAJOR).$(LIBVER_MINOR).$(LIBVER_PATCH)
 DESTDIR?=
 PREFIX ?= /usr/local
 CFLAGS ?= -O3
-CFLAGS += -I. -std=c99 -Wall -Wextra -Wundef -Wshadow -Wcast-align -Wcast-qual -Wstrict-prototypes -pedantic
+CFLAGS += -I. -std=gnu99 -Wall -Wextra -Wundef -Wshadow -Wcast-align -Wcast-qual -Wstrict-prototypes -pedantic
```
Cyan4973: Need reference to properly understand the definition of `gnu99`, and most importantly the remaining differences with c99.
Note also that such command should work with alternative compilers, such as clang.

Cyan4973:
Thanks @lpsantil . It's a very interesting submit.

Most of the proposed changes are fine.
Nonetheless, as everything else should still work as usual, for the time being, there are 2 issues that need to be solved : 

- `make install` is broken with this patch. See https://travis-ci.org/Cyan4973/lz4/jobs/64049092
- `gnu99` is slightly different from `c99`, but it's unclear by how much, and what could be the impact. In a bid to reduce confusion, I would prefer to keep `c99` if that's possible. If not impossible, please explain the issue, and point to references about using `gnu99` instead of `c99`.

lib/Makefile
lpsantil: Copy-paste error on my part.

lib/Makefile
lpsantil: Part of the issue is that djgpp has a (semi-standard) _DOS_ API in addition to a _POSIX_ (almost-)compliant API.  `-std=gnu99` works around an issue in where `__STRICT_ANSI__` is defined with `-std=c99` causing DJ's headers to essentially nullify [`io.h`][0], [`fcntl.h`][1], and [`unistd.h`][2].  Essentially, it seems like a bug in djgpp's headers. The djgpp maintainers are currently discussing pushing djgpp 2.05 with gcc 5.1.0, so it is probably a good time to get such bugs fixed.

I can test the change on OSX 10.10 and make sure there are no issues in compiling or basic CLU usage with `clang`.  Beyond that, I'm open to your suggestions.  I was able to do a brief check of `clang -std=gnu99 -E dM - </dev/null` (dump preprocessor defines) and noted that it seemed to undefine `__STRICT_ANSI__` as well.  I assume `-std=gnu99` is a _C99+_ mode [(3)][3][(4)][4][(5)][5], per the docs.  

[0]: http://www.delorie.com/bin/cvsweb.cgi/djgpp/include/io.h?rev=1.8
[1]: http://www.delorie.com/bin/cvsweb.cgi/djgpp/include/fcntl.h?rev=1.11
[2]: http://www.delorie.com/bin/cvsweb.cgi/djgpp/include/unistd.h?rev=1.21
[3]: https://gcc.gnu.org/onlinedocs/gcc-4.9.2/gcc/Standards.html
[4]: https://gcc.gnu.org/c99status.html
[5]: https://gcc.gnu.org/onlinedocs/gcc-4.9.2/gcc/C-Extensions.html#C-Extensions

lpsantil:
I see what went wrong with `make install`.  When I broke apart `liblz4` target into `liblz4-static` and `liblz4-dynamic` in `lib/Makefile` and then added the `.PHONY` line, `install` target stopped building `.so` file.  Obviously, on DOS, there are no standard DLLs (though djgpp has [DXE3][0]).  Simple enough to fix.

[0]: http://www.delorie.com/djgpp/v2faq/faq22_15.html

t-mat:
First of all, I love this PR :godmode::+1:

Since we could not maintain DOS (or DOS-like) environment, we will not be able to keep testing.
So I propose that add `contrib/` directory for unofficial contributions like this PR.

Here is a directory structure I supposed :

 - lz4
   - contrib/
     - djgpp/
         - README.md
         - Makefile
         - programs/
            - Makefile

And standard procedure to build contributions :

```
cd /path/to/lz4/
cd contrig/djgpp/
cat README.md | less
sudo apt-get install -y ...
make
```

Cyan4973:
Looks like a pretty good suggestion

lpsantil:
In that case, I'll revert my `Makefile` changes and re-organize based on the discussed file structure.

lpsantil:
FYI, I used `git show djgpp~4:Makefile >Makefile` (and the like) to back out my `{.,lib,programs}/Makefile` changes.

Cyan4973:
Is it ready for a merge ?

lpsantil:
Yes.

tokiclover:
Why waste this amount of space for contributor sub-license when the main
license is already 2-clause (new/simplified) BSD and the progs on GPLv2?

A sensible approach would be to add COPYING.BSD-2/COPYING.GPL-2 in
the root of the source directory; and then add an AUTHORS file for this
purpose in listing contributor per what was contributed.

Of course, using a header per file is not negated by this... quite the
contrary.

Otherwise, any contributor would bring in a copy of the same license again
and again.

Many projects use this approach to ease life on the licensing terms and
maintain AUTHORS file and headers instead.

t-mat:
@tokiclover, it's a good point !  But since your objection is more broader than this PR, could you create new issue for license and directory structure problem ?

I believe some files are also technically in "no-license" (e.g. cmake_unofficial/CMakeLists.txt) and these files will increase by future contributions.

tokiclover:
This is rather a suggestion rather than... pointing to a issue. Sure, I
could open an issue ticket for this and suggest the simplest approach,
meaning `COPYING.{BSD-2-clause,GPLv2}` files addition and usage of
dedicated header in files for authors/contributors and addition of
`AUTHORS` file for the main/noticeable/regular contributors per
contribution.

Cyan4973:
One core reason for separating LICENSE/COPYING files into different directory
was that many users were complaining it was too difficult to understand
which part of the source code is BSD, an which one is GPL.

Yes, even though there was a header at the top of each file,
and a detailed reminder into several of them,
on top of being also notified on the homepage.

These complaints totally disappeared
after separating BSD & GPL files into different directory
with each their dedicated license files.

My fear with moving again general license file at the top repository directory
would be a replay of these past confusions.

tokiclover:
2015-05-29 13:35 GMT+02:00 Yann Collet <notifications@github.com>:

> My fear with moving again general license file at the top repository
> directory
> would be a replay of these past confusions.
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/Cyan4973/lz4/pull/114#issuecomment-106779785>.
>

Good point! Still, those that need that kind of information should know a
little
about this kind of things with such a common approach which is widely used
in numerous projects.

Already opened a ticket #115 for this! Lets follow the discussion there,
then.

-- 
Error 404: gentoo user <0123456789> was not found

PR #118, opened 2015-06-16
Add example which uses the frame api through the library

keszybz:


Cyan4973:
Hi Zbigniew

This looks like a great addition.

Unfortunately, it doesn't pass the automated tests on Travis CI.
I guess there is an issue with creating the library, so the linking stage fails.

Another detail is that your example only adresses compression.
This looks complementary to @t-mat suggestion to add an LZ4 frame decompression example.
I guess it would be great to have both, since without decompression, there is no validation that compression actually worked as expected.

_(Note : alternatively, it is also possible to use the normal `lz4` utility to validate the result)_

keszybz:
streaming-examples fails because the check clones the repository and only builds the examples, so lz4.so is not available. It now builds the lib first in the updated commit. LD_LIBRARY_PATH should do the right thigh for unixy systems. I'm not sure what the proper way for windows is to make sure that the newly-built version of the library is loaded.

I added second commit which uses lz4 to test the output file. lz4 is also build in streaming-examples to make sure they pass.

keszybz:
A note to self: don't use master branch for pull requests in the future.

A note to others: previous commit which is now disappeared was 4153c8c346,

Cyan4973:
That's a very good job. Thanks @keszybz !
I'll redirect the pull request to dev.
Apart from that, I see nothing else left to do.

Cyan4973:
Closed - replaced by #122 to redirect result to dev
(come on github, give us this feature : https://github.com/isaacs/github/issues/18 )

t-mat:
I'm late, but great work @keszybz !

PR #123, opened 2015-06-25
Fix cmake build to use gnu flags on clang

toddlipcon:
Previously, the cmake build was only adding -fPIC and -std=c99 on
gcc. However, these flags are also appropriate when building with
clang.


Cyan4973:
Thanks for the patch Todd.
Could you make it merged into dev, instead of master ?
Thanks

toddlipcon:
Sure, best I can tell I have to re-open the pull request. Will do.

PR #128, opened 2015-06-28
Improved implementation of wild copy

mtopolnik:
As discussed in https://github.com/Cyan4973/lz4/issues/126.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -67,7 +67,6 @@
 **************************************/
 #include "lz4.h"
 
-
 /**************************************
 *  Compiler Options
 **************************************/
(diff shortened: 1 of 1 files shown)


Cyan4973:
This version still requires some clean up to at least pass compilation tests with less warnings.
Then it will become easier to concentrate on critical memory boundaries issues.

mtopolnik:
I did try to correct the const-ignoring casts, but on my side i do not get these warnings. Unfortunately I didn't catch all the places where `const` needs to be added.

mtopolnik:
Some notes: i couldn't reproduce the segfault even after testing my code against various files on the file system. However, I learned that in some cases it is possible to get segfault for unaligned pointer derefence---and that's something my code is guilty of. If you don't mind, I might create another PR just to be able to see the build errors from your CI system and perhaps manage to remove the errors.

Cyan4973:
Sure, the wildcopy branch is yours to use. I will refrain from doing any change to it, in an effort to avoid repository desynchronisation.

For information, my experience was exactly the reverse as yours : I couldn't make this version work, on any file. It would always segfault during decompression. Strange we have so different behaviors...

Also, do you use the same compilations flags as suggested within the Makefile, or do you use your own ?

mtopolnik:
I'm building on Mac, that could be key. All I do is `make && make install`, then go use it from the cmdline. I contacted my colleague who is an expert in C (unlike me) so I'll try to heed his advice on how to make the code safer.

PR #132, opened 2015-06-29
Different approach to pointer arithmetics

mtopolnik:
I created this PR to see the results of CI builds.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -113,19 +113,19 @@
 **************************************/
 #if defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L)   /* C99 */
 # include <stdint.h>
-  typedef  uint8_t BYTE;
-  typedef uint16_t U16;
-  typedef uint32_t U32;
-  typedef  int32_t S32;
-  typedef uint64_t U64;
-  typedef  int64_t S64;
+  typedef   uint8_t BYTE;
+  typedef  uint16_t U16;
+  typedef  uint32_t U32;
+  typedef   int32_t S32;
+  typedef  uint64_t U64;
+  typedef uintptr_t ADDR;
 #else
   typedef unsigned char       BYTE;
   typedef unsigned short      U16;
   typedef unsigned int        U32;
   typedef   signed int        S32;
   typedef unsigned long long  U64;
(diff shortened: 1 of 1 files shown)


Cyan4973:
OK, this version is much cleaner. With most of the noise level down, it's now possible to concentrate on the more critical issue, regarding clean respect of memory boundaries. `valgrind` can be a pretty good tool for such hunt.

t-mat:
`intptr_t`, `uintptr_t` (C89, C99:7.18.1.4/1, C11:7.20.1.4/1) and `ptrdiff_t` (C89, C99, C11:6.5.6/9) are standard compliant and better type for some platform such as x32 ABI.

mtopolnik:
As @t-mat said, and replying to Yann's inline code comment, I'm using ADDR so as to make use of the more modern types where applicable. It would have been even more correct to define as follows: `typedef intptr_t ADDR; typedef uintptr_t UADDR;` and, as a fallback, `typedef ptrdiff_t ADDR; typedef size_t UADDR;` but I wasn't sure about the consequence of using a signed equivalent to a pointer. Theoretically that would take 65 bits to accommodate all cases.

ido:
Just FYI, in x86_64/amd64 pointer contents are only 48 bits wide, even though the register that holds them is 64 bits wide:
http://en.wikipedia.org/wiki/X86-64#Canonical_form_addresses

Those top 16 bits are used in various types of pointers by non-portable libraries like Facebook's folly to add additional functionality to custom pointer types, for example see the comments in the following two files:
https://github.com/facebook/folly/blob/master/folly/PackedSyncPtr.h
(...PackedSyncPtr uses one of the 16 bits as a spin lock, so an atomic 64 bit pointer can also be a spin lock on the resource it points to...)
https://github.com/facebook/folly/blob/master/folly/DiscriminatedPtr.h
(...DiscriminatedPtr stores type info, etc. in the top 16 bits.)

mtopolnik:
Interesting, @ido... With today's servers already delivered with 4 TB of RAM, and with some state-of-the-art SSD arrays providing `mmap` shims which redirect anon `mmap` to SSD, currently allowing up to 16 TB of JVM heap, this will soon have to be extended.

PR #133, opened 2015-06-30
Fix a cast, change inlining

mtopolnik:


mtopolnik:
At this point all tests seem to be passing, except for 32-bit ones (which is expected becase my code is specialized to 64-bit little-endian at the moment).

mtopolnik:
There is one other subtlety detected by valgrind: in `MT_wildCopyMisalignedLE` I have allowed myself to read from an address before the start of the malloc'd block, then carefully mask out the part of data which doesn't belong to the block. This can be rewritten to read within bounds, then apply a right shift to reach the same outcome.

Cyan4973:
Sounds pretty good.
Sure, this last modification is welcomed to finally silence this last out-of-bound memory warnings.

Later it will be time to investigate 32-bits.
But before that, let's make some more thorough performance tests ...

mtopolnik:
Absolutely, in my tests the regular case performance does suffer somewhat at the moment. My observation was a ~15% drop on a typical file compressible by 4:1.

Cyan4973:
That's the result I got too.
Cyclic data get the boost as expected, but normal ones are for the moment significantly slowed down.

I suspect some future work on branch prediction.

PR #134, opened 2015-06-30
Avoid out-of-bounds reading

mtopolnik:
Reworked reading from before the malloc'd buffer into within-bound reading followed by a left shift (little-endian logic applies here).

Unfortunately, the performance of `MT_wildCopyNarrowLE` is halved when unaligned pointer ops are replaced with `memcpy`. More work should be applied to avoid `memcpy`.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -214,21 +214,23 @@ FORCE_INLINE void LZ4_wildCopy(void* dstPtr, const void* srcPtr, void* dstEnd)
 static void MT_wildCopyMisalignedLE(BYTE* destStart, const BYTE* srcStart, BYTE* const destEnd)
 {
     ADDR bytesToAlignDest = -((ADDR)destStart) & 7;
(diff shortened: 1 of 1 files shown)


mtopolnik:
I have completed the implementation of `MT_wildCopyNarrowLE` which uses aligned writing, but it didn't solve the performance issue. However, I also found out that the issue is specific to `clang` (on my local system) whereas `gcc` does rather well with the current code.

The code is committed [here](https://github.com/mtopolnik/lz4/blob/wildcopy_align_narrow/lib/lz4.c#L282) should there be any future need to evaluate it.

Cyan4973:
I guess there will be a need to work on branching predictibility to make this solution fly under "normal" conditions too.

mtopolnik:
One possile idea to try out is using `|` instead of `||` in the first `if` condition. Theoretically the compiler should already know what is best for performance, but there still may be a chance for improvement.

PR #136, opened 2015-07-03
Inline small helper functions to increase performance

jack-pappas:
While working on some performance optimizations for blosc/c-blosc a while back, I noticed that MSVC doesn't inline some of the small functions in lz4 that check for native integer size, CPU endianness, or are used to read small pieces of data from memory. This occurs with VS2013 community edition and compiling with /O2 and the normal inlining heuristics (where MSVC may choose to inline functions even if they're not explicitly marked with 'inline').

These functions are called quite often and the overhead of calling them is large relative to their execution time. I've found that inlining them improves compression/decompression performance significantly (10-20%). The values returned by some of the functions can be determined at compile-time, so inlining them also allows compilers to make better use of dead-code elimination and shrink the size of the code.

I haven't benchmarked these changes using any other compilers, so I don't know if they'll see a similar performance increase or if their heuristics already cause these functions to be inlined. Even if they are already inlining the functions, I don't think it will hurt performance to make the inlining explicit.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -133,23 +133,23 @@
 **************************************/
 #define STEPSIZE sizeof(size_t)
 
-static unsigned LZ4_64bits(void) { return sizeof(void*)==8; }
+FORCE_INLINE unsigned LZ4_64bits(void) { return sizeof(void*)==8; }
 
-static unsigned LZ4_isLittleEndian(void)
+FORCE_INLINE unsigned LZ4_isLittleEndian(void)
 {
     const union { U32 i; BYTE c[4]; } one = { 1 };   /* don't use static : performance detrimental  */
     return one.c[0];
 }
 
 
-static U16 LZ4_read16(const void* memPtr)
+FORCE_INLINE U16 LZ4_read16(const void* memPtr)
 {
     U16 val16;
     memcpy(&val16, memPtr, 2);
     return val16;
 }
 
-static U16 LZ4_readLE16(const void* memPtr)
+FORCE_INLINE U16 LZ4_readLE16(const void* memPtr)
 {
     if (LZ4_isLittleEndian())
     {
@@ -162,7 +162,7 @@ static U16 LZ4_readLE16(const void* memPtr)
     }
 }
 
-static void LZ4_writeLE16(void* memPtr, U16 value)
+FORCE_INLINE void LZ4_writeLE16(void* memPtr, U16 value)
 {
     if (LZ4_isLittleEndian())
     {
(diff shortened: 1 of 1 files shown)


t-mat:
Hi, @jack-pappas
Interesting work !

First of all, this project doesn't accept PR to `master` branch.  So please send it to `dev` branch.  (Please see "Branch Policy" in [README.md](https://github.com/Cyan4973/lz4/blob/master/README.md))

I've checked your work, and confirmed there is a improvement.  But for my environment (MSVS Community 2013 12.0.31101.00 Update 4), differences are quite smaller than yours : 0.4% for compression, 1.2% for decompression.
And in my environment, these small functions are completely inlined in both current `master` and your `inline-small-functions` branch.  Perhaps this is the reason of less improvement.

I want to know the reason of this difference.
Could you kindly try the following procedure and post the result ?  I'd also like to ask you my `CL_FLAGS` has significant difference from yours or not.

## Building procedure

```
cd /d "%USERPROFILE%"
mkdir lz4-pr136
cd lz4-pr136
git clone -b master https://github.com/Cyan4973/lz4.git master
git clone -b inline-small-functions https://github.com/jack-pappas/lz4.git inline-small-functions

call "%VS120COMNTOOLS%vsvars32.bat"
set CL_FLAGS=/nologo /O2 /Oy- /Oi /MD /GS /GL /Gy /Gd /FA /Ilib
set LZ4_LIBS=lib\lz4.c lib\lz4hc.c lib\lz4frame.c lib\xxhash.c
set LZ4_SRCS=programs/lz4cli.c programs/lz4io.c programs/bench.c

pushd master
cl %CL_FLAGS% %CL_FLAGS% %LZ4_LIBS% %LZ4_SRCS%
popd
pushd inline-small-functions
cl %CL_FLAGS% %CL_FLAGS% %LZ4_LIBS% %LZ4_SRCS%
popd

@rem Benchmark
for /l %i in (1,1,10) do @master\lz4 -b master\README.md
for /l %i in (1,1,10) do @inline-small-functions\lz4 -b master\README.md

@rem Gather environment info
chcp 437
set | find /i "PROCESSOR_"
systeminfo | findstr /B /C:"OS Name" /C:"OS Version"
cl | find /i "version"
```

Here is a [my environment info and full benchmark log](https://gist.github.com/t-mat/27c0fa4ea82edc5008b3)

jack-pappas:
Hi @t-mat, thank you for the quick response.

I ran your benchmark and got similar results, so I went back and compared your compile flags to what I'd used for my test. c-blosc uses CMake as the build system, and I'd compiled the benchmark using the standard ``RelWIthDebInfo`` configuration (``cmake --build . --config RelWithDebInfo``). It turns out that configuration adds ``/Ob1`` to the compile flags, overriding the implicit ``/Ob2`` that comes from the ``/O2`` flag. Compiling with the ``Release`` configuration replaces the explicit ``/Ob1`` with an explicit "/Ob2", in which case the performance of ``master`` and my change is the same. That being the case, I don't think this change is needed so I'll close the issue.

PR #141, opened 2015-08-15
Added debianization

bioothod:
Currently all debian distros ship rather old and incomplete subset of lz4, in particular they all miss frame processing implementation.
I'm not sure whether new package will ever find its way into debian distros, but having all debianizations in the source is quite handy for those who may wish to build it locally.

Debianization is rather simple (there are lintian bugs and warnings for instance). Tested on wheezy.


Cyan4973:
Thanks, looks like a good commit request.

Note that no commit is ever pulled directly into master. New commits must either land first into their own feature branch, or into dev branch.

bioothod:
Ok, I see. I didn't know about branch naming policy, so created PR against master.
Should I remake pull request or you will merge it by hands?
I only updated CMakeFiles.txt outside of the debian tree, so there should be no merge conflicts.

Cyan4973:
> Should I remake pull request or you will merge it by hands?

As you wish

bioothod:
I've added https://github.com/Cyan4973/lz4/pull/142 against `dev` branch, is it ok?

Cyan4973:
yep. replaced by #142

PR #150, opened 2015-09-02
Remove blank page from manual.

mwgamera:
The manual page for lz4, unlz4, lz4cat programs had 8 empty lines before the `.TH` request which made the output start with a blank page (or just with empty lines when typesetting for terminal).

### DIFF
--- a/programs/lz4.1
+++ b/programs/lz4.1
@@ -1,13 +1,13 @@
-\"
-\" lz4.1: This is a manual page for 'lz4' program. This file is part of the
-\" lz4 <https://code.google.com/p/lz4/> project.
-\" Author: Yann Collet
-\"
-
-\" No hyphenation
+.\"
+.\" lz4.1: This is a manual page for 'lz4' program. This file is part of the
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks for the feedback.

Strangely, I never observed the effect mentioned (blank page)
so I guess it happens under specific conditions.

As a minor reminder, all contributions must be merged into master first.
Here you'll find that the next version of man page is already a little bit different, enough to break git history and require another PR.

I can also copy by hand your modifications. They are simple enough.

Regards

mwgamera:
Ah, sorry for the wrong branch. I will pay more attention next time.
Please apply the patch as appropriate it if that's not too much of a trouble.

As for the specific conditions, it might depend on version of troff and its man macros, but it happens for me with groff 1.22.3 which is used in a lot of systems, so you can probably observe it too if you try to typeset it as PostScript or as PDF (something like `man -t ./lz4.1 > lz4.ps` or just `groff -Tpdf -man lz4.1 > lz4.pdf`).

Issue #157, opened 2015-09-21
LZ4F_getFrameInfo fails on valid frame with no compressed blocks


hntrmrrs:
If I attempt to compress an empty buffer (e.g., by using the lz4 example command-line utility on an empty file), getting the frame info fails upon decompression. However, decompressing without using LZ4F_getFrameInfo appears to work as expected.

``` C
LZ4F_preferences_t prefs;
LZ4F_decompressionContext_t dCtx = NULL;
LZ4F_compressionContext_t cctx = NULL;
LZ4F_frameInfo_t frame_info;

memset(&prefs, 0, sizeof(prefs));
memset(&frame_info, 0, sizeof(frame_info));
prefs.frameInfo.blockMode = LZ4F_blockIndependent;
prefs.frameInfo.contentChecksumFlag = LZ4F_contentChecksumEnabled;

char buf[200000];
size_t errorCode;
errorCode = LZ4F_createCompressionContext(&cctx, LZ4F_VERSION);
if (LZ4F_isError(errorCode)) fprintf(stderr, "failed: %s\n", LZ4F_getErrorName(errorCode));
size_t written = 0;
errorCode = LZ4F_compressBegin(cctx, buf, 200000, &prefs);
if (LZ4F_isError(errorCode)) fprintf(stderr, "failed: %s\n", LZ4F_getErrorName(errorCode));
written += errorCode;
errorCode = LZ4F_compressEnd(cctx, buf + written, 200000 - written, NULL);
if (LZ4F_isError(errorCode)) fprintf(stderr, "failed: %s\n", LZ4F_getErrorName(errorCode));
errorCode = LZ4F_freeCompressionContext(cctx);
if (LZ4F_isError(errorCode)) fprintf(stderr, "failed: %s\n", LZ4F_getErrorName(errorCode));

errorCode = LZ4F_createDecompressionContext(&dCtx, LZ4F_VERSION);
if (LZ4F_isError(errorCode)) fprintf(stderr, "failed: %s\n", LZ4F_getErrorName(errorCode));
size_t avail_in = written;
result = LZ4F_getFrameInfo(dCtx, &frame_info, static_cast<const void*>(buf), &avail_in);
if (LZ4F_isError(errorCode)) fprintf(stderr, "failed: %s\n", LZ4F_getErrorName(errorCode));
```

The above code snippet fails with:

```
failed: ERROR_frameHeader_incomplete
```

Cyan4973:
Thanks for notification.
[...]

PR #158, opened 2015-09-22
Fix bug where minimally sized frame headers are not decoded

hntrmrrs:
I believe this is the first step toward solving issue #157.

The remaining issue is actually more subtle and may not actually be incorrect behaviour. `LZ4F_getFrameInfo()` called on an empty frame will return `-LZ4F_ERROR_frameHeader_incomplete` when the decode stage is `<= dstage_storageHeader`. If an empty frame is encountered, `LZ4F_decompress()` will simply consume `srcBuffer` without writing anything to `dstBuffer`. Should `LZ4F_getFrameInfo()` return the *most recent* frame info? That's how I read the documentation, but it could just need clarification.

### DIFF
--- a/lib/lz4frame.c
+++ b/lib/lz4frame.c
@@ -130,6 +130,7 @@ typedef struct LZ4F_cctx_s
 typedef struct LZ4F_dctx_s
 {
     LZ4F_frameInfo_t frameInfo;
+    U32    frameInfoValid;
     U32    version;
     U32    dStage;
     U64    frameRemainingSize;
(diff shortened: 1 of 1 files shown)


hntrmrrs:
The second commit attempts to solve the problem by storing an additional flag in the decompression context which is `1` if the frame info structure is valid and `0` if not.

Cyan4973:
Thanks, it looks like a good fix.
As the proposal includes changing the size of an internal structure, I'll review it a bit more thoroughly, just to make sure there is no subtle side effect.

hntrmrrs:
Alternatively we could make use of an unused bit in `dstage` because it's
unlikely to become such a large state machine that it would need all 32.
What do you think?

Cyan4973:
That's a fairly good work around.
But maybe we won't even need that.
I'm just being cautious regarding struct size, and as long as it causes no problem, I prefer the greater readability of dedicated variable.

Drako:
are there any news on this one?
I seem to be running into the ERROR_frameHeader_incomplete issue currently and a fix would be nice :-)

Cyan4973:
You are right this is long overdue.
Latest update in "dev" branch should fix this issue

PR #161, opened 2015-10-30
20151021 examples

KyleJHarper:
Per conversation here:
https://groups.google.com/forum/#!topic/lz4c/jQ9rMY6gvbg

Adds examples for basic compression/decompression and a slightly more advanced example demonstrating the call stacks for each.

PR #162, opened 2015-10-30
Additional examples

KyleJHarper:
Creating new PR to supersede #161 so we target the correct (dev) branch.


Cyan4973:
Thanks, that's a great addition

PR #165, opened 2015-11-04
adapt generic to _destSize, preserving old usage and adding _destSize…

infidob:
… usage


Cyan4973:
That's a good direction.

Note however that a number of [automated CI tests have failed](https://travis-ci.org/Cyan4973/lz4/builds/89216791), so it would probably be better to solve them.

 (you can just update your branch, it seems tests are automatically run again at each update).

PR #175, opened 2015-12-23
Adding unaligned access support for aarch64

rockyzhang:
For ARMv8 processors, unaligned access has been supported. The "__ARM_FEATURE_UNALIGNED" is only turned on for 32bit ARM gcc for armv7-a processors. Thus __aarch64__ is needed to turn on the XXH_USE_UNALIGNED_ACCESS option.

### DIFF
--- a/lib/xxhash.c
+++ b/lib/xxhash.c
@@ -40,7 +40,8 @@ You can contact the author at :
  * If you know your target CPU supports unaligned memory access, you want to force this option manually to improve performance.
  * You can also enable this parameter if you know your input data will always be aligned (boundaries of 4, for U32).
  */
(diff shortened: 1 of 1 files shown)


Cyan4973:
Note that "master" corresponds to r131, and is not latest branch.

You'll find an updated version into "dev" branch, where all PR must be merged first (no direct merge to master is allowed).

In this case, it's important as the xxhash library version has been updated. You'll find that your patch does no longer apply, since unaligned access has been completely revamped.
See http://fastcompression.blogspot.fr/2015/08/accessing-unaligned-memory.html for more details.

I expect the "dev" branch to automatically use unaligned access capability of arm64 cpus. At least, that's what I observed in my tests.

Cyan4973:
Closing.

Note that this patch would have been better director to [xxhash](https://github.com/Cyan4973/xxHash) project. Unaligned memory accesses have been updated there too, and should prove optimal on arm64 cpus. 

Feedbacks on real hardware are welcomed.

PR #179, opened 2016-01-10
Fix build error on MinGW

kou:
Header file name is case insensitive on Windows but it is case sensitive
on Linux. "Windows.h" can't be found on Linux.


Cyan4973:
Good catch @kou !

Mind to target your PR at "dev" branch instead of "master " ?
(see [branch policy](https://github.com/Cyan4973/lz4/blob/master/README.md) )

kou:
Sorry. I re-created PR as #181.

PR #185, opened 2016-02-13
Add support for safe in-place decoding

jwerner-chromium:
I'm trying to import LZ4 code into the coreboot.org open source firmware project. We work in an environment with very tight memory and I want to use LZ4 for in-place decompression (where you load the compressed input at the end of the output buffer so that it will overwrite itself during decompression).

This should work in the vast majority of cases as long as the output buffer is at least a few bytes larger than the actual output data, but it is important for us to reliably detect whether it worked or not (rather than risking corrupted output data). I think the small extra check added with this patch should be all that's needed... please let me know if I'm overlooking something.


Cyan4973:
Thanks @jwerner-chromium, I believe this is a good patch which reaches its intention.

The maximum overwriting risk is `WILDCOPYLENGTH ` beyond current pointer, as you correctly test.
This length is, btw, 8 bytes with current version of LZ4, so maybe it should be documented somewhere, for people trying to achieve in-place decompression.

Both the test and the move from `memcpy()` to `memmove()` are supposed to be very minor hits, likely with insignificant impact on decompression speed. But it would be better to test it. Have you got benchmark numbers from before / after ?

As a sidenote, the [frame format](https://github.com/Cyan4973/lz4/blob/master/lz4_Frame_format.md) also includes a checksum verification, useful to check for payload corruption. But it also adds a few bytes in header, footer, and metadata.

Regards

jwerner-chromium:
Thanks for the quick review and merge, Yann! I've run some more extensive tests now and have benchmarked my ARM64 Cortex-A53 to decompress about 1-2% slower with this patch (decompressing ~3.5MB of object code in ~14.95 instead of ~14.70 ms). I hope that's an acceptable cost. (I've also tried to benchmark an Intel Haswell-U, but it was an order of magnitude slower than the A53 for the same data (~707ms) so I think there was something wrong with the optimization there. Still, the difference this patch made was even smaller.)

I know I could also use content checksums, but I presume calculating the checksum would take way longer so I prefer this approach.

Issue #192, opened 2016-03-30
lz4 -h should print library version in addition to cli version


jzhuge:
"lz4 -h" should print the library version by LZ4_versionNumber() because one cli version may work with multiple lib versions, e.g., "r128" works with 1.6.0/1.7.0/1.7.1. Other softwares (e.g. hadoop) only bundle the library code. It will make it easier to perform compatibility check if "lz4 -h" prints the lib version.

Is Linux lz4 package version actually the git tag of this repo?

Cyan4973:
It's a good point @jzhuge .
I'll review your PR.

Cyan4973:
Here is the current behavior in `dev` branch : 
```
./lz4 -h
*** LZ4 command line interface 64-bits v1.7.2, by Yann Collet ***
Usage :
      ./lz4 [arg] [input] [output]
```

Now library, cli and release number are one and the same, and displayed when invoking `lz4 -h` or `-lz4 -V` .

PR #193, opened 2016-03-30
lz4cli: print library version along with cli version

jzhuge:
Resolve issue #192.

### DIFF
--- a/lib/lz4frame.c
+++ b/lib/lz4frame.c
@@ -205,8 +205,8 @@ static void LZ4F_writeLE32 (BYTE* dstPtr, U32 value32)
 static U64 LZ4F_readLE64 (const BYTE* srcPtr)
 {
     U64 value64 = srcPtr[0];
-    value64 += (srcPtr[1]<<8);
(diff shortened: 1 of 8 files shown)


Cyan4973:
As a minor detail,
one thing I would like to preserve is a welcomed message using < 80 characters,
to print on a single line.

One way to do so could be to remove the compilation date, and use the space for the library version.
It would also make lz4 cli build reproducible, which is slowly becoming a desirable property.

t-mat:
Hi @jzhuge,  It's good PR :)  I found there are 2 minor issues:

- Since this PR is also useful to check actual dynamic linked libraries version, `LZ4_versionNumber()` would be better than `LZ4_VERSION_NUMBER`.
- We should consider once about multiple dot version number such as `1.7.3`.

jzhuge:
Great comments. I will make changes.

Cyan4973:
Ah, and now that I noticed,
no PR to master please.
They must go first through "dev" branch (or their own feature branch when available).

PR #194, opened 2016-03-31
lz4cli: print library version

jzhuge:
Resolve issue #192.

> *** LZ4 CLI 64-bits r132 (lib 1.7.2), by Yann Collet ***

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -414,6 +414,7 @@ typedef enum { full = 0, partial = 1 } earlyEnd_directive;
 *  Local Utils
 **************************************/
 int LZ4_versionNumber (void) { return LZ4_VERSION_NUMBER; }
+const char* LZ4_versionString (void) { return LZ4_VERSION_STRING; }
 int LZ4_compressBound(int isize)  { return LZ4_COMPRESSBOUND(isize); }
 int LZ4_sizeofState() { return LZ4_STREAMSIZE; }
 
--- a/lib/lz4.h
+++ b/lib/lz4.h
@@ -50,8 +50,15 @@ extern "C" {
 #define LZ4_VERSION_MAJOR    1    /* for breaking interface changes  */
 #define LZ4_VERSION_MINOR    7    /* for new (non-breaking) interface capabilities */
 #define LZ4_VERSION_RELEASE  2    /* for tweaks, bug-fixes, or development */
+
 #define LZ4_VERSION_NUMBER (LZ4_VERSION_MAJOR *100*100 + LZ4_VERSION_MINOR *100 + LZ4_VERSION_RELEASE)
 int LZ4_versionNumber (void);
+
+#define LZ4_STR(str) #str
+#define LZ4_XSTR(str) LZ4_STR(str)
+#define LZ4_VERSION_STRING LZ4_XSTR(LZ4_VERSION_MAJOR) "." \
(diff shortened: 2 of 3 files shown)


lib/lz4.h
```diff
@@ -50,8 +50,10 @@ extern "C" {
 #define LZ4_VERSION_MAJOR    1    /* for breaking interface changes  */
 #define LZ4_VERSION_MINOR    7    /* for new (non-breaking) interface capabilities */
 #define LZ4_VERSION_RELEASE  2    /* for tweaks, bug-fixes, or development */
+#define LZ4_VERSION_STRING   "1.7.2"
```
Cyan4973: In order to avoid accidental unsynchronization, a personal favorite would be for `LZ4_VERSION_STRING` to be generated from `LZ4_VERSION_MAJOR`, `LZ4_VERSION_MINOR` and `LZ4_VERSION_RELEASE`.

There is an equivalent construction available in zstd cli (might be shortened a bit) :
https://github.com/Cyan4973/zstd/blob/master/programs/zstdcli.c#L76

jzhuge:
@Cyan4973 pls review the latest change.

lib/lz4.h
```diff
@@ -50,9 +50,17 @@ extern "C" {
 #define LZ4_VERSION_MAJOR    1    /* for breaking interface changes  */
 #define LZ4_VERSION_MINOR    7    /* for new (non-breaking) interface capabilities */
 #define LZ4_VERSION_RELEASE  2    /* for tweaks, bug-fixes, or development */
+
 #define LZ4_VERSION_NUMBER (LZ4_VERSION_MAJOR *100*100 + LZ4_VERSION_MINOR *100 + LZ4_VERSION_RELEASE)
 int LZ4_versionNumber (void);
 
+#define QUOTE(str) #str
```
Cyan4973: OK, so this is the `zstdcli.c` construction re-used as-is.

It works, although there is a subtle difference here :
`zstdcli.c` declarations remain "private", only for its module.
`lz4.h` declarations will be included. 
As a consequence, any public macro / type / function will pollute the naming space of importing project.

To reduce risks of naming collisions, it is important to prefix any imported definition with `LZ4_`.
Here, it applies to `QUOTE` and to `EXPAND_AND_QUOTE`  (you may select a shorter name for this one if you feel it helps readability).

lib/lz4.c
```diff
@@ -414,6 +414,7 @@ typedef enum { full = 0, partial = 1 } earlyEnd_directive;
 *  Local Utils
 **************************************/
 int LZ4_versionNumber (void) { return LZ4_VERSION_NUMBER; }
+const char* LZ4_versionString (void) { return LZ4_VERSION_STRING; }
```
Cyan4973: I know that `"hello" "world"` inside a `printf()` will work the same as `"hello world"`.
But I'm not sure if it also works outside of `printf()`, and in this specific case as a `return` argument.

I haven't tested it, so maybe just a simple test will answer my worries.

mwgamera: Yes it will. There's nothing special about `printf`, adjacent string literals are always concatenated in C.

lib/lz4.h
jzhuge: Good point. Fixed.

Cyan4973:
Looks great to me !

jzhuge:
Thanks @Cyan4973 and @t-mat for the review.

PR #195, opened 2016-04-02
New SONAME / code reuse

jengelh:
1. Is something I had to do for a Linux distribution. It must not happen that different ABI sets use the same SONAME and I chose "1.7" as a new identifier.
2. Maximum code reuse seems like a beneficial goal.

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -39,7 +39,7 @@ RELEASE?= r131
 DESTDIR?=
 PREFIX ?= /usr/local
 CFLAGS ?= -O3
-CFLAGS += -std=c99 -Wall -Wextra -Wundef -Wshadow -Wcast-qual -Wcast-align -Wstrict-prototypes -pedantic -DLZ4_VERSION=\"$(RELEASE)\"
+CFLAGS += -DXXH_NAMESPACE=LZ4_ -std=c99 -Wall -Wextra -Wundef -Wshadow -Wcast-qual -Wcast-align -Wstrict-prototypes -pedantic -DLZ4_VERSION=\"$(RELEASE)\"
 FLAGS  := -I../lib $(CPPFLAGS) $(CFLAGS) $(LDFLAGS)
 
 BINDIR := $(PREFIX)/bin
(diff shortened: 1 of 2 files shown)


programs/Makefile
```diff
@@ -39,7 +39,7 @@ RELEASE?= r131
 DESTDIR?=
 PREFIX ?= /usr/local
 CFLAGS ?= -O3
-CFLAGS += -std=c99 -Wall -Wextra -Wundef -Wshadow -Wcast-qual -Wcast-align -Wstrict-prototypes -pedantic -DLZ4_VERSION=\"$(RELEASE)\"
+CFLAGS += -DXXH_NAMESPACE=LZ4_ -std=c99 -Wall -Wextra -Wundef -Wshadow -Wcast-qual -Wcast-align -Wstrict-prototypes -pedantic -DLZ4_VERSION=\"$(RELEASE)\"
```
Cyan4973: `-DXXH_NAMESPACE` is only required to compile xxhash.
Since, in the rest of the file, programs are linked against lz4 library and do not compile xxhash anymore, this define seems unnecessary.

Cyan4973:
Could you explain why removing `LIBVER_MINOR` from the soname is a good thing ?

Cyan4973:
> It must not happen that different ABI sets use the same SONAME

Could you be more specific regarding targeted ABI change in the context of lz4 ?

So your proposition is to make `LIBVER_MINOR` part of the soname.
I'm not completely sure about what would break as a consequence.
I really don't know all this, so I feel very conservative.

Anyway, 2 issues to fix : 
- No PR into master. Please represent to "dev" branch (or a dedicated feature branch if you prefer)
- Linking with lz4 library only works in installation mode. But many tests only generate the binary without installing anything. They are currently broken with this patch. This must be solved.

Possible solution : only link with`-llz4` during `make install`

t-mat:
@Cyan4973, I suppose this PR adresses [changing `LZ4_STREAMDECODESIZE_U32` in r124](https://github.com/Cyan4973/lz4/commit/ca353cfe6e6ed078501774f0a25d2eb8190036c9#diff-5d9e16e39ec0d516e1169d36712b9d23L237).  But since description in lz4.h [clearly warns about using `LZ4_streamDecode_t` in dynamic linking context](https://github.com/Cyan4973/lz4/blob/f0a0e4430bd983ce3e054735717f89a095416f24/lib/lz4.h#L256), I think this change isn't problem.

Perhaps, we may move these lines to `lz4_static.h` in LZ4-ver2 :)

I'll wait @jengelh 's further explanation too.

jengelh:
Exactly as t-mat says. This is how the alarms sound:
```
$ abidiff  tr123/liblz4.so.1 tr124/liblz4.so.1
Functions changes summary: 0 Removed, 2 Changed (5 filtered out), 5 Added functions
Variables changes summary: 0 Removed, 0 Changed, 0 Added variable

5 Added functions:

  'function LZ4_streamHC_t* LZ4_createStreamHC()'    {LZ4_createStreamHC}
  'function int LZ4_freeStreamHC(LZ4_streamHC_t*)'    {LZ4_freeStreamHC}
  'function int LZ4_loadDictHC(LZ4_streamHC_t*, const char*, int)'    {LZ4_loadDictHC}
  'function void LZ4_resetStreamHC(LZ4_streamHC_t*, int)'    {LZ4_resetStreamHC}
  'function int LZ4_saveDictHC(LZ4_streamHC_t*, char*, int)'    {LZ4_saveDictHC}

2 functions with some indirect sub-type change:

  [C]'function int LZ4_compressHC_continue(void*, const char*, char*, int)' has some indirect sub-type changes:
    parameter 0 of type 'void*' changed:
      in pointed to type 'void':
        entity changed from 'void' to 'typedef LZ4_streamHC_t'

        size changed from 0 to 2097536 bits
  [C]'function LZ4_streamDecode_t* LZ4_createStreamDecode()' has some indirect sub-type changes:
    return type changed:
      in pointed to type 'typedef LZ4_streamDecode_t':
        underlying type 'struct __anonymous_struct__' changed:
          size changed from 128 to 256 bits
          1 data member change:
           type of 'unsigned int __anonymous_struct__::table[4]' changed:
             in array type 'unsigned int[4]':
               size changed from 128 to 256
               subrange 1 changed length from 4 to 8
```
abidiff is part of libabigail.

>Linking with lz4 library only works in installation mode.

Things like automake or cmake would easily address that (including some other shortcomings like all -W being currently eradicated when adding own CFLAGS), but judging from the directory name `cmake_unofficial`, it does not look like you are willing to make use of it, or are you?

Cyan4973:
> judging from the directory name cmake_unofficial, it does not look like you are willing to make use of it, or are you?

Indeed,
this is a 3rd-party build system, included in the repo for whoever would need it, but I don't actively maintain it.
I'm focused on the Makefile instead.

Cyan4973:
Looking at your example, `abidiff` looks like an interesting tool.
It would be helpful to control ABI continuity.

Unfortunately, its installation process is not straightforward.
I would have loved to just `apt-get` it, but there is apparently no way, making compilation from source the only option.

`abidiff` is included in a very large project (100 MB source tarball !) which is prone to dependency hell. There is a nice autoconf process which allows detecting problems before compilation, but that's all it does. Solving them is not magical. Quite a pity.

jengelh:
>abidiff is included in a very large project (100 MB source tarball !

Hm I don't see the hardship. For me, the tarball, as generated by tarring up a git pulldown, is less than 1 MB. Cf. https://build.opensuse.org/package/show/devel:tools/libabigail .

jengelh:
Oh well, looks like they added some testsuite data (which might be ignorable?).

PR #196, opened 2016-04-07
Optimized The Code.

Vishnu-Hacks:
The Makefile and the lz4io.c have been optimized for the x86. Compiler options are added to the make and expensive operations on lz4io.c has been replaced with bitwise operations. The Compression speed of the code have been increased by 24.9 MB/s and Decompression speed has been increased by 105MB/s.

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -39,9 +39,15 @@ RELEASE?= r131
 DESTDIR?=
 PREFIX ?= /usr/local
 CFLAGS ?= -O3
-CFLAGS += -std=c99 -Wall -Wextra -Wundef -Wshadow -Wcast-qual -Wcast-align -Wstrict-prototypes -pedantic -DLZ4_VERSION=\"$(RELEASE)\"
-FLAGS  := -I../lib $(CPPFLAGS) $(CFLAGS) $(LDFLAGS)
+CFLAGS += -std=c99 -Wall   -Wextra -Wundef -Wshadow -Wcast-qual -Wcast-align -Wstrict-prototypes -pedantic -DLZ4_VERSION=\"$(RELEASE)\"
 
+
+ifeq ($(shell uname --hardware-platform), i386)
+           CFLAGS += -march=prescott
+   endif
+
+
+FLAGS  := -I../lib $(CPPFLAGS) $(CFLAGS) $(LDFLAGS)
 BINDIR := $(PREFIX)/bin
 MANDIR := $(PREFIX)/share/man/man1
 LZ4DIR := ../lib
--- a/programs/lz4io.c
+++ b/programs/lz4io.c
@@ -384,7 +384,7 @@ int LZ4IO_compressFilename_Legacy(const char* input_filename, const char* output
         /* Compress Block */
         outSize = compressionFunction(in_buff, out_buff+4, inSize, outBuffSize, compressionlevel);
         compressedfilesize += outSize+4;
-        DISPLAYUPDATE(2, "\rRead : %i MB  ==> %.2f%%   ", (int)(filesize>>20), (double)compressedfilesize/filesize*100);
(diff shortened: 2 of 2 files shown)


programs/lz4io.c
```diff
@@ -384,7 +384,7 @@ int LZ4IO_compressFilename_Legacy(const char* input_filename, const char* output
         /* Compress Block */
         outSize = compressionFunction(in_buff, out_buff+4, inSize, outBuffSize, compressionlevel);
         compressedfilesize += outSize+4;
-        DISPLAYUPDATE(2, "\rRead : %i MB  ==> %.2f%%   ", (int)(filesize>>20), (double)compressedfilesize/filesize*100);
+        DISPLAYUPDATE(2, "\rRead : %i MB  ==> %.2f%%   ", (int)(filesize>>20), (double)compressedfilesize/((filesize<<6)+26);
```
Cyan4973: filesize * 64 + 26 != filesize * 100

programs/lz4io.c
```diff
@@ -539,8 +541,11 @@ static int LZ4IO_compressFilename_extRess(cRess_t ress, const char* srcFileName,
             if (sizeCheck!=outSize) EXM_THROW(35, "Write error : cannot write compressed block");
 
             /* Read next block */
-            readSize  = fread(srcBuffer, (size_t)1, (size_t)blockSize, srcFile);
+            readSize  = fread(srcBuffer, (size_t)1, (size_t)changeBlock, srcFile);
             filesize += readSize;
+
+	    i+=10;
+	    changeBlock=(size_t)LZ4IO_GetBlockSize_FromBlockId (readSize+i);
```
Cyan4973: `LZ4IO_GetBlockSize_FromBlockId()` is used during decompression to get size of next compressed block,

but here this code lies within `LZ4IO_compressFilename_extRess()`

programs/Makefile
```diff
@@ -39,9 +39,15 @@ RELEASE?= r131
 DESTDIR?=
 PREFIX ?= /usr/local
 CFLAGS ?= -O3
-CFLAGS += -std=c99 -Wall -Wextra -Wundef -Wshadow -Wcast-qual -Wcast-align -Wstrict-prototypes -pedantic -DLZ4_VERSION=\"$(RELEASE)\"
-FLAGS  := -I../lib $(CPPFLAGS) $(CFLAGS) $(LDFLAGS)
+CFLAGS += -std=c99 -Wall   -Wextra -Wundef -Wshadow -Wcast-qual -Wcast-align -Wstrict-prototypes -pedantic -DLZ4_VERSION=\"$(RELEASE)\"
+
+
+ifeq ($(shell uname --hardware-platform), i386)
+           CFLAGS += -march=prescott
```
Cyan4973: This is hardware specific.

Can you prove that `-march=prescott` will always be beneficial when `uname --hardware-platform` returns `i386` ?
What about AMD platforms ? What about `clang` ?

Sure enough, there are _some_ situations in which it will prove better, like probably for your platform.
That being said, setting this parameter as "default" (for i386) suppose it will be beneficial for the vast majority of systems concerned, a fairly strong claim that would deserve to be backed by good evidences.

Note that it is expected that some platforms will benefit from specific adjustments.
For such case, it's possible to build LZ4 like this : 
`CFLAGS="-O3 -march=prescott" make install`

This would basically perform the same as your patch, but remain in the domain of "system specific" settings.

Cyan4973:
Off topic, but I've been looking at your blog, where you present your interesting experiments with lz4 code.

I noted the following sentence : 
> By providing this maximum cache size other than the 16kb default value , the compressor shows a 4.24% increase in compression than the latest release. Then was a disappointing part, I quickly look into the latest release to make the same change over there too, but, unfortunately, there was no such piece of code to manage the memory cache size and no similar kind of code. 

Well, the piece of code is present, look here :
https://github.com/Cyan4973/lz4/blob/master/lib/lz4.h#L66

It changed place, from `lz4.c` to `lz4.h`, in order to be readable by library users (in `lz4.c`, it was only "private").

PR #201, opened 2016-04-18
cmake: Install liblz4.pc

cernekee:
This performs the same substitutions as lib/Makefile.


Cyan4973:
Yep, it's a good straightforward patch @cernekee .

PR #204, opened 2016-04-19
Alloc failure fixes

irwand:
Check ALLOCATOR() failure and bail out, or it will crash.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -731,6 +731,10 @@ int LZ4_compress_fast(const char* source, char* dest, int inputSize, int maxOutp
     void* ctxPtr = &ctx;
 #endif
 
+#if (HEAPMODE)
+    if (!ctxPtr) { return 0; }
+#endif
+
     int result = LZ4_compress_fast_extState(ctxPtr, source, dest, inputSize, maxOutputSize, acceleration);
 
 #if (HEAPMODE)
@@ -963,6 +967,10 @@ int LZ4_compress_destSize(const char* src, char* dst, int* srcSizePtr, int targe
     void* ctx = &ctxBody;
 #endif
 
+#if (HEAPMODE)
+    if (!ctx) { return 0; }
+#endif
+
     int result = LZ4_compress_destSize_extState(ctx, src, dst, srcSizePtr, targetDstSize);
 
 #if (HEAPMODE)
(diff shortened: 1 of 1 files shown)


Cyan4973:
That's very good point @irwand .

I'm actually surprised such an issue is still present in the code. So your patch is welcomed.
I guess it's hidden by the fact that almost everyone uses stack mode instead.

There is just a very minor issue with this patch : 
it doesn't compile on VC and other non-C99 compilers.
Reason is : all declarations must be done _first_ , _before_ any statement.
Here `if (!ctx) { return 0; }` stands before `int result = `.

I recommend testing compilation with `-std=c99` turned off, to check this.

Solution can be trivial : enclose new variable declaration and usage into their own scope `{ }`.

Regards

irwand:
Thank you for that feedback @Cyan4973,
I kept it simple and just declared result first. I couldn't reproduce the compile problem on my end. I tried gcc -std=c90 -ansi and gcc still didn't complain about that.

Btw, there are some '//' comments in lz4.c that could be problematic for pedantic C compilers.
If they were to be fixed, it should be on separate change potentially.
Have a great day!

t-mat:
> I tried gcc -std=c90 -ansi and gcc still didn't complain about that.

Hi @irwand,  Just an FYI, `-pedantic` shows desired warning.

```
/*
gcc -std=c90 -pedantic test.c
test.c: In function 'main':
test.c:13:9: warning: ISO C90 forbids mixed declarations and code [-Wpedantic]
         int x = 0;
         ^
 */
#include <stdio.h>

int main(int argc, const char** argv) {
    {
        printf("hello\n");
        int x = 0;
        printf("x = %d\n", x);
    }

    {
        int y = 0;
        printf("hello again\n");
        printf("y = %d\n", y);
    }
    return 0;
}
```

t-mat:
> Btw, there are some '//' comments in lz4.c that could be problematic for pedantic C compilers.

Good point.  We may have to add something like `c89test` (`CFLAGS=-std=c89 -pedantic`) in our `Makefile` and `.travis.yml` to avoid such potential problems.

irwand:
Hi @t-mat 
I did try -pedantic, but it had so many complaints about ISO C90 does not support 'long long', so I thought we can't use that anyway, and didn't pay close attention to its output.. :)

I tried again with my latest change in this branch, disabling long long warning:

`~/src/lz4$ gcc -c lib/lz4.c -std=c89 -pedantic -Wall -Wno-long-long -o /tmp/lz4.o`
`lib/lz4.c: In function ‘LZ4_isLittleEndian’:`
`lib/lz4.c:162:54: error: C++ style comments are not allowed in ISO C90`
`     const union { U32 i; BYTE c[4]; } one = { 1 };   // don't use static : performance detrimental`

`lib/lz4.c:162:54: error: (this will be reported only once per input file)`
`lib/lz4.c: At top level:`
`lib/lz4.c:209:8: warning: extra tokens at end of #endif directive [-Wendif-labels]`
` #endif // LZ4_FORCE_MEMORY_ACCESS`

Looks like we're almost there, just need to change those // -- that should be done separately.

Is there anything else on this patch that I need to work on to get this merged?
Thank you!

t-mat:
> Is there anything else on this patch that I need to work on to get this merged?

This PR looks good for me. How about you @Cyan4973 ?

note: LZ4 project should tackle CI test for `-std=c89` as another issue.

Cyan4973:
Apologies for limited availability these past weeks.
Yes, it's a good patch.

And I also agree to open an item for `c89` tests on CI environment.

PR #211, opened 2016-05-20
port lz4 to tms570 (Texas Instuments Compiler, BE), stm32f4 (IAR)

vvromanov:
Port include only lz4.c & lz4.h
- Embedding system usually not use dynamic memory. I hide all dynamic memory calls under HEAPMODE #ifdef
- Add LZ4_DEPRECATED definition for TI compiler. 
- Disable warnings "variable "val" is used before its value is set"
- Disable warning "shift count is too large". On this systems size_t is 32 bit and shift it 32 bites right useless. Maybe you need handle this case
- Add pragmas to temporary disable  missing-prototypes warning

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -43,6 +43,7 @@
  */
 #define HEAPMODE 0
 
+
 /*
  * ACCELERATION_DEFAULT :
  * Select "acceleration" for LZ4_compress_fast() when parameter value <= 0
@@ -124,9 +125,11 @@
 /**************************************
 *  Memory routines
 **************************************/
+#if (HEAPMODE)
 #include <stdlib.h>   /* malloc, calloc, free */
 #define ALLOCATOR(n,s) calloc(n,s)
 #define FREEMEM        free
+#endif /* HEAPMODE */
 #include <string.h>   /* memset, memcpy */
 #define MEM_INIT       memset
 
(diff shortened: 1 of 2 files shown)


Cyan4973:
Thanks for the patch and interesting suggestions.
It seems there are a number of places where we can probably improve the code to make this portability easier.

lib/lz4.c
```diff
@@ -124,9 +125,11 @@
 /**************************************
 *  Memory routines
 **************************************/
+#if (HEAPMODE)
```
Cyan4973: If there is a need to totally remove any reference to `<stdlib.h>`, then it will have to be done differently.
A few functions depend on them (typically `LZ4_createXXXX`). No `malloc()` means no way to use them.

lib/lz4.c
```diff
@@ -188,17 +191,38 @@ static void LZ4_write16(void* memPtr, U16 value) { ((unalign*)memPtr)->u16 = val
 
 static U16 LZ4_read16(const void* memPtr)
 {
+#ifdef __TI_COMPILER_VERSION__
+    /* Suppess warning: #551-D variable "ctx" is used before its value is set*/
```
Cyan4973: Maybe the action here would be to remove the warning.
Since we never met it during our tests, maybe some information on the test conditions would be useful.

lib/lz4.c
```diff
@@ -333,7 +357,20 @@ static unsigned LZ4_NbCommonBytes (register size_t val)
             return (__builtin_clzll((U64)val) >> 3);
 #       else
             unsigned r;
+/* Disable warning "shift count is too large" */
```
Cyan4973: Classical issue.
This source code path is supposed to be only triggered in 64 bits mode.
So it will never generate any binary code on a 32-bits target.
Nonetheless, a compiler may complain, without realizing this is dead code.

We can probably get rid of it without these `#pragma`. A cast typically.

lib/lz4.c
```diff
@@ -1115,6 +1172,9 @@ int LZ4_compress_fast_continue (LZ4_stream_t* LZ4_stream, const char* source, ch
 }
 
 
+#ifdef __GNUC__
+#pragma GCC diagnostic ignored "-Wmissing-prototypes"
```
Cyan4973: These missing prototypes are intentional.
These functions are supposed to be used only by test tools.

I wish we could find a way to solve that without relying too much on pragma.

lib/lz4.c
vvromanov: <stdlib.h> can be moved outside of #if. If some function in project use malloc or free in debug build, compiler include all heap management stuff including special segment for heap etc.. In this case I get linking warnings.

t-mat: @vvromanov, I recommend you to fix this problem first.  Because currenly [all tests are failed](https://travis-ci.org/Cyan4973/lz4/jobs/131685995) by this problem.
Since [lib/lz4hc.c #includes lib/lz4.c](https://github.com/Cyan4973/lz4/blob/1f47f3f7479c6986b6794f9ad7f13f66ecadc4ff/lib/lz4hc.c#L69), you should also take caring about it.

lib/lz4.c
vvromanov: This is just compilation warning. I use arm_15.12.2.LTS compiler. You can download this compiler (it is bundled inside Code Composer Studio) http://processors.wiki.ti.com/index.php/Download_CCS

t-mat:
Thanks for reposting @vvromanov :)

Since your patch contains many good points which provide wider compatibility, we really appreciate your patch.  But currently your patch causes another compatibility issue.  I think we'll fix these issues and find a middle way of them.
I suppose it takes for a while to merge, but please hang on it!

Cyan4973:
I'll close this PR, as the code has changed significantly since.

Still, if someone is willing to make lz4 TI compatible, it will be a welcomed PR.

PR #214, opened 2016-06-04
Add LZ4F_disableChecksum() function

jonathanunderwood:
This PR adds a variant of a function taken from the lz4tools project[1] to disable checksum checking. This was added to the lz4tools project to allow blocks in a lz4 file to be read out of order.

The lz4tools project was written by @darkdragn

[1] https://github.com/darkdragn/lz4tools

### DIFF
--- a/lib/lz4frame.c
+++ b/lib/lz4frame.c
@@ -1477,3 +1477,16 @@ size_t LZ4F_decompress(LZ4F_decompressionContext_t decompressionContext,
     *dstSizePtr = (dstPtr - dstStart);
     return nextSrcSizeHint;
 }
+
+
+
+/* LZ4F_disableChecksum()
+* Call this function to allow you to read compressed blocks out of order.
+*/
+void LZ4F_disableChecksum(LZ4F_decompressionContext_t decompressionContext)
+{
+    LZ4F_dctx_t* dctxPtr = (LZ4F_dctx_t*)decompressionContext;
+
+    dctxPtr->frameInfo.contentChecksumFlag = LZ4F_noContentChecksum;
+}
+
--- a/lib/lz4frame.h
+++ b/lib/lz4frame.h
@@ -297,6 +297,10 @@ size_t LZ4F_decompress(LZ4F_decompressionContext_t dctx,
  * After a frame is fully decoded, dctx can be used again to decompress another frame.
  */
 
+void LZ4F_disableChecksum(LZ4F_decompressionContext_t decompressionContext);
+/* LZ4F_disableChecksum()
+* Call this function to allow you to read compressed blocks out of order.
+*/
 
 #if defined (__cplusplus)
 }


jonathanunderwood:
I wonder if I should also add a LZ4F_enableChecksum() function too ?

t-mat:
It would be nice to take `LZ4F_contentChecksum_t` as argument:

```
void LZ4F_setContentChecksum(
      LZ4F_decompressionContext_t decompressionContext
    , LZ4F_contentChecksum_t contentChecksum);
```

But I'm realized that what is actual purpose of this function?
Since LZ4F is designed as automatic (de)compression APIs, I'm not sure why you (lz4tools) need such override functionality and how safely access lz4-frames.

Cyan4973:
Sorry for the delay, my availability is very limited these days.

If I do understand your proposal correctly, you want the decoder to not process the checksum even if one is defined in the frame.

Fair enough. Indeed, this option doesn't exist yet in `LZ4F`.
`LZ4F_decompressOptions_t` structure was created for such purpose, although it's sent with each request, instead of being triggered only once.

But to understand if it's the right direction, I'm wondering about its usage.

The main benefit I can see from disabling checksum checking is reduced CPU usage.

But I don't see the relation with "reading block out of order" inside a frame.
First, "out of order" is only possible with independent blocks, so it's not applicable to all frames.
Second, it's still necessary to seek into the frame to get the correct block, or the correct set of block.

That's something the format allows, but `LZ4F` doesn't offer anything for this.

Hence I don't see how `LZ4F` could be used for such scenario.

Moreover, using a frame format designed for streaming for such usage is kinda inefficient.
Streaming and "random access" have very different patterns and limitation.
They each deserve their own optimizations.

For a "random access oriented" format, there is @carlomedas' [4mc](https://github.com/carlomedas/4mc) that may be more suitable.

jonathanunderwood:
It's usage can be seen here:

https://github.com/darkdragn/lz4tools/blob/master/lz4tools/lz4file.py

My crude understanding is that this then allows a tarred and lz4'd file to be seekable, i.e. to allow you to pull out one or two files from the middle of the file.

But I'm not the lz4tools author - @darkdragn is and maybe able to comment further. 

In an email to me he described it thusly: "i_t was something I added, for the purpose of making my lz4file objects seekable. With it enabled, you are required to read the blocks for a compressed file in order, or it throws an error, so that was my little work around. On first read through I would leave it enabled, to build the dictionary of decompressed block sizes, the disable the Checksum form the decompression context, so I can read the blocks out of order. (This is going to sound odd, but there are 2 checksums. An over all that is compounded through out the file, and one per block. That only disables the compounded checksum.)_"

Cyan4973:
I believe "random access" scenarios would be better served by a custom format and API, such as @carlomedas 's 4mc. In which case, there is no reliance on LZ4 frame format (or each block could also become a single independent block).
And anyway, streaming API does not offer any function to allow random access.

PR #218, opened 2016-06-21
Updating incorrect hyperlink to lz4 frame format

mandliya:
Just a rename. Incorrect file name here gives 404 error.


Cyan4973:
Thanks @mandliya .
Issue is already fixed in the "dev" branch.

Note, in case of future PR, that our policy requires contributions to stage first through "dev".
See ["branch policy"](https://github.com/Cyan4973/lz4#lz4---extremely-fast-compression)

PR #219, opened 2016-07-07
Clean up links, repository-wide

BenWiederhake:
I noticed that a few `http://...` links could be replaced by `https://...` links, which are usually deemed "safer".  Also, the self-reference is broken, as it still points to a googlecode repository.  This raises a few warnings when used downs stream ([example](https://github.com/tgingold/ghdl/pull/117)).


Cyan4973:
These are good points @BenWiederhake . Thanks for the PR !

PR #221, opened 2016-07-25
Fix build error with MinGW on Linux

kou:
Linux uses case sensitive file system. So "Windows.h" doesn't exist.

    % make CC=x86_64-w64-mingw32-gcc EXT=.exe SHARD_EXT=dll
    ...
    lz4io.c:69:73: fatal error: Windows.h: No such file or directory
    compilation terminated.
    ...


Cyan4973:
good point @kou !

Could you redirect your PR into "dev" branch ?
Direct PR to master are not permitted. See [branch policy](https://github.com/Cyan4973/lz4#lz4---extremely-fast-compression) .

kou:
Sorry. I recreated a pull request as #222.
(I did the same thing before... Sorry...)

PR #226, opened 2016-08-27
Fix POSIX portability and cmake file

gsauthof:
Main changes:

- request POSIX.2001 definitions in a portable fashion
- fix compile on Solaris 10
- cmake build file fixes

cf. the commit messages for details

### DIFF
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -49,7 +49,8 @@
 #  pragma warning(disable : 4127)      /* disable: C4127: conditional expression is constant */
 #endif
 
-#define _POSIX_SOURCE 1        /* for fileno() within <stdio.h> on unix */
+/* cf. http://man7.org/linux/man-pages/man7/feature_test_macros.7.html */
+#define _XOPEN_VERSION 600 /* POSIX.2001, for fileno() within <stdio.h> on unix */
 
 
 /****************************
--- a/cmake_unofficial/CMakeLists.txt
+++ b/cmake_unofficial/CMakeLists.txt
@@ -92,21 +92,23 @@ endif()
 #warnings
 
 if(MSVC)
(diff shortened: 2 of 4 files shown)


programs/lz4cli.c
```diff
@@ -49,7 +49,7 @@
 #  pragma warning(disable : 4127)      /* disable: C4127: conditional expression is constant */
 #endif
 
-#define _POSIX_SOURCE 1        /* for fileno() within <stdio.h> on unix */
+#define _XOPEN_VERSION 600 /* POSIX.2001, for fileno() within <stdio.h> on unix */
```
Cyan4973: Could you please provide a source which justify the betterness of this detection macro ?

Cyan4973:
Mostly good !
Just one request for source regarding a detection macro

programs/lz4cli.c
gsauthof: @Cyan4973 - yes, I can, the source is included in the commit message:

    request POSIX.2001 definitions for fileno()
    
    this also fixes compilation on Solaris 10
    
    cf. test_macros(7) Linux man page:
    
    > _POSIX_SOURCE
    >   Defining this obsolete macro with any value is equivalent to  defin‐
    >   ing _POSIX_C_SOURCE with the value 1.
    >
    >   Since  this macro is obsolete, its usage is generally not documented
    >   when discussing feature test macro requirements in the man pages.

Cyan4973: Sorry, I meant a link to the authoritative source.
For example : http://man7.org/linux/man-pages/man7/feature_test_macros.7.html

It can be quite useful to add such link into code comments, so that future reviewers are aware of the choice's basis.

gsauthof:
@Cyan4973 - ok, I added the link to that man page as code comment and rebased the PR.

PR #233, opened 2016-09-03
Appveyor tests 

inikep:
- works with VS2010 and newer
- fixed VS and MinGW compilation


Cyan4973:
great !

inikep:
This should be done a long time ago :)

PR #234, opened 2016-09-03
Mem.h from zstd 0.8.1

inikep:
It makes MEM functions easier to maintain.


Cyan4973:
arf, sorry,
the intention is good,
but it is important that `lz4.c` remains a single-file project, with only `lz4.h` as dependency.

inikep:
I just thought about it and that is why it's in a separate branch.

PR #238, opened 2016-09-22
Add LZ4LIB_API

kou:
It's based on Zstandard's ZSTDLIB_API.

See also: https://github.com/Cyan4973/lz4/issues/216#issuecomment-226245432

Deprecated functions aren't LZ4LIB_API targets. Because we don't need to
export deprecated functions from now.

There are same LZ4LIB_API definitions in each header files instead of
including a common header file because LZ4_DEPRECATED is defined so.

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -43,7 +43,7 @@ LIBVER := $(shell echo $(LIBVER_SCRIPT))
 
 DESTDIR?=
 PREFIX ?= /usr/local
-CPPFLAGS= -DXXH_NAMESPACE=LZ4_
+CPPFLAGS= -DXXH_NAMESPACE=LZ4_ -DLZ4_DLL_EXPORT=1
 CFLAGS ?= -O3
 CFLAGS += -Wall -Wextra -Wundef -Wshadow -Wcast-align -Wcast-qual -Wstrict-prototypes
 FLAGS   = $(CPPFLAGS) $(CFLAGS) $(LDFLAGS)
--- a/lib/lz4.h
+++ b/lib/lz4.h
@@ -48,6 +48,23 @@ extern "C" {
  * A library is provided to take care of it, see lz4frame.h.
 */
 
+/*-***************************************************************
+*  Export parameters
+*****************************************************************/
+/*!
+*  LZ4_DLL_EXPORT :
+*  Enable exporting of functions when building a Windows DLL
+*/
+#if defined(_WIN32)
+#  if defined(LZ4_DLL_EXPORT) && (LZ4_DLL_EXPORT==1)
+#    define LZ4LIB_API __declspec(dllexport)
+#  else
+#    define LZ4LIB_API __declspec(dllimport)
+#  endif
+#else
+#  define LZ4LIB_API
+#endif
+
 /*-************************************
 *  Version
 **************************************/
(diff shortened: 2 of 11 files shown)


lib/Makefile
```diff
@@ -43,7 +43,7 @@ LIBVER := $(shell echo $(LIBVER_SCRIPT))
 
 DESTDIR?=
 PREFIX ?= /usr/local
-CPPFLAGS= -DXXH_NAMESPACE=LZ4_
+CPPFLAGS= -DXXH_NAMESPACE=LZ4_ -DLZ4_DLL_EXPORT=1
```
Cyan4973: So it's enabled by default in `Makefile` ?
I'm a bit surprised, as I was expecting this feature to be Visual only.

lib/lz4.h
```diff
@@ -48,6 +48,23 @@ extern "C" {
...
 */
 
+/*-***************************************************************
+*  Export parameters
+*****************************************************************/
+/*!
+*  LZ4_DLL_EXPORT :
+*  Enable exporting of functions when building a Windows DLL
+*/
+#if defined(_WIN32)
```
Cyan4973: How does that work with mingw ? (since it's enabled by default in `Makefile`)

lib/lz4frame.h
```diff
@@ -49,14 +49,30 @@ extern "C" {
...
+/*-***************************************************************
+*  Export parameters
+*****************************************************************/
+/*!
+*  LZ4_DLL_EXPORT :
+*  Enable exporting of functions when building a Windows DLL
+*/
+#if defined(_WIN32)
+#  if defined(LZ4_DLL_EXPORT) && (LZ4_DLL_EXPORT==1)
+#    define LZ4LIB_API __declspec(dllexport)
```
Cyan4973: Since the macro `lz4.h` and `lz4frame.h` macros have same name,
there is a risk of redefinition warning in case both are included in same project.
It feels safer to have 2 different names,
for example `LZ4FLIB_API` for this file

lib/lz4hc.h
```diff
@@ -44,6 +44,23 @@ extern "C" {
...
 
+/*-***************************************************************
+*  Export parameters
+*****************************************************************/
+/*!
+*  LZ4_DLL_EXPORT :
+*  Enable exporting of functions when building a Windows DLL
+*/
+#if defined(_WIN32)
+#  if defined(LZ4_DLL_EXPORT) && (LZ4_DLL_EXPORT==1)
```
Cyan4973: same comment regarding macros redefinitions

lib/lz4.h
kou: MinGW also support `dllexport`/`dllimport`: http://www.mingw.org/wiki/sampledll

If `dllexport` isn't used, all functions are exported by default. If `dllexport` is used, functions marked as `dllexport` are only exported.

For example, `LZ4_XXH32` is also exported when we don't use `dllexport`.

I think that there is no problem that we use `dllexport` with MinGW.

lib/Makefile
kou: `dllexport` is required in Visual C++ but not required in MinGW.

I enable it for MinGW to use `dllimport`. If we use `dllimport`, we need to use the current approach or the following approach:

```c
#if defined(_WIN32) && !defined(__MINGW32__)
#  if defined(LZ4_DLL_EXPORT) && (LZ4_DLL_EXPORT==1)
#    define LZ4LIB_API __declspec(dllexport)
#  else
#    define LZ4LIB_API __declspec(dllimport)
#  endif
#else
#  define LZ4LIB_API
#endif
```

`dllimport` isn't required. It's just for optimization: https://blogs.msdn.microsoft.com/russellk/2005/03/20/lnk4217/

> it is purely a performance optimization

If we don't use `dllimport`, we can use more simple definition like Zstd:

```c
#if defined(_WIN32) && defined(LZ4_DLL_EXPORT) && (LZ4_DLL_EXPORT==1)
#  define LZ4LIB_API __declspec(dllexport)
#else
#  define LZ4LIB_API
#endif
```

If you prefer to remove `dllimport`, i"ll do it.

lib/lz4frame.h
kou: I've renamed to `LZ4FLIB_API`.

lib/lz4hc.h
kou: I've renamed to `LZ4HCLIB_API`.

Cyan4973:
Looks good to me

PR #239, opened 2016-09-22
Use lower case for header file name

kou:
"WinIoCtl.h" works on Windows because Windows uses case insensitive file
system. But it doesn't work on GNU/Linux (for cross compiling with
MinGW) because GNU/Linux uses case sensitive file system.


Cyan4973:
Good point !

PR #241, opened 2016-09-23
add a flag to disable static libs

vapier:


Cyan4973:
better option would be to separate dynamic and static library generation,
hence creating ability to target one or the other.

Also, separating static and dynamic library generation is one thing,
removing static library from installation is another one.

Considering expectation from LZ4 package, static library installation should remain the default choice.
Removing it shall be "opt in" at this stage.

It also needs to be properly tested (add new test cases).

vapier:
my commit doesn't change the default behavior -- static libs are still built & installed if the user does nothing.  only if they explicitly request it be disabled will it be so.

not everyone wants to install the static libs.  i added this explicitly because in Gentoo we make the build & install optional (with the default being off).

Cyan4973:
OK, I mistakenly understood otherwise.
In which case, the PR is not intrusive, so it's a good start.

There is a test error, but it might related to the test system rather than the patch. Let me have a look ...

vapier:
i can respin the patch to create new targets, and then have the knob control which targets get used by default

i wasn't sure on the preferred style (naming/etc...) as i couldn't find an example already in the code base like it.  if you want diff naming/etc..., just lemme know

vapier:
thanks!

PR #248, opened 2016-10-25
Fix broken frameCompress.c example

Leandros:
The example to show frame compression was broken, due to not writing out the header [here](https://github.com/lz4/lz4/blob/dev/examples/frameCompress.c#L55).
This is fixing it.

### DIFF
--- a/examples/frameCompress.c
+++ b/examples/frameCompress.c
@@ -53,6 +53,14 @@ static int compress_file(FILE *in, FILE *out, size_t *size_in, size_t *size_out)
 	}
 
 	printf("Buffer size is %zu bytes, header size %zu bytes\n", size, n);
+	k = fwrite(buf, 1, offset, out);
+	if (k < offset) {
+		if (ferror(out))
+			printf("Write failed");
+		else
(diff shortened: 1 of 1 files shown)


Cyan4973:
I don't think it was broken.

`LZ4F_compressUpdate()` only starts writing in output buffer starting from `offset` position.

https://github.com/Leandros/lz4/blob/0dda1e32f7c61704e12a5c44710a3609cc8a1c92/examples/frameCompress.c#L71

From `0` to `offset` is where the frame header is.
The write operation starts from `0` (`buff`), so it doesn't miss the header : 
https://github.com/Leandros/lz4/blob/0dda1e32f7c61704e12a5c44710a3609cc8a1c92/examples/frameCompress.c#L82

As a matter of fact, the proposed version fails automated tests : 
https://travis-ci.org/lz4/lz4/jobs/170576314

Leandros:
You're absolutely right!
I copied the version and somehow the `offset = 0;` slipped past the inner braces and got reset after the first iteration, missing to write out the header.

PR #250, opened 2016-11-03
Fix test mode and write to null

### LINKED ISSUE #199: lz4 -t always succeeds on any file
Running lz4 -t on any readable file reports success and exits with status 0.  This is unexpected.  I think lz4 should complain if it is passed a file that is not in lz4 format.  The -t option is supposed to "test compressed file integrity" per the help screen, however it does no such test AFAIKT.

normg@bunk>lz4 -t /usr/bin/ssh
Successfully decoded 429936 bytes
normg@bunk>ls -l /usr/bin/ssh
-rwxr-xr-x 1 root root 429936 Aug 17  2015 /usr/bin/ssh

Norm Green

terrelln:
Fix test mode to not always return success.
Don't ask for permission to overwrite `nulmark`.

Another option, `g_passThrough` is necessary because if the user passes `-t`, we shouldn't silently ignore errors, even if `-f` is passed as well.  Thanks to @t-mat for this part of the patch.

Fixes #199 

Before:

    > echo "hello world" > file
    > lz4 -t file
    successfully decoded 12 bytes
    > lz4 -tf file
    successfully decoded 12 bytes
    > lz4 file null
    Warning : /dev/null already exists
    Overwrite ? (Y/n) : n
    Error 12 : No. Operation aborted : /dev/null already exists
    > lz4 file /dev/null
    Warning : /dev/null already exists
    Overwrite ? (Y/n) : n
    Error 12 : No. Operation aborted : /dev/null already exists

After:

    > lz4 -t file
    Error 44 : Unrecognized header : file cannot be decoded
    > lz4 -tf file
    Error 44 : Unrecognized header : file cannot be decoded
    > lz4 file null
    Compressed 12 bytes into 31 bytes ==> 258.33%
    > lz4 file /dev/null
    Compressed 12 bytes into 31 bytes ==> 258.33%

### DIFF
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -393,7 +393,7 @@ int main(int argc, const char** argv)
                 case 'c': forceStdout=1; output_filename=stdoutmark; displayLevel=1; break;
 
                     /* Test integrity */
-                case 't': decode=1; LZ4IO_setOverwrite(1); output_filename=nulmark; break;
+                case 't': decode=1; LZ4IO_setPassThrough(0); output_filename=nulmark; break;
 
                     /* Overwrite */
                 case 'f': LZ4IO_setOverwrite(1); break;
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -393,7 +393,7 @@ int main(int argc, const char** argv)
                 case 'c': forceStdout=1; output_filename=stdoutmark; displayLevel=1; break;
 
                     /* Test integrity */
-                case 't': decode=1; LZ4IO_setPassThrough(0); output_filename=nulmark; break;
+                case 't': decode=1; LZ4IO_setTestMode(1); output_filename=nulmark; break;
 
                     /* Overwrite */
                 case 'f': LZ4IO_setOverwrite(1); break;
--- a/programs/lz4io.c
+++ b/programs/lz4io.c
@@ -136,6 +136,7 @@ static clock_t g_time = 0;
 *  Local Parameters
 **************************************/
 static int g_overwrite = 1;
+static int g_passThrough = 1;
 static int g_blockSizeId = LZ4IO_BLOCKSIZEID_DEFAULT;
 static int g_blockChecksum = 0;
 static int g_streamChecksum = 1;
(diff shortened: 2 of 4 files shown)


Cyan4973:
Good for most of it.
Just the last test makes me wonder.

`lz4 file /dev/null`
I read it as "write compressed file _as_ `/dev/null`".

This is very different from writing _into_ `/dev/null`, such as : 
`lz4 file -c > /dev/null`.

The difference is that, given the right privilege (root),
this command may _erase_ `/dev/null` (and replaced it with a normal file).

One can make a test to check that with `sudo lz4 file /dev/null`,
though beware that a non-functioning `/dev/null` can lead to an unstable system.
Better do that into a fresh sandbox.

I believe `lz4 file /dev/null` should fail, due to insufficient rights.

Cyan4973:
Some notes on limit test cases.

Be `text` a readable, uncompressed, text file :

```
gzip -t   text    # fail
gzip -tf  text    # displays text
gzip -d   < text    # fail
gzip -df  < text    # displays text

xz -t   text    # fail
xz -tf  text    # fail
xz -d   < text    # fail
xz -df  < text    # fail
```

The difference is that `xz` doesn't support "pass-through" mode when using `-df`, a mode which was added to use `zcat` transparently on both uncompressed and compressed files.
Though it seems gzip also displays on `-tf`, which is kind of weird.

So here would be my preferred behavior : 

```
lz4 -t   text    # fail
lz4 -tf  text    # fail
lz4 -d   < text    # fail
lz4 -df  < text    # displays text
```

To fix too : relevant test cases, which may currently expect a `gzip` behaviour.

terrelln:
It is safe to `fopen('/dev/null', 'w')`, and lz4 currently won't fail on `lz4 -f text /dev/null` or `sudo lz4 -f text /dev/null`, before or after this patch.  It would be unsafe if lz4 called `remove("/dev/null")` however.  `zstd Makefile -o /dev/null` does not ask if we want to overwrite.

Fixing the tests now.

Cyan4973:
> It would be unsafe if lz4 called `remove("/dev/null")` however

Good point.

PR #257, opened 2016-11-07
added --rm option

inikep:
added --rm option
Copy owner, file permissions and modification time
fixed appveyor tests

### DIFF
--- a/lib/lz4frame.h
+++ b/lib/lz4frame.h
@@ -66,7 +66,7 @@ extern "C" {
 #endif
 
 #if defined(_MSC_VER)
-#  define LZ4F_DEPRECATE(x) __declspec(deprecated) x
+#  define LZ4F_DEPRECATE(x) x   /* __declspec(deprecated) x - only works with C++ */
 #elif defined(__clang__) || (defined(__GNUC__) && (__GNUC__ >= 6))
 #  define LZ4F_DEPRECATE(x) x __attribute__((deprecated))
 #else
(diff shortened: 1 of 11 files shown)


lib/lz4frame.h
```diff
@@ -274,7 +274,7 @@ typedef struct {
  * That is, it should be == 0 if decompression has been completed fully and correctly.
  */
 LZ4FLIB_API LZ4F_errorCode_t LZ4F_createDecompressionContext(LZ4F_dctx** dctxPtr, unsigned version);
-LZ4FLIB_API LZ4F_errorCode_t LZ4F_freeDecompressionContext(LZ4F_dctx* dctx);
+LZ4FLIB_API LZ4F_errorCode_t LZ4F_freeDecompressionContext(LZ4F_dctx* const dctx);
```
Cyan4973: What's the objective of this `const` statement ?
As part of the prototype definition, it's supposed to be useless (while it can still be somewhat useful in the implementation).

inikep: I added it only to match function definition what was signaled by Visual Studio.

visual/VS2010/lz4/lz4.vcxproj
```diff
@@ -28,24 +28,24 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <CharacterSet>Unicode</CharacterSet>
+    <CharacterSet>MultiByte</CharacterSet>
```
Cyan4973: Could you give me a refresh of MultiByte vs Unicode, and why one is better than the other ?

PR #260, opened 2016-11-08
cmake: significant changes to accomodate a new "bundled" mode.

nemequ:
This makes it much easier to include LZ4 in a CMake-based project;
you only need to add_subdirectory(lz4/contrib/cmake_unofficial),
then you can use the lz4 target (which is the library).  In bundled
mode a static library will be built and nothing will be installed,
although the lz4 and lz4c command line tools will still be built.

This also makes some other much less intrusive changes, including:

 * Version information is now parsed from lib/lz4.h so the
   CMakeLists.txt file no longer needs to manually be kept up to date
   manually.
 * pkg-config file uses ${prefix} where appropriate instead of writing
   out full paths, which is important for people on Windows.
 * Warning flags are now checked instead of blindly added.
 * The GNUInstallDirs module is used to determine installation
   locations.
 * The static library generates position independent code, which means
   it's usable when building a shared library.
 * The CLI tools link to liblz4 instead of duplicating the object code.

### DIFF
--- a/contrib/cmake_unofficial/CMakeLists.txt
+++ b/contrib/cmake_unofficial/CMakeLists.txt
@@ -1,110 +1,157 @@
-PROJECT(LZ4 C)
-set(CPACK_PACKAGE_DESCRIPTION_SUMMARY "LZ4 compression library")
-set(CPACK_PACKAGE_VERSION_MAJOR 1)
-set(CPACK_PACKAGE_VERSION_MINOR 7)
-set(CPACK_PACKAGE_VERSION_PATCH 2)
-set(VERSION_STRING	" \"${CPACK_PACKAGE_VERSION_MAJOR}.${CPACK_PACKAGE_VERSION_MINOR}.${CPACK_PACKAGE_VERSION_PATCH}\" ")
-include(CPack)
+set(LZ4_TOP_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../..")
 
-cmake_minimum_required (VERSION 2.6)
-INCLUDE (CheckTypeSize)
-check_type_size("void *" SIZEOF_VOID_P)
-IF(SIZEOF_VOID_P STREQUAL  "8")
-    set (CMAKE_SYSTEM_PROCESSOR "64bit")
-    MESSAGE( STATUS "64 bit architecture detected size of void * is " ${SIZEOF_VOID_P})
-ENDIF()
+# Parse version information
+file(STRINGS "${LZ4_TOP_SOURCE_DIR}/lib/lz4.h" LZ4_VERSION_MAJOR REGEX "^#define LZ4_VERSION_MAJOR +([0-9]+) +.*$")
+string(REGEX REPLACE "^#define LZ4_VERSION_MAJOR +([0-9]+) +.*$" "\\1" LZ4_VERSION_MAJOR "${LZ4_VERSION_MAJOR}")
+file(STRINGS "${LZ4_TOP_SOURCE_DIR}/lib/lz4.h" LZ4_VERSION_MINOR REGEX "^#define LZ4_VERSION_MINOR +([0-9]+) +.*$")
+string(REGEX REPLACE "^#define LZ4_VERSION_MINOR +([0-9]+) +.*$" "\\1" LZ4_VERSION_MINOR "${LZ4_VERSION_MINOR}")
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks, it's a great patch @nemequ .

A very minor comment is that I don't know if PIC mode is the better setting for static library.
I know at least one build system which would not like it (but it wouldn't use `cmake` either, so it's not an immediate issue).
I also believe that's a very minor issue, and we can wait for feedback, see if there really is any need to modify such setting.

nemequ:
> A very minor comment is that I don't know if PIC mode is the better setting for static library.

My understanding is that It is required if you want to link it into a shared library.  *e.g.*, if you're trying to build `libfoo.so` (or `libsquash0.8-plugin-lz4.so` :wink:) which includes `liblz4.a`, if `liblz4.a` isn't position independent you'll end up with something like

```
/usr/bin/ld: lz4/contrib/cmake_unofficial/liblz4.a(lz4frame.c.o): relocation R_X86_64_32S against `.rodata' can not be used when making a shared object; recompile with -fPIC
```

So IMHO it should be on by default.  I don't really think it's necessary, but I'd be okay with adding an option to disable it:

```diff
diff --git a/contrib/cmake_unofficial/CMakeLists.txt b/contrib/cmake_unofficial/CMakeLists.txt
index 9324338..01af33e 100644
--- a/contrib/cmake_unofficial/CMakeLists.txt
+++ b/contrib/cmake_unofficial/CMakeLists.txt
@@ -74,12 +74,22 @@ set(LZ4_CLI_SOURCES
   "${LZ4_PROG_SOURCE_DIR}/lz4io.c"
   "${LZ4_PROG_SOURCE_DIR}/datagen.c")
 
+# Whether to use position independent code for the static library.  If
+# we're building a shared library this is ignored and PIC is always
+# used.
+option(LZ4_POSITION_INDEPENDENT_LIB "Use position independent code for static library (if applicable)" ON)
+if(LZ4_POSITION_INDEPENDENT_LIB OR BUILD_SHARED_LIBS)
+  set(LZ4_POSITION_INDEPENDENT_CODE TRUE)
+else()
+  set(LZ4_POSITION_INDEPENDENT_CODE FALSE)
+endif()
+
 # liblz4
 add_library(lz4 ${LZ4_SOURCES})
 set_target_properties(lz4 PROPERTIES
   SOVERSION "${LZ4_VERSION_STRING}"
   VERSION "${LZ4_VERSION_STRING}"
-  POSITION_INDEPENDENT_CODE TRUE)
+  POSITION_INDEPENDENT_CODE ${LZ4_POSITION_INDEPENDENT_CODE})
 
 # lz4
 add_executable(lz4cli ${LZ4_CLI_SOURCES})
```

If you want it off by default it should definitely be configurable.

BTW, I'm not sure how comfortable you are with CMake, but I'm willing to help with maintenance… just mention me on an issue if you need something.

Cyan4973:
> If you want it off by default it should definitely be configurable.

I have no definitive position on the matter.
I just remember that when I tried to changed the static library compilation setting to `fPIC` I was burned by an internal Facebook build system which did not like it. So I kept it that way ever since.
I can understand that your use case is opposite, and building a dynamic library from the static one actually requires `fPIC`.
The way to make it configurable with `make` is to put such flag into `LDFLAGS`, this way they can be easily changed on the command line.

I'm all fine with the solution you propose. Having `fPIC` on by default is not a concern either, since the build system I'm referring to will not use `cmake` anyway.
You propose a way to make it configurable, so it feels all good.

> I'm willing to help with maintenance… just mention me on an issue if you need something.

Great !
I believe it would be fair to add your name into the `cmake` header (along a shortened OS license), so that people know you are the actual author and maintainer of this new edition.

nemequ:
Intersting; I thought PIC was pretty safe.

>  The way to make it configurable with make is to put such flag into LDFLAGS, this way they can be easily changed on the command line.

Yeah, you can do that with CMake too, but it's not really considered idiomatic… using the `POSITION_INDEPENDENT_CODE` option makes it easier to switch between different compilers/linkers where the flags may be different (i.e., MSVC).

Do you want me to file a PR with that patch from earlier?

>  I believe it would be fair to add your name into the `cmake` header (along a shortened OS license), so that people know you are the actual author and maintainer of this new edition.

*Shrug*.  If you want that's fine, but I'm guessing people will just file issues/PRs here not mailing the maintainer directly so it probably doesn't matter.  Please be aware, though, that I'm not planning on watching the project (LZ4 is very active, too much noise), so just mention me or assign the issue to me (or send me an e-mail, poke me on IRC, etc.).

For the license, if anything I'd prefer to just use a public domain dedication.  I've been working on putting together similar patches for lots of compression libraries so a lot of this code is shared across several projects, something else could get messy.

Cyan4973:
ok, on all points

PR #264, opened 2016-11-10
Dev

inikep:


Cyan4973:
An excellent batch of improvements

programs/lz4cli.c
```diff
@@ -417,20 +419,33 @@ int main(int argc, const char** argv)
...
+                                if (B < 4) badusage(exeName);
+                                if (B <= 7) {
+                                    blockSize = LZ4IO_setBlockSizeID(B);
+                                    BMK_SetBlockSize(blockSize);
+                                    DISPLAYLEVEL(2, "using blocks of size %u KB \n", (U32)(blockSize>>10));
+                                } else {
+                                    if (B < 32) badusage(exeName);
+                                    BMK_SetBlockSize(B);
+                                    if (B >= 1024) {
+                                        DISPLAYLEVEL(2, "bench: using blocks of size %u KB \n", (U32)(B>>10));
```
terrelln: This can be printed in non-benchmark mode, which could be confusing.  Should we push this message this back until we know whether we are benchmarking or not?

Cyan4973: That's a good point @terrelln .
Ideally, such message should be displayed only on entering the section which triggers benchmark mode (see line #535).

inikep: > This can be printed in non-benchmark mode, which could be confusing. 
I would like to know that I'm using options that will be ignored or used only in benchmark mode.

The questions here are:
a) shouldn't we print all override of options e.g. `-B100D200X300b` ?
b) shouldn't we print options that will be provided but ignored e.g. `-B100` without `-b` ?

Cyan4973: We don't have a clearly defined behaviour for such cases.
I guess a first task would be to define one.

PR #265, opened 2016-11-11
Fix license and remove references to zstd

terrelln:
@inikep you intended to use the regular BSD license right?


Cyan4973:
pretty good catches 👍

inikep:
Good catches and thanks for fixes

PR #279, opened 2016-11-27
Add "man" target to Makefile

t-mat:
Yay!  I finally make free time.  So I did trivial thing.

- `make man` makes manpage `lz4.1` from `lz4.1.md`
- It needs `ruby`, `gem` and `ronn`.  These are same as xxhash is required.

Open problems are

- `LIBVER_*` variables in the `Makefile` are duplicated.  It may harm future maintainability.
- Since `ronn` is no longer maintained, same kind of another tool such as `md2man` may be better choice.  But I use `ronn` for now.
- It would be nice if we output HTML manpage for lz4.org

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -27,6 +27,17 @@
 # lz4c32: Same as lz4c, but forced to compile in 32-bits mode
 # ##########################################################################
 
+# Version numbers
+LIBVER_SRC := ../lib/lz4.h
+LIBVER_MAJOR_SCRIPT:=`sed -n '/define LZ4_VERSION_MAJOR/s/.*[[:blank:]]\([0-9][0-9]*\).*/\1/p' < $(LIBVER_SRC)`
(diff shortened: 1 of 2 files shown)


Cyan4973:
It's a very welcomed PR @t-mat .

> LIBVER_* variables in the Makefile are duplicated. It may harm future maintainability.

Right. Sure, a single version would be better, but it's perfectly acceptable for a first PR.
Duplicating them looks like a fairly minor nuisance.

> Since ronn is no longer maintained, same kind of another tool such as md2man may be better choice. 

Right. Here also, no urgency,
we'll stick to the plan to deliver both `lz4.1` and `lz4.md` as part of the repository,
in order to not mandate user systems to have `ronn` pre-installed.
Which means, we only need to use `ronn` from time to time, and from controlled machines.

> It would be nice if we output HTML manpage for lz4.org

Good point. Maybe we could use the same trick as `*_format.md`,
that is, link the `.md` file directly from `lz4.org` ?

Cyan4973:
I note there are a few `<br/>` remaining here and there.
Is that intentional ?

PR #292, opened 2016-12-21
improved POSIX

inikep:


inikep:
Unicode is good for WinAPI applications but it's a bad choice for portable executables.
In our case Unicode/Multi-byte is used only for `UTIL_prepareFileList()` from `programs\util.h`. This function returns a file list which is used by compressor. The compressor's CLI uses standard C functions that will not accept Unicode names (it accepts `char*`). Therefore selecting Multi-byte does this conversion for us. The real conversion of lz4.exe to Multi-Byte was done in LZ4 v1.7.3. Here we convert only our testing programs.

josephsmeng:
why not just leave the unicode by default and just use the multi-bytes
version function FindFirstFileA?

inikep:
This is another possible solution.
Do you see any advantages of using Unicode for lz4.exe?
Please keep in mind that `liblz4` and `liblz4-dll` still use Unicode.

Cyan4973:
Is this PR ready for a merge, or are there still some points under discussion ?

inikep:
It's ready. I already introduced proposed changes.

PR #293, opened 2016-12-27
improved *BSD and Solaris compatibility

inikep:


lib/Makefile
```diff
@@ -112,10 +107,35 @@ clean:
...
+ifneq (,$(filter $(shell uname),Linux Darwin GNU/kFreeBSD GNU OpenBSD FreeBSD NetBSD DragonFly SunOS))
+
+ifneq (,$(filter $(shell uname),SunOS))
+INSTALL ?= ginstall
+else
+INSTALL ?= install
+endif
+
+ifneq (,$(filter $(shell uname),OpenBSD FreeBSD NetBSD DragonFly SunOS))
+PREFIX  ?= /usr
```
Cyan4973: This one is unexpected, though it could be SUN specific :
- `/usr` is for "distribution controlled" software. See it as the target directory for `apt-get install` on Ubuntu for example. The installer is also distribution controlled.
- while `/usr/local` is for "local installation", such as someone downloading lz4 from github, and running locally `make install`.

Therefore, the `Makefile` provided in the project directory should, by default, install to `/usr/local`. Sun (now Oracle) should be in charge of installing lz4 into `/usr` if they see it fit to belong to the distribution.

Now, it could also be that Sun convention is different.
In which case, a link to the place which explains it would be welcomed in the source code.

programs/Makefile
```diff
@@ -112,26 +108,48 @@ preview-man: clean-man man
...
 
+
+ifneq (,$(filter $(shell uname),SunOS))
+INSTALL ?= ginstall
+else
+INSTALL ?= install
+endif
+
+ifneq (,$(filter $(shell uname),OpenBSD FreeBSD NetBSD DragonFly SunOS))
+PREFIX  ?= /usr
```
Cyan4973: same comment as previously

Cyan4973:
Pretty good !
A single request, on making sure `/usr` is a valid `PREFIX` target for a local installation (which is the default `Makefile` scenario), as it seems to contradict "distrib vs local" binary separation which is generally in place for unix systems.

lib/Makefile
inikep: AFAIR there was no `/usr/local` at all. I checked `/usr/local` installation and `lz4` was not detected (not in default path). Then I installed `xz` and used the same paths that is `/usr/`. 

When you look at https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard
There is a sentence `Most of these directories exist in all UNIX operating systems and are generally used in much the same way; however, the descriptions here are those used specifically for the FHS, and are not considered authoritative for platforms other than Linux.`

inikep: For FreeBSD:
```
root@:~ # whereis lzma
lzma: /usr/bin/lzma /usr/share/man/man1/lzma.1.gz /usr/ports/archivers/lzma
root@:~ # whereis xz
xz: /usr/bin/xz /usr/share/man/man1/xz.1.gz
```

I tried to build `zip` from FreeBSD ports:
```
cd /usr/ports/archivers/zip
make install clean
```
what gives:
```
root@:/usr/ports/archivers/zip # whereis zip
zip: /usr/local/bin/zip /usr/local/man/man1/zip.1.gz /usr/ports/archivers/zip
```

inikep: I checked FreeBSD and `/usr/local` is possible but:
a) `/usr/local/share/man/man1` has to be changed to `/usr/local/man/man1`
b) `cc -llz4` doesn't detect `liblz4` but there is a lot of different libraries in this directory so I assume that `/usr/local/lib` has to be added manually to path

inikep: I checked Solaris and most additional packages prefer `/opt` instead of `/usr/local`. Moreover `lz4`, libraries, manual are not available in path. They can be accessed with:
`/usr/local/bin/lz4`
`man -M "/usr/local/man" lz4`

PR #296, opened 2017-01-02
Multiple CMake improvements

mgorny:
In order to get the unofficial CMake production-ready, a few fixes:

* [x] correct SOVERSION on shared library (assuming the one used in Makefiles is correct),
* [x] add `BUILD_STATIC_LIBS` option to support building both shared & static libs simultaneously,
* [x] install lz4cat and unlz4 symlinks,
* [x] install manpages.


Cyan4973:
Great patch @mgorny !

PR #306, opened 2017-01-16
Create .travis.ylm

stefb965:


Cyan4973:
Could you describe what your PR is expected to achieve ?

We already have a [`.travis.yml`](https://github.com/lz4/lz4/blob/dev/.travis.yml), it's unclear why we should add a `.travis.ylm`.

Cyan4973:
closed for lack of feedback

PR #313, opened 2017-02-06
Add reporting of position where LZ4 stream ends; Checking of macros

ehem:
Reporting how large the output is actually pretty low-value since simply looking at the output file size will tell you that.  In the case where a LZ4 stream has data added on to the end, knowing where the LZ4 stream ends is highly valuable, since it allows recovery of the appended data.  An example of this is Qualcomm-ARM Linux kernels, where hardware data is appended to a LZ4-compressed kernel image.

The implementation is _distinctly_ suboptimal, but sufficient for easy cases.  If input is a pipe, it will merely display "-1" as the point.  Really should have the various portions of the decoder keep track, but that is rather more involved.

The macro fixes were merely something staring me in the face when I was going after the above.  With these changes the various macros won't cause interesting misbehavior.  For example now the code
```
if (<some_condition>) DISPLAYUPDATE(val0|val1, "Something happened!\n");
else a_normal_function();
```
Will compile successfully and behave as expected without the special knowledge of DISPLAYUPDATE() being a macro.

### DIFF
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -83,13 +83,13 @@ static U32 g_compressibilityDefault = 50;
 *  console display
 ***************************************/
 #define DISPLAY(...)         fprintf(stderr, __VA_ARGS__)
-#define DISPLAYLEVEL(l, ...) if (g_displayLevel>=l) { DISPLAY(__VA_ARGS__); }
+#define DISPLAYLEVEL(l, ...) do { if (g_displayLevel>=(l)) DISPLAY(__VA_ARGS__); } while(0)
 static U32 g_displayLevel = 2;   /* 0 : no display;   1: errors;   2 : + result + interaction + warnings;   3 : + progression;   4 : + information */
 
-#define DISPLAYUPDATE(l, ...) if (g_displayLevel>=l) { \
+#define DISPLAYUPDATE(l, ...) do { if (g_displayLevel>=(l)) \
             if ((clock() - g_time > refreshRate) || (g_displayLevel>=4)) \
             { g_time = clock(); DISPLAY(__VA_ARGS__); \
-            if (g_displayLevel>=4) fflush(stdout); } }
+            if (g_displayLevel>=4) fflush(stdout); } } while(0)
 static const clock_t refreshRate = CLOCKS_PER_SEC * 15 / 100;
 static clock_t g_time = 0;
 
@@ -100,15 +100,15 @@ static clock_t g_time = 0;
 #ifndef DEBUG
 #  define DEBUG 0
 #endif
-#define DEBUGOUTPUT(...) if (DEBUG) DISPLAY(__VA_ARGS__);
+#define DEBUGOUTPUT(...) do { if (DEBUG) DISPLAY(__VA_ARGS__); } while(0)
 #define EXM_THROW(error, ...)                                             \
-{                                                                         \
+do {                                                                      \
     DEBUGOUTPUT("Error defined at %s, line %i : \n", __FILE__, __LINE__); \
-    DISPLAYLEVEL(1, "Error %i : ", error);                                \
+    DISPLAYLEVEL(1, "Error %i : ", (error));                              \
     DISPLAYLEVEL(1, __VA_ARGS__);                                         \
     DISPLAYLEVEL(1, "\n");                                                \
     exit(error);                                                          \
-}
+} while(0)
 
 
 /* *************************************
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -74,7 +74,7 @@
 *  Macros
 ***************************************/
 #define DISPLAY(...)           fprintf(stderr, __VA_ARGS__)
-#define DISPLAYLEVEL(l, ...)   if (displayLevel>=l) { DISPLAY(__VA_ARGS__); }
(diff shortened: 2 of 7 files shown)


ehem:
Seems the debug build for Windows uses the equivalent of -Werror (rather odd, I'd expect such a thing in a release build, not a debug build).  It is complaining about a conditional expression being constant, which is indeed true (do { ... } while(0)); GCC and clang don't warn about this situation for good reason!  This is a build script bug, not a bug in these commits.

If you want to debate the issue, notice this change caused the missing semicolons to be flagged.  This is definitely an improvement!

I must advise examining the situation with the Travis CI Apple build.  The log is indicating the job is being terminated due to too much output, as such it may not be safe to rely on the results of that test.  Looks like the tests were designed for interactive operation, providing output in the middle then backspacing over to write updates.  This results in more output, which is great if a person is watching, but generates lots of output and would explain the log from automated testing.

Cyan4973:
There are several interesting topics in this PR.

1) The first one, "report position where unrecognized data occurs", is a good feature request.
I believe it would deserve to be separated so that we can study it in detail.
In the proposed implementation, I see 2 issues that could be improved, or explained : 

    + `const long position = ftell(finput);` -> how does this construction work with large files (> 4 GB). I'm mostly worried of 32-bits systems, and of Windows / Unix portability (seems one consider `long` to be 32 bits, while the other make it a `64-bits`).

    + tests : could we add at least one test which ensure the feature works as expected ? Best if it can fail when the feature is not implemented.

2) Regarding OS-X test console limit :
that's a good catch. There is indeed an issue to track.
I suspect the last test output too much data, even though it uses only ~400 lines.
That's because most messages are "status" messages, which override previous one without going to a newline. Nevertheless, for Travis, that's still data, even if not displayed.
Should be fixed.

3) using `do { ... } while(0)` construction : 
while I'm not opposed by default, this construction does indeed expose to the compiler warning issue you found.
My position in similar situations has been to go to `{ ... }` construction.
It still means some macros should be changed to ensure brackets,
typically from `if (condition) {...}` to `{ if (condition) {... } }`
the only exception being single-invocation macros, such as `#define DISPLAY(...) fprintf(stderr, __VA_ARGS__)` .

ehem:
1. That was a quick sample implementation, there are far better strategies, but that was handy for a quick implementation.  The real solution is to keep track as bytes are decompressed, that though requires rather more extensive changes (at which point speed may become an issue).
Personally I tend more towards `position = lseek(fileno(finput), 0, SEEK_CUR);`.  Looks like there is a `const off_t position = ftello(finput);` option too.

2.  Different word choice, but the exact same thing I was trying to suggest (they were designed to limit _lines_ of output, but Travis tracks _bytes_ of output).

3. Your alternative doesn't work, look at the example I suggested above.
```
if(<some_condition>) DISPLAYUPDATE(2, "Some condition");
else some_function();
```
If you merely use braces without `do/while(0)` this turns into:
```
if(<some_condition>) { if (g_displayLevel>=(2)) DISPLAY("Some condition"); };
else some_function();
```
At which point the extra semicolon causes an error.  As such the `do/while(0)` is _required_ otherwise you _must_ have all developers _know_ DISPLAYUPDATE() is a macro and this macro does **not** behave like a function, despite looking like one.  `do/while(0)` can also be used in other situations:
```
do {
    if(<error_condition>) break;
    a_function_call();
    if(<function_call_failed>) break;
    return <a_successful_result>;
} while(0);
<cleanup_code>
```
(some coding styles ban goto, at which point this construct gets heavily (ab)used)  The most sane thing to do is figure out how to tell that Windows compiler _not_ to issue that warning or not turn that warning into an error.  I'm unaware of other constructs to achieve this result, unless you want to use inline functions.

Cyan4973:
1. We'll go for the full solution of tracking the nb of bytes. This is what `zstd` does, and it proved reasonably simple to implement. In contrast, `lseek()`, `ftello()` and equivalent quickly become painful from a portability perspective.

3. This is a minor detail, so I'm not oppose to it. Just keep in mind that `DISPLAYUPDATE()` is not part of any public interface, it's merely a local internal macro, limited scope and limited usages. None of them uses `if ()` construction btw, hence `while (0)` fixes a potential issue which does not exist yet.
I would have used a `static` function if it was possible, but `__VA_ARGS__` makes it undesirable.
Removing the Visual warning can be done through [#pragma](https://github.com/lz4/lz4/blob/dev/programs/lz4io.c#L38).

Cyan4973:
I initially misunderstood your request.

You want to know where the decompression ended within the compressed stream when it failed.
This is a bit more complex, because we have multiple ways of consuming the stream, so we would need to instrument them all with additional status information to keep in sync. The complexity involved looks too large for the feature.

The proposed method, to use `ftell()`, is a lot simpler, and can be indeed be easily integrated. That's what d18084c65561cddec48f74c3a91363c9f8e6f709 does.

Note however that `ftell()` will fail if input is `stdin`,
and even on files, it's more than likely to fail on files >= 2 GB.

Still, I figure this is still a useful feature, even if limited to files < 2 GB.
This is probably a large majority of use cases.

For all other use cases, the program will simply continue to behave as before, that is displaying "Stream followed by undecodable data" without reporting the position where it failed in the compressed stream.

PR #314, opened 2017-02-07
Add reporting of position where LZ4 stream ends; Checking of macros

ehem:
This is pretty much the same as #313, except for the master branch.  Cleanup macros and add reporting of where the valid LZ4 stream ends, if it is followed by unknown data.

Exactly the same issues apply.  Most likely the Windows-based debug compile will fail due to using the equivalent of -Werror, which on MS likes to flag "do { ... } while(0)".  Neither GCC nor clang flag this for good reason.  Simplest solution may be to tell that compiler not to flag use of a constant condition, this is definitely a build bug, not a bug in these patches.

### DIFF
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -83,13 +83,13 @@ static U32 g_compressibilityDefault = 50;
 *  console display
 ***************************************/
 #define DISPLAY(...)         fprintf(stderr, __VA_ARGS__)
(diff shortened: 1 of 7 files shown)


terrelln:
The master branch is only for releases. All development happens in `dev` and `dev` gets merged into `master` each release.

Cyan4973:
Correct, we only take PR into `dev`.
They are later merged into `master` when deemed stable enough, typically before a release

Issue #308, opened 2017-01-24
Test failure in 1.7.5


lfam:
Greetings!

I'm working on distro packaging for GNU Guix [0]. I'm trying to update our lz4 package to 1.7.5, but the test suite is failing like this:

```
make -C tests test
make[1]: Entering directory '/tmp/guix-build-lz4-1.7.5.drv-0/lz4-1.7.5/tests'
make -C ../programs clean lz4 CFLAGS="-O3  -g -Wall -Wextra -Wundef -Wcast-qual -Wcast-align -Wshadow -Wswitch-enum -Wdeclaration-after-statement -Wstrict-prototypes -Wpointer-arith -Wstrict-aliasing=1 "
make[2]: Entering directory '/tmp/guix-build-lz4-1.7.5.drv-0/lz4-1.7.5/programs'
Cleaning completed
gcc -O3  -g -Wall -Wextra -Wundef -Wcast-qual -Wcast-align -Wshadow -Wswitch-enum -Wdeclaration-after-statement -Wstrict-prototypes -Wpointer-arith -Wstrict-aliasing=1  -I../lib -DXXH_NAMESPACE=LZ4_  -c -o ../lib/lz4.o ../lib/lz4.c
gcc -O3  -g -Wall -Wextra -Wundef -Wcast-qual -Wcast-align -Wshadow -Wswitch-enum -Wdeclaration-after-statement -Wstrict-prototypes -Wpointer-arith -Wstrict-aliasing=1  -I../lib -DXXH_NAMESPACE=LZ4_  -c -o ../lib/lz4hc.o ../lib/lz4hc.c
gcc -O3  -g -Wall -Wextra -Wundef -Wcast-qual -Wcast-align -Wshadow -Wswitch-enum -Wdeclaration-after-statement -Wstrict-prototypes -Wpointer-arith -Wstrict-aliasing=1  -I../lib -DXXH_NAMESPACE=LZ4_  -c -o ../lib/xxhash.o ../lib/xxhash.c
gcc -O3  -g -Wall -Wextra -Wundef -Wcast-qual -Wcast-align -Wshadow -Wswitch-enum -Wdeclaration-after-statement -Wstrict-prototypes -Wpointer-arith -Wstrict-aliasing=1  -I../lib -DXXH_NAMESPACE=LZ4_  -c -o ../lib/lz4frame.o ../lib/lz4frame.c
[...]

PR #317, opened 2017-02-14
Don't use "foo && false || true"

esiegerman:
This is mentioned as an aside in @mwgamera's comment to #308, but isn't directly related to that bug.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -236,17 +236,17 @@ test-lz4-basic: lz4 datagen unlz4 lz4cat
 	./datagen -g256MB | $(LZ4) -vqB4D | $(LZ4) -t
 	@echo "hello world" > tmp
 	$(LZ4) --rm -f tmp
-	ls -ls tmp         && false || true   # must fail (--rm)
-	ls -ls tmp.lz4
-	$(PRGDIR)/lz4cat tmp.lz4              # must display hello world
-	ls -ls tmp.lz4
+	test ! -f tmp                      # must fail (--rm)
+	test   -f tmp.lz4
+	$(PRGDIR)/lz4cat tmp.lz4           # must display hello world
+	test   -f tmp.lz4
 	$(PRGDIR)/unlz4 --rm tmp.lz4
-	ls -ls tmp
-	ls -ls tmp.lz4     && false || true   # must fail (--rm)
-	ls -ls tmp.lz4.lz4 && false || true   # must fail (unlz4)
-	$(PRGDIR)/lz4cat tmp                  # pass-through mode
-	ls -ls tmp
-	ls -ls tmp.lz4     && false || true   # must fail (lz4cat)
+	test   -f tmp
+	test ! -f tmp.lz4                  # must fail (--rm)
+	test ! -f tmp.lz4.lz4              # must fail (unlz4)
+	$(PRGDIR)/lz4cat tmp               # pass-through mode
+	test   -f tmp
+	test ! -f tmp.lz4                  # must fail (lz4cat)
 	$(LZ4) tmp                         # creates tmp.lz4
 	$(PRGDIR)/lz4cat < tmp.lz4 > tmp3  # checks lz4cat works with stdin (#285)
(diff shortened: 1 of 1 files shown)


tests/Makefile
```diff
@@ -236,17 +236,17 @@ test-lz4-basic: lz4 datagen unlz4 lz4cat
 	./datagen -g256MB | $(LZ4) -vqB4D | $(LZ4) -t
 	@echo "hello world" > tmp
 	$(LZ4) --rm -f tmp
-	ls -ls tmp         && false || true   # must fail (--rm)
-	ls -ls tmp.lz4
-	$(PRGDIR)/lz4cat tmp.lz4              # must display hello world
-	ls -ls tmp.lz4
+	test ! -f tmp                      # must fail (--rm)
```
Cyan4973: cc @inikep 
what is the portability of `test` command ?

tests/Makefile
```diff
@@ -262,22 +262,22 @@ test-lz4-hugefile: lz4 datagen
 
 test-lz4-testmode: lz4 datagen
 	@echo "\n ---- bench mode ----"
-	$(LZ4) -bi1
+	  $(LZ4) -bi1
 	@echo "\n ---- test mode ----"
-	./datagen | $(LZ4) -t             && false || true
-	./datagen | $(LZ4) -tf            && false || true
+	! ./datagen | $(LZ4) -t
```
Cyan4973: To which result does `!` refer to ?
The intention here is to get result from `$(LZ4)`

Cyan4973:
This is overall a welcomed change, even though there are a few questions left.
Portability constraints can be pretty hard at times ....

tests/Makefile
inikep: It seems to be fine for Linux/Unix/*BSD/MacOS/MinGW:
https://en.wikipedia.org/wiki/Test_(Unix)

esiegerman: `test` is POSIX -- http://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html -- so approximately as portable as other UNIXisms the Makefile depends on.

I don't know a lot about building UNIXy stuff under Windows, but lz4's Makefiles assume the presence of enough UNIX stuff that I'm pretty sure it's safe to assume you have `test` as well.

tests/Makefile
esiegerman: "|" binds more tightly than "!", so this does the right thing here.

According to [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_09_02):

> If the pipeline [i.e. series of commands separated by `|`] does not begin with the ! reserved word, the exit status shall be the exit status of the last command specified in the pipeline. Otherwise, the exit status shall be the logical NOT of the exit status of the last command.

&lt;aside&gt;

To get the behaviour you _don't_ want here, but which is sometimes useful[1], there's no standard way that I know of, but `bash` has a `pipefail` extension; if that option is set, the exit status of the pipeline as a whole is the last _nonzero_ status of any of the simple commands (or 0 if they all succeed).  But even there, "!" inverts the exit status of the pipeline as a whole.  (And you're not setting `pipefail` in any case...)

[1] Here's a case where it's the first command whose status one wants:

     make test | tee logfile || echo "test failed"

(which (minus the "|| echo" part) is how I got here in the first place -- trying to capture the output of "make test" triggered #308 :-/ ).

&lt;/aside&gt;

esiegerman:
> Portability constraints can be pretty hard at times ....

Agreed!  I tested these changes of course, but even so, your question about "! cmd1 | cmd2" led me to seriously question whether I'd screwed up; I had to go off and read specs and man pages to convince myself (a) that I'd got it right and (b) that the behaviour in question is as intrinsic to Bourne-shell-ness as I'd assumed :-)

Several decades of shell-scripting experience makes it easy to do cleanups like this pretty much on autopilot -- but alas, its all too easy to forget which bits are portable and which bits aren't; so thanks for forcing me to question my assumptions.

PR #318, opened 2017-02-15
Explicitly create $(DESTDIR)$(LIBDIR)/ at install time

esiegerman:
This is needed on systems where it isn't the parent of
$(PKGCONFIGDIR), and so doesn't get created implicitly.


Cyan4973:
Good point !

PR #319, opened 2017-02-16
Circle CI tests

inikep:
fixed fseek with original MinGW (http://www.mingw.org/)

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -78,7 +78,7 @@ lz4c:
 	$(MAKE) -C $(PRGDIR) clean $@ CFLAGS="$(CFLAGS)"
 
 lz4c32:   # create a 32-bits version for 32/64 interop tests
-	$(MAKE) -C $(PRGDIR) clean $@ CFLAGS="-m32 $(CFLAGS)"
+	$(MAKE) -C $(PRGDIR) clean $@ CFLAGS="-m32 $(CFLAGS) -I/usr/include/x86_64-linux-gnu"
 	cp $(LZ4) $(LZ4)c32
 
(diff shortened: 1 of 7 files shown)


tests/Makefile
```diff
@@ -301,6 +301,7 @@ test-lz4c: lz4c datagen
 	./datagen -g256MB | $(LZ4)c -l -v    | $(LZ4)c   -t
 
 test-lz4c32: CFLAGS+=-m32
+test-lz4c32: MOREFLAGS+="-I/usr/include/x86_64-linux-gnu"
```
Cyan4973: Why is this directory required ?
Is it stable ?

inikep: The issue is a conflict of gcc-multilib and gcc-arm described here:
https://bugs.launchpad.net/ubuntu/+source/gcc-defaults/+bug/1300211

With Travis we have different containers but with Circle CI there is a single one.
The best solution I found 
https://bugs.launchpad.net/ubuntu/+source/gcc-defaults/+bug/1300211/comments/19
was to add `-I`.
Without it there is an error for 32-bit targets:
`/usr/include/linux/errno.h:1:23: fatal error: asm/errno.h: No such file or directory`

inikep: I'm not sure if it's needed here anymore because I've started using in Circle.yml:
`make -C tests test32 MOREFLAGS="-I/usr/include/x86_64-linux-gnu" && make clean`

Cyan4973: Good point, it's likely better to keep it within `circle.yml` if it's specific to this environment

inikep: I will check it

PR #328, opened 2017-03-05
Fix test-lz4-basic

ldv-alt:
When no output filename is specified and stdout is not a terminal,
lz4 doesn't attempt to guess an output filename and uses stdout for
output.

This change fixes test-lz4-basic when run without a terminal
by specifying output filenames.


Cyan4973:
👍 
should fix #308

ldv-alt:
yes, it fixes #308

PR #329, opened 2017-03-07
lz4hc: Add LZ4_compressHC_destSize

remittor:


Cyan4973:
Thanks for the patch @remittor .
It's a very important patch, with important impact, so it needs special care.

2 things I would like to see to validate this new feature : 
- tests, especially fuzzer tests.
  There is already a framework for `_destSize()` variant : 
  https://github.com/lz4/lz4/blob/dev/tests/fuzzer.c#L319
  I would like this framework to be extended to also include the new `LZ4_compressHC_destSize()` (or its equivalent).
- Benchmark. 
  The patch modify an inner function, already in use.
  Therefore, what happens to the "regular" HC version ?
  Benchmark numbers shall prove that speed and compression ratio are not impacted negatively by the modification.

remittor:
Fuzzer result: `LZ4_decompress_safe() failed on data compressed by LZ4_compressHC_destSize`
I need help.

remittor:
Problem condition: https://github.com/lz4/lz4/blob/eb51b2b8d9254add9d72eb26064518d5bcf1e9ee/lib/lz4.c#L1225
`if (unlikely(cpy>oend-12)) {`
What is this magic number?

remittor:
I fix this error:
`if (cpy > oend-LASTLITERALS+1) goto _output_error;` /* add + 1 */
All fuzzer test completed OK! (without LZ4_compressHC_destSize)

But now i have truble here: https://github.com/lz4/lz4/blob/eb51b2b8d9254add9d72eb26064518d5bcf1e9ee/lib/lz4.c#L1155

Cyan4973:
Pay attention to [LZ4 parsing restrictions](https://github.com/lz4/lz4/blob/dev/doc/lz4_Block_format.md#parsing-restrictions), which makes ending a block a tricky operation.

remittor:
Test result after fix:
- Fuzzer: OK (65536 iterations)
- Speed before adding LZ4_compressHC_destSize:
> $ ./fullbench -c10 silesia.tar
> *** LZ4 speed analyzer v1.7.5 64-bits, by Yann Collet ***
> silesia.tar :
> 10-LZ4_compress_HC              :211947520 -> 78009943 (36.81%),   22.3 MB/s
- Speed after adding LZ4_compressHC_destSize:
> $ ./fullbench -c10 silesia.tar
> *** LZ4 speed analyzer v1.7.5 64-bits, by Yann Collet ***
> silesia.tar :
> 10-LZ4_compress_HC              :211947520 -> 78009943 (36.81%),   22.5 MB/s

Cyan4973:
> frametest: FAIL

Since frametests were not changed, and do not target the new variant `LZ4_compressHC_destSize()`, this suggests that introduced modifications affect the behavior of the "regular" function.

Though how exactly, it is difficult to tell.

remittor:
Latest patch fix frametest. It's amazing!

Cyan4973:
Congratulations ! Seems to pass all tests !
Let me have a closer look at the code now

Cyan4973:
So far so good,
your patch sustain advanced resiliency tests,
and benchmark differences are within noise level.
I'll merge it, and add your contribution in the NEWS section

Cyan4973:
Some small but notable changes : 
- Created a new `LZ4_HC_STATIC_LINKING_ONLY` section, where new prototypes must stage some time before reaching stable status.
- Changed `LZ4_compressHC_destSize()` into `LZ4_compress_HC_destSize()`, for better conformance to API coding style
- Moved `LZ4_compress_HC_destSize()` from `deprecated` category into `static linking` section : it's a useful one-pass prototype, with good use cases on its own.

Cyan4973:
I just noticed that `LZ4_compress_HC_continue_destSize()` always auto-init,
which means it does not really `continue` with previous blocks as history.

It's slightly misleading, and of course, it's less efficient, especially on small blocks.
Is there a reason for that ? 
Would `LZ4_compress_HC_continue_destSize()` fail if it used previous blocks as history ?

For independent blocks compression, there is already `LZ4_compress_HC_destSize()`, which looks like the good choice for this usage.

remittor:
> It's slightly misleading, and of course, it's less efficient, especially on small blocks.
> Is there a reason for that ?
> Would `LZ4_compress_HC_continue_destSize()` fail if it used previous blocks as history ?

It's my fail. Function `LZ4_compress_HC_continue_destSize()` must be deleted!
I create a special C++ class for sending files via HTTP , which provided following methods :
```c
  int compress_frame(const void * src, int & srcSize, void * dest, int maxDestSize);
  INT64 set_file(HANDLE file, INT64 start_pos = 0, INT64 read_size = 0);
  int compress_file(void * dest, int maxDestSize);
  int decompress_stream(const void * src, int srcSize, void * dest, int destSize);
```
My frames compatible with original LZ4 frames (frame header = 4 bytes).

Cyan4973:
I've modified a bit your code and a few intermediate functions,
and now `LZ4_compress_HC_continue_destSize()` works as intended,
that is it can `continue` a compression stream, using previous blocks as history.
This results in much better compression ratios.

There is a limitation though : it only works up to level 10,
since your implementation is based on Hash Chain.

For `LZ4_compress_HC_destSize()` it's easier : compression level is transparently limited to level 10.

PR #330, opened 2017-03-10
fix bench

inikep:


Cyan4973:
Good point @inikep !

inikep:
This was submitted as a bug for Lizard so I also prepared fix for upstream.

PR #331, opened 2017-03-11
Export only those symbols that are part of public API

ldv-alt:
Windows builds already limit exporting of functions to those marked with LZ4LIB_API tag.
This short series implements the same behavior on other platforms assuming that gcc >= 4 (released ~12 years ago) is used.


Cyan4973:
Good patch,
providing same strategy as `zstd` for library visibility

PR #334, opened 2017-03-17
lz4hc: Fix static link LZ4_compress_HC_continue_destSize for MSVC

remittor:


Cyan4973:
Good point @remittor .
Apparently, `g++` doesn't like this patch.
Let's investigate why ...

Cyan4973:
I've updated the `dev` branch with a fix similar to yours,
but which seems to pass `g++` tests, at least on my dev station.
Let's wait for CI tests results

PR #337, opened 2017-03-23
cli: add GNU separator -- specifying that all following arguments are files

ldv-alt:
This option is supported by other compressors with compatible cli,
so add it to lz4 as well for better compatibility.


Cyan4973:
You are totally right !
And your patch is very clean ! Thanks !

PR #341, opened 2017-03-24
Ignore extensions in exe name matching

### LINKED ISSUE #335: LZ4 detection of lz4cat and unlz4 invocation doesn't work on MS-Windows
LZ4 detects that it has been invoked as 'lz4cat' or 'unlz4' by comparing its argv[0] to fixed strings.  However, on MS-Windows argv[0] can be something like "d:\foo\bar\lz4cat.exe", and the code is not prepared to deal with the .exe suffix.  This caused some tests to fail for me.

Here's a proposed patch, which also fixes a problem whereby fixed-style d:\foo/bar\baz file names would not be supported correctly:

--- programs/lz4cli.c~0	2017-01-03 07:09:52.000000000 +0200
+++ programs/lz4cli.c	2017-03-21 10:16:21.459614400 +0200
@@ -244,6 +244,7 @@ static const char* lastNameFromPath(cons
 {
     const char* name = strrchr(path, '/');
     if (name==NULL) name = strrchr(path, '\\');   /* windows */
+    else if (strchr(name, '\\')) name = strchr(name, '\\'); /* mixed */
     if (name==NULL) return path;
     return name+1;
 }
@@ -297,7 +298,7 @@ int main(int argc, const char** argv)
     LZ4IO_setOverwrite(0);
 
     /* lz4cat predefined behavior */
-    if (!strcmp(exeName, LZ4CAT)) {
+    if (!strncmp(exeName, LZ4CAT, strlen(LZ4CAT))) {
         mode = om_decompress;
         LZ4IO_setOverwrite(1);
         LZ4IO_setRemoveSrcFile(0);
@@ -306,7 +307,7 @@ int main(int argc, const char** argv)
         displayLevel=1;
         multiple_inputs=1;
     }
-    if (!strcmp(exeName, UNLZ4)) { mode = om_decompress; }
+    if (!strncmp(exeName, UNLZ4, strlen(UNLZ4))) { mode = om_decompress; }
 
     /* command switches */
     for(i=1; i<argc; i++) {

iburinoc:
Add some extra logic so that `./lzcat` matches, `lzcat.exe` matches, but `lzcatsss` doesn't.

Fixes #335

### DIFF
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -242,10 +242,19 @@ static void waitEnter(void)
 
 static const char* lastNameFromPath(const char* path)
 {
-    const char* name = strrchr(path, '/');
-    if (name==NULL) name = strrchr(path, '\\');   /* windows */
-    if (name==NULL) return path;
-    return name+1;
+    const char* name = path;
+    if (strrchr(name, '/')) name = strrchr(name, '/') + 1;
+    if (strrchr(name, '\\')) name = strrchr(name, '\\') + 1; /* windows */
+    return name;
+}
+
+/*! exeNameMatch() :
+    @return : a non-zero value if exeName matches test, excluding the extension
+   */
+static int exeNameMatch(const char* exeName, const char* test)
+{
+    return !strncmp(exeName, test, strlen(test)) &&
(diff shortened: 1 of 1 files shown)


Cyan4973:
Looks good !

Issue #336, opened 2017-03-22
LZ4 on MS-Windows thinks that null device is a console


Eli-Zaretskii:
LZ4 uses `_isatty` to detect whether its standard input/output is a console device.  But `_isatty` on MS-Windows actually detects character devices, and "NUL", the null device, is a character device.  So LZ4 behaves as if the null device is a console, which fails several features, and caused failures in the test suite.

I attach a patch to fix that.  It uses a GCC extension, so if that is inappropriate for MS Visual Studio, the CPP condition should be changed to use `__MINGW32__` instead of `_WIN32`.  Alternatively, the macro `IS_CONSOLE` should be converted to an inline function, in which case the GCC extension is not needed.
[diff2.txt](https://github.com/lz4/lz4/files/860507/diff2.txt)

iburinoc:
Hi @Eli-Zaretskii, thanks for contributing a patch to fix the issue!  Quick question, is there a reason your patch excludes `OS2`, `MSDOS`, and `CYGWIN` from the updated code?  They would probably suffer from the same issue, right?

Eli-Zaretskii:
No, the problem is specific to MS-Windows.  The MSDOS build uses the DJGPP library, where this issue doesn't exist, and Cygwin is a Posix-compliant environment, so its `isatty` definitely should work correctly.  OS2 I know nothing about, so you need to ask an OS2 expert about that.  The solution uses Windows-specific APIs, so it will most probably not work on OS2.

PR #342, opened 2017-03-27
Fix IS_CONSOLE returning 1 for NUL on windows

iburinoc:
Fixes https://github.com/lz4/lz4/issues/336 with slight modification to the patch, as it didn't build on MSVC.

Tested that CYGWIN doesn't need the patch (it doesn't seem to have `NUL` anyway)


Cyan4973:
Great !

PR #359, opened 2017-05-23
Add DLL files to the INSTALL target

PierreNav:
Hello guys,
For my project I need to package dll and lib files from the cmake install target.
Hence I added this line in the CMakeLists.

Thanks,
I will be glad to discuss in more details if needed.


Cyan4973:
Thanks @PierreNav , looks good to me

PR #367, opened 2017-06-19
Fix gcc7 Wimplicit-fallthrough warnings

Chocobo1:
For the default Wimplicit-fallthrough=3 level,
the comment should start with "fall*"

~~About the fix in lz4cli.c, `gcc 7.1.1` will only suppress the warning when done like this:~~
```c++
#ifdef UTIL_HAS_CREATEFILELIST
                    /* recursive */
                case 'r': recursive=1;  /* without break */
#endif
                    /* fall-through */
```
~~so I opt to use `__attribute__ ((fallthrough));` which can be placed in the `#ifdef` block~~


Cyan4973:
Thanks !

PR #371, opened 2017-06-26
tests/Makefile: don't use LIBDIR as variable

jku:
LIBDIR may be overriden with a environment variable: In this case make
clean breaks in tests/. Use another variable name.


jku:
In case the context isn't obvious: the root Makefile uses LIBDIR to specify actual libdir, and this can be overriden with a environment variable. This works totally fine, except in tests/ where the same variable name is used for a completely different thing.

Cyan4973:
Good point @jku !
Thanks for the patch !

PR #372, opened 2017-07-08
Fix formatting of concatenation example in docs

ferdnyc:
The formatting in the `lz4.1` man page section "Concatenation of .lz4 files" is somewhat corrupted, with the two example commands run together on one line. Obviously, `lz4 file1 > foo.lz4 lz4 file2 >> foo.lz4` is not a valid shell command. So, at the cost of a little vertical space, I dropped the (two-line) "script" down onto a separate indented paragraph. To recover that lost space, I joined the sentence explaining the equivalency into a single line, with the commands bolded according to the surrounding style.

So this:
<pre>
   <b>Concatenation of .lz4 files</b>
       It  is  possible  to  concatenate <b>.lz4</b> files as is. <b>lz4</b> will decompress
       such files as if they were a single <b>.lz4</b> file. For example: lz4 file1 >
       foo.lz4 lz4 file2 >> foo.lz4

       then lz4cat foo.lz4

       is equivalent to : cat file1 file2
</pre>
becomes this:
<pre>
   <b>Concatenation of .lz4 files</b>
       It  is  possible  to  concatenate <b>.lz4</b> files as is. <b>lz4</b> will decompress
       such files as if they were a single <b>.lz4</b> file. For example:

           lz4 file1 > foo.lz4
           lz4 file2 >> foo.lz4

       Then <b>lz4cat foo.lz4</b> is equivalent to <b>cat file1 file2</b>.
</pre>

I chose not to bold the commands in the intended "script" paragraph as they weren't inline with text, though that kind of feels like dealer's choice re: house style.

...I also updated the Markdown version of the man page to match, as the example's formatting was similarly corrupted there.

### DIFF
--- a/programs/lz4.1
+++ b/programs/lz4.1
@@ -60,13 +60,15 @@ Consequently, \fBlz4 \-m \-\-rm\fR behaves the same as \fBgzip\fR\.
 .IP "" 0
 .
 .SS "Concatenation of \.lz4 files"
-It is possible to concatenate \fB\.lz4\fR files as is\. \fBlz4\fR will decompress such files as if they were a single \fB\.lz4\fR file\. For example: lz4 file1 > foo\.lz4 lz4 file2 >> foo\.lz4
+It is possible to concatenate \fB\.lz4\fR files as is\. \fBlz4\fR will decompress such files as if they were a single \fB\.lz4\fR file\. For example:
 .
-.P
-then lz4cat foo\.lz4
+.IP "" 4
+lz4 file1 > foo\.lz4
+.br
+lz4 file2 >> foo\.lz4
 .
-.P
-is equivalent to : cat file1 file2
+.IP "" 0
+Then \fBlz4cat foo\.lz4\fR is equivalent to \fBcat file1 file2\fR\.
 .
 .SH "OPTIONS"
 .
--- a/programs/lz4.1
+++ b/programs/lz4.1
@@ -67,7 +67,7 @@ lz4 file1 > foo\.lz4
 .br
 lz4 file2 >> foo\.lz4
 .
-.IP "" 0
+.P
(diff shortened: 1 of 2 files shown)


Cyan4973:
Thanks @ferdnyc 

These are very good points.
Note that `lz4.1` is created _from_ `lz4.1.md` using `make man`.
And I suspect the display issue comes from this transformation.

Unfortunately, any modification done directly into `lz4.1` will probably be lost on next `lz4.1.md` update, since the "source of truth" is `lz4.1.md`.

ferdnyc:
You know, I hadn't caught that `lz4.1` was a generated file. That's what I get for using Github to make changes, instead of checking out and actually building the project.

Fortunately, it looks like, with my changes to `lz4.1.md`, `lz4.1` is generated identical (in terms of `man -l ./lz4.1` output) to the changes I'd made by hand! So you can just check in the one commit for the Markdown changes, and the other two are redundant. Sorry about that!

ferdnyc:
> And I suspect the display issue comes from this transformation.

Oh, on this point, though: It doesn't seem that way, the formatting of that section was equally corrupted in the Markdown file. In fact, `lz4.1` was (and is, when regenerated after applying my  7e8ee7f change) a surprisingly accurate representation of the Markdown it's generated from, when rendered with commandline `man`. (Gnome Yelp, OTOH, makes a much bigger mess of formatting it.)

Perhaps the misformatting was introduced when `lz4.1.md` was originally generated? (Assuming it didn't start life as a Markdown file.)

Cyan4973:
I cannot "selectively" apply your PR,
but can certainly port the relevant part manually.

ferdnyc:
Yeah, sorry, I'd realized. Already preparing a new branch with just the one change. I'll send a PR in a few minutes.

PR #373, opened 2017-07-08
Fix formatting of concatenation example

ferdnyc:
The "Concatenation of .lz4 files" section contains example commands
that are run together on one line, making them invalid. Wrap them
in a code block and clean up surrounding formatting.


Cyan4973:
thanks !

PR #378, opened 2017-08-07
Allow to predefine FORCE_INLINE macro.

deymo:
FORCE_INLINE macro is defined based on the compiler used. When using
gcc, it will include "__attribute__((always_inline))" forcing gcc to
always inline all the functions marked as FORCE_INLINE. However, this
can cause a performance degradation of about 15%.

This patch allows to set the FORCE_INLINE macro from the compiler
command line to either "static" or "static inline" giving allowing it
to inline functions as needed when performing optimizations.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -96,19 +96,26 @@
 *  Compiler Options
 **************************************/
 #ifdef _MSC_VER    /* Visual Studio */
-#  define FORCE_INLINE static __forceinline
 #  include <intrin.h>
 #  pragma warning(disable : 4127)        /* disable: C4127: conditional expression is constant */
 #  pragma warning(disable : 4293)        /* disable: C4293: too large shift (32-bits) */
-#else
-#  if defined(__GNUC__) || defined(__clang__)
-#    define FORCE_INLINE static inline __attribute__((always_inline))
-#  elif defined(__cplusplus) || (defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) /* C99 */)
-#    define FORCE_INLINE static inline
-#  else
-#    define FORCE_INLINE static
-#  endif
 #endif  /* _MSC_VER */
+
+#ifndef FORCE_INLINE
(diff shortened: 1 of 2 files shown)


Cyan4973:
Interesting !
Is the 15% performance degradation specific to a platform/target, or generic across `gcc` ?

deymo:
Sorry for the lack of context, I tried this in x86_64 (desktop) and the difference was significant so I went to benchmark it more accurately for all the commits between the last tag I was using (r131) and 1.7.5. The biggest regression was introduced by this commit 258a5e7fa44825488fe4683f3a04c36fc944a23f because it effectively caused FORCE_INLINE to include the always_inline for gcc. The other commit that introduced a small degradation (about 3%) was this a7672dc4f91e1be2046b50ac4b14d89f5fea45f3 , which is earlier and meant to improve performance by unrolling 8 bytes. For me that was a little bit worse but the difference was small and YMMV.
Since this force-inline vs let the compiler decide might depend on the compiler, I'd opted for allowing to select it from the command line (basically from CFLAGS) so users can pick their choice without changing your defaults.

Cyan4973:
Interesting.
If I do remember correctly, `attribute((always_inline))` was added to improve `clang` performance, which was severely behind.
It was also supposed to have no impact on `gcc`.

I should try some tests on my side too.
Maybe compiler version could be at stake too.

Cyan4973:
I added your patch because it's well constructed, and harmless by default,
only a user aware of this possibility can actually modify `FORCE_INLINE` definition at compilation time.

I also made some speed tests on 2 different machines, with both `gcc` and `clang`.
For information, I couldn't get any positive change by modifying `FORCE_INLINE` definition.
Best I could get was, on one system, with `-DFORCE_INLINE=static` and `gcc-7`, a 3% increase in decompression speed, but with a corresponding 2% decrease in compression speed. All other tests were at best equivalent, if not downright detrimental (especially `clang`, which would lose -15% performance).

PR #380, opened 2017-08-11
Frame Dictionary API

Cyan4973:
Look into lib/lz4frame_static.h

### DIFF
--- a/lib/lz4frame.c
+++ b/lib/lz4frame.c
@@ -158,7 +158,7 @@ static void LZ4F_writeLE64 (void* dst, U64 value64)
 #define LZ4F_BLOCKSIZEID_DEFAULT LZ4F_max64KB
 
 static const size_t minFHSize = 7;
-static const size_t maxFHSize = LZ4F_HEADER_SIZE_MAX;   /* 15 */
+static const size_t maxFHSize = LZ4F_HEADER_SIZE_MAX;   /* 19 */
 static const size_t BHSize = 4;
 
 
@@ -291,7 +291,7 @@ static size_t LZ4F_compressBound_internal(size_t srcSize,
 size_t LZ4F_compressFrameBound(size_t srcSize, const LZ4F_preferences_t* preferencesPtr)
 {
     LZ4F_preferences_t prefs;
-    size_t const headerSize = maxFHSize;      /* max header size, including magic number and frame content size */
+    size_t const headerSize = maxFHSize;      /* max header size, including optional fields */
 
     if (preferencesPtr!=NULL) prefs = *preferencesPtr;
     else memset(&prefs, 0, sizeof(prefs));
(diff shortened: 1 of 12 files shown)


lib/lz4frame_static.h
```diff
@@ -82,6 +82,60 @@ typedef enum { LZ4F_LIST_ERRORS(LZ4F_GENERATE_ENUM) } LZ4F_errorCodes;
...
+ *********************************/
+typedef struct LZ4F_CDict_s LZ4F_CDict;
+
+/*! LZ4_createCDict() :
+ *  When compressing multiple messages / blocks with the same dictionary, it's recommended to load it just once.
+ *  LZ4_createCDict() will create a digested dictionary, ready to start future compression operations without startup delay.
+ *  LZ4_CDict can be created once and shared by multiple threads concurrently, since its usage is read-only.
+ * `dictBuffer` can be released after LZ4_CDict creation, since its content is copied within CDict */
+LZ4F_CDict* LZ4F_createCDict(const void* dictBuffer, size_t dictSize);
+void        LZ4F_freeCDict(LZ4F_CDict* CDict);
```
terrelln: The other lz4f free functions return an `LZ4F_errorCode_t`, do we want to as well? I don't particularly care either way.

lib/lz4frame.c
```diff
@@ -360,32 +366,99 @@ size_t LZ4F_compressFrame(void* dstBuffer, size_t dstCapacity, const void* srcBu
...
+    cdict->fastCtx = LZ4_createStream();
+    cdict->HCCtx = LZ4_createStreamHC();
+    if (!cdict->dictContent || !cdict->fastCtx || !cdict->HCCtx) {
+        LZ4F_freeCDict(cdict);
+        return NULL;
+    }
+    memcpy(cdict->dictContent, dictStart, dictSize);
+    LZ4_resetStream(cdict->fastCtx);
+    LZ4_loadDict (cdict->fastCtx, (const char*)cdict->dictContent, (int)dictSize);
+    LZ4_resetStreamHC(cdict->HCCtx, LZ4HC_CLEVEL_DEFAULT);
```
terrelln: Does this work for the OPT LZ4HC, since it uses a different method for loading the dictionary content?

Cyan4973: Not directly. User requesting levels 11+ are silently downgraded to level 10, in order to remain compatible with the search structure (hash chain).

Long term plan is to get rid of the Binary Tree implementation in lz4opt, and replace it with the same hash chain as lz4hc, so that there is a continuation between both modes.

lib/lz4frame_static.h
Cyan4973: Good point.
The `LZ4F_errorCode_t` was there in case a `free` operation could fail.
But effectively, it never fails, so the return type is not very useful.
In fact, it creates additional charge into the user code, because now it adds to add some test and special exception code in case the function would fail.

As a consequence, nowadays, I'm more in the mood of having `free` function with a `void` type, to make it clear it never fails, and does not need extra-care.

That being said, it's true that this choice makes this prototype stand apart from other `free` functions defined in `lz4frame`. And it would be a good reason to use the same return type, even if the function always succeeds.

Which choice is better ? Well, I'm not sure. Feedback welcomed.

Cyan4973: note : I was considering to re-use existing lz4frame API design, hence have `LZ4F_freeCDict()` generate an `LZ4F_errorCode_t` return code even if it's not useful and invite users to do more work, just to keep it equivalent to existing `LZ4F_freeCompressionContext()` and `LZ4F_freeDecompressionContext()`. 

But then I realised that even `LZ4F_createCDict()` is different : 
its return value is a pointer to the target object `LZ4F_CDict*`, or `NULL` if there is an error : 
This is different from `LZ4F_createCompressionContext()`, which returns an `LZ4F_errorCode_t`, and requires its first argument to be a mutable pointer to target object `LZ4F_CDict**`. 
This design feels inferior, as it's more complex to read (and gets one step closer to [triple-start programmer status](http://wiki.c2.com/?ThreeStarProgrammer)), but also because it's not possible to `const` the resulting pointer, which I consider valuable these days.

I'm now wondering if it makes any sense to half-adapt `CDict` API only on the `free` side.

Alternatively, the other option is to refactor the entire `LZ4F_createCDict()/LZ4F_freeCDict()` API to look like `LZ4F_createCompressionContext()/LZ4F_freeCompressionContext()`. But that feels less and less beneficial...

To sum up : 
```
LZ4F_CDict* const cdict = LZ4F_createCDict(...);
if (cdict==NULL) {
```
vs
```
LZ4F_CDict* cdict; 
LZ4F_errorCode_t createError = LZ4_createCDict(&cdict, ...);
if (LZ4F_isError(createError)) {
```

PR #382, opened 2017-08-15
better respect GNU standard Makefile conventions

Cyan4973:
supports lowercase directory variables 
add an "Installation" section in README.md 
added an INSTALL file

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -60,7 +60,7 @@ ifeq ($(shell uname), Darwin)
 	SHARED_EXT = dylib
 	SHARED_EXT_MAJOR = $(LIBVER_MAJOR).$(SHARED_EXT)
 	SHARED_EXT_VER = $(LIBVER).$(SHARED_EXT)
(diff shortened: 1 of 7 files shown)


lib/Makefile
```diff
@@ -148,17 +154,17 @@ install: lib liblz4.pc
 	@$(INSTALL_DATA) liblz4.pc $(DESTDIR)$(PKGCONFIGDIR)/
 	@echo Installing libraries
 ifeq ($(BUILD_STATIC),yes)
-	@$(INSTALL_LIB) liblz4.a $(DESTDIR)$(LIBDIR)/liblz4.a
+	@$(INSTALL_DATA) liblz4.a $(DESTDIR)$(LIBDIR)/liblz4.a
```
terrelln: Why is this `INSTALL_DATA` and the shared library `INSTALL_PROGRAM`?

Cyan4973: My understanding is that a dynamic library is considered an "executable",
so it must use `INSTALL_PROGRAM` which kind of `chmod 755` the file,
while a static library is just an archive with `*.o` files into it,
so it's plain data, read at link stage.

PR #383, opened 2017-08-15
Block checksum

Cyan4973:
lz4frame now supports block checksum, as defined in the specification.
Restored block checksum command (`-BX`) on CLI


terrelln:
It looks good to me. There is a lot of noise because of the formatting/name changes, but I think I read over all the block checksum changes, and it looked right.

Issue #385, opened 2017-08-20
datagen: Unrecognized header


itoffshore:
In [Alpine Linux](http://alpinelinux.org/) now we run the included unit tests for all packages. The test log shows all tests passing successfully but I see the following errors (possibly related to Alpine using Musl C not Glibc):

```
Error 44 : Unrecognized header : file cannot be decoded 
! ./datagen | ../programs/lz4 -tf

Error 44 : Unrecognized header : file cannot be decoded 
\n ---- pass-through mode ----
! ./datagen | ../programs/lz4 -d  > /dev/null

Error 44 : Unrecognized header : file cannot be decoded 
./datagen | ../programs/lz4 -df > /dev/null
```

& also in the decompression function tests `tmp` is not found (I think this may be due to different options between `coreutils` & the builtin `busybox` tools):
```
Decompression functions : 
Read : 3068 MB   ==> 0.39%   
Compressed 3221225472 bytes into 12676935 bytes ==> 0.39%                      
Completed in 19.69 sec 
stdin                : decoded 3221225472 bytes                                
  899    1-LZ4_decompress_fast           :     18092 ->   109.9 MB/s
  963    3-LZ4_decompress_fast_usingDict :     18092 ->   119.5 MB/s
 6323    4-LZ4_decompress_safe           :     18092 ->   111.9 MB/s
 6621    6-LZ4_decompress_safe_usingDict :     18092 ->   118.4 MB/s
 6923    7-LZ4_decompress_safe_partial   :     18092 ->   112.8 MB/s
 1175    8-LZ4_decompress_safe_forceExtD :     18092 ->   119.2 MB/s
 1206   ls: tmp: No such file or directory
make[1]: *** [Makefile:185: test-lz4-contentSize] Error 1
```

This is the test log in full:

```
./fasttest
./frametest -T3mn
Starting lz4frame tester (64-bits, 1.8.0)
Seed = 5207
gcc -Os -fomit-frame-pointer -g -Wall -Wextra -Wundef -Wcast-qual -Wcast-align -Wshadow -Wswitch-enum -Wdeclaration-after-statement -Wstrict-prototypes -Wpointer-arith -Wstrict-aliasing=1  -I../lib -I../programs -DXXH_NAMESPACE=LZ4_ -I../lib -DXXH_NAMESPACE=LZ4_  -c -o bench.o bench.c
Basic tests completed 
[...]

PR #386, opened 2017-08-21
`make test` compatible with parallel execution (`-j#`)

Cyan4973:
each test section runs with its own set of files


Cyan4973:
fixes #385

PR #389, opened 2017-08-23
FIX: added prefix LZ4_ to  FORCE_INLINE to avoid redefinition warnings w…

tcpan:
…hen used with other libraries, e.g. murmur hash implementation in smhasher on github.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -101,21 +101,21 @@
 #  pragma warning(disable : 4293)        /* disable: C4293: too large shift (32-bits) */
 #endif  /* _MSC_VER */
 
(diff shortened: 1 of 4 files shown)


Cyan4973:
I'm fine with changing `FORCE_INLINE` into `LZ4_FORCE_inline`,
but not when it comes to `xxhash.c`.
`xxhash` being an independent module, it has its own "emulated namespace", so it's better to switch to `XXH_FORCE_INLINE` there.

tcpan:
Closed.  reforked and modified as suggested in order to avoid having a single commit with the correct changes.  A separate pull request will be created.

PR #391, opened 2017-08-24
FIX: added prefix to FORCE_INLINE to prevent redefinition error durin…

tcpan:
…g compilation when used with other libraries that define FORCE_INLINE,  e.g. murmur hash implementation in smhasher on github.


Cyan4973:
Looks good to me !

Issue #397, opened 2017-09-07
Streaming decompression issue w/emscripten


jscheid:
Hi,

I'm porting the [blockStreaming_ringBuffer.c](https://github.com/lz4/lz4/blob/dev/examples/blockStreaming_ringBuffer.c) example to emscripten and I've run into an issue. I've tried to get to the bottom of it, here is my analysis, which could be completely wrong...

Because the address space is not shared and there is no OS in memory, `decBuf` happens to be allocated at a low address (say 0x159c). At some point the dictionary grows to 9216 bytes. Now `lowLimit` gets to be negative [here](https://github.com/lz4/lz4/blob/1ad9a36cfb9fc669b0b18bfcd902d3e52a3d2cbb/lib/lz4.c#L1131). Because the pointer is unsigned, it becomes something like 0xfffff19c (emscripten is 32-bit). Now [this condition](https://github.com/lz4/lz4/blob/1ad9a36cfb9fc669b0b18bfcd902d3e52a3d2cbb/lib/lz4.c#L1187) will trigger always (practically speaking) because it is an unsigned comparison. 💥 

The following patch appears to fix it:

```diff
diff --git a/lib/lz4.c b/lib/lz4.c
index 96422f8..7f48e36 100644
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1128,7 +1128,6 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
     BYTE* const oend = op + outputSize;
     BYTE* cpy;
     BYTE* oexit = op + targetOutputSize;
-    const BYTE* const lowLimit = lowPrefix - dictSize;
 
     const BYTE* const dictEnd = (const BYTE*)dictStart + dictSize;
     const unsigned dec32table[] = {0, 1, 2, 1, 4, 4, 4, 4};
@@ -1184,7 +1183,7 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
         /* get offset */
         offset = LZ4_readLE16(ip); ip+=2;
         match = op - offset;
-        if ((checkOffset) && (unlikely(match < lowLimit))) goto _output_error;   /* Error : offset outside buffers */
+        if ((checkOffset) && (unlikely(match + dictSize < lowPrefix))) goto _output_error;   /* Error : offset outside buffers */
         LZ4_write32(op, (U32)offset);   /* costs ~1%; silence an msan warning when offset==0 */
 
         /* get matchlength */
```

[...]

PR #400, opened 2017-09-12
Add basic, failing test for emscripten

jscheid:
See lz4/lz4#397

@Cyan4973 here is a very first iteration on the idea of adding emscripten tests, suitable for testing how lz4 behaves when encode/decode buffers are allocated at low memory locations.

I can run this on my Mac via Homebrew as follows:

```sh
brew install emscripten
export LLVM=/usr/local/Cellar/emscripten/1.37.18/libexec/llvm/bin
export PATH=/usr/local/Cellar/emscripten/1.37.18/libexec/llvm/bin:$PATH
make -C tests test-emscripten
```

You should be able to observe the following:
- The test still fail after applying the first patch from #397 
- Applying the second patch from #397 will make the tests pass
- They will still pass if you increase `MESSAGE_MAX_BYTES` from 1024 to 4096
- However, they will fail again once you switch around `inpBuf` and `decBuf` in
  the `buffers` struct (leaving `MESSAGE_MAX_BYTES` at 4096), suggesting that there is still an issue unaddressed.

Do you think you can take it from here? Let me know if I can be of more help.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -548,7 +548,7 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
 
                 match = LZ4_getPositionOnHash(h, cctx->hashTable, tableType, base);
                 if (dict==usingExtDict) {
-                    if ((int)match < (int)source) {
+                    if ((ptrdiff_t)match < (ptrdiff_t)source) {
                         refDelta = dictDelta;
                         lowLimit = dictionary;
                     } else {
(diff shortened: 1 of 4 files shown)


Cyan4973:
That's a pretty good beginning, let's start investigating

Cyan4973:
The part which tries to launch `safari` on my laptop just fails,
but the console section seems to trigger a consistent scenario : 
```
block 0: PASS, decPtr=0xe6c
block 1: PASS, decPtr=0xe6d
block 2: PASS, decPtr=0x10d5
block 3: PASS, decPtr=0x1164
block 4: PASS, decPtr=0x149e
block 5: PASS, decPtr=0x1895
block 6: PASS, decPtr=0x1ac4
block 7: PASS, decPtr=0x1d48
block 8: PASS, decPtr=0x1ece
block 9: PASS, decPtr=0x2279
block 10: PASS, decPtr=0x2490
block 11: PASS, decPtr=0x257e
block 12: PASS, decPtr=0x2846
block 13: PASS, decPtr=0x2bbc
block 14: PASS, decPtr=0x2d95
block 15: PASS, decPtr=0x2e71
block 16: PASS, decPtr=0x31cc
FAIL: 2
```

jscheid:
@Cyan4973 try using chrome:

```sh
make -C tests test-emscripten EMRUN_FLAGS=--browser=chrome
```

I've just pushed a commit that fixes all issues I can find... with this change I can't make it fail anymore, no matter what the message size and which order the buffers appear.

Cyan4973:
There are 2 parts in this PR : 
- The emscripten test : this one is pretty interesting, as it adds a new test case to check, and with a bit of work, could be integrated into the LZ4 CI tests suite
- The fix : I'm sure it works for your use case, but unfortunately, I can see it fail for other use cases, where the force cast to `ptrdiff_t` doesn't end well.

So I guess I would prefer 2 separate issues.
The most promising one is the emscripten test, which is a pre-requisite to investigate a solution for it.

jscheid:
Hi @Cyan4973, sure, that makes sense. After rebasing onto current dev branch I can't get the emscripten tests to pass anymore anyway. New PR at #430

I'll close this one for now. Thanks again for looking into this.

PR #402, opened 2017-09-22
Fix dict segfault

felixhandte:
With my forthcoming changes to support using dictionaries from the command line, it's possible to induce a segfault at lz4frame.c:1642 (`memcpy(dctx->tmpOutBuffer, oldDictEnd - newDictSize, newDictSize);`). This is because when reading from a file, the first call to `LZ4F_decompress` is with just the magic number. `dstage_init` isn't reached in that call, so `dctx->tmpOutBuffer` isn't allocated.


Cyan4973:
Looks good !
Note that this patch will only be present in next release,
so, regarding your current work on the caller side,
you may have to ensure you don't trigger that case when using an older version of the library.

PR #403, opened 2017-09-28
Support Dictionaries on the Command Line

felixhandte:


programs/lz4io.c
```diff
@@ -395,8 +404,50 @@ typedef struct {
     void*  dstBuffer;
     size_t dstBufferSize;
     LZ4F_compressionContext_t ctx;
+    LZ4F_CDict* cdict;
 } cRess_t;
 
+static void LZ4IO_loadDict(char **dictionaryBuffer, size_t *dictionarySize) {
```
Cyan4973: Some coding style guidelines : 
1) prefer `void*` as buffer type. Keep `char*` to indicate it's an ascii/text content (typically for `printf()`).
2) I believe that this function objective is to deliver a buffer filled with dictionary content.
In which case, deliver it by value.
Prototype becomes : 
`void* LZ4IO_createDictBuffer(size_t* dictionarySize);`
`create` is an important keyword, that states that data a memory area has been allocated. It must be paired with a `free` later in the same context.
3) This function can fail, for whatever reason. Document it.
For functions which deliver allocated buffers, it's typical to consider result `NULL` as an error.

programs/lz4io.c
```diff
@@ -395,8 +404,50 @@ typedef struct {
...
+    FILE* dictionaryFile;
+    size_t blockSize = 64 KB;
+    size_t dictionaryBufferSize = blockSize;
+    size_t readSize;
+    *dictionarySize = 0;
+    *dictionaryBuffer = malloc(dictionaryBufferSize);
```
Cyan4973: `malloc()` can fail. In which case `*dictionaryBuffer == NULL`.
Always check `malloc()` result, and act accordingly.

programs/lz4io.c
```diff
@@ -395,8 +404,50 @@ typedef struct {
...
+
+    if (!g_dictionaryFilename) {
```
Cyan4973: Would it be possible to have an argument `const char* dictionaryFilename` in prototype instead ?

programs/lz4io.c
```diff
@@ -395,8 +404,50 @@ typedef struct {
...
+        EXM_THROW(25, "Dictionary error : no filename provided");
+    }
+
+    dictionaryFile = LZ4IO_openSrcFile(g_dictionaryFilename);
```
Cyan4973: `LZ4IO_openSrcFile()` can fail.
Check its result : https://github.com/lz4/lz4/blob/dev/programs/lz4io.c#L228

programs/lz4io.c
```diff
@@ -395,8 +404,50 @@ typedef struct {
...
+
+    do {
+        if (*dictionarySize + blockSize > dictionaryBufferSize) {
+            dictionaryBufferSize *= 2;
+            *dictionaryBuffer = realloc(*dictionaryBuffer, dictionaryBufferSize);
```
Cyan4973: `realloc()` is a pretty capricious function.
If it fails, `*dictionaryBuffer == NULL`, but buffer previously accessed through `*dictionaryBuffer` is still valid, but cannot be accessed anymore (since we just lost its pointer). So it cannot be freed, so it becomes a memory leak.

In this case though, it does not make a big different, because you `EXM_THROW()` right afterwards, which terminates the program. So no harm done.

Just keep in mind that using `realloc()` can be relatively tricky. Avoid it when not necessary.

programs/lz4io.c
```diff
@@ -395,8 +404,50 @@ typedef struct {
...
+
+    if (!g_dictionaryFilename) {
+        EXM_THROW(25, "Dictionary error : no filename provided");
+    }
+
+    dictionaryFile = LZ4IO_openSrcFile(g_dictionaryFilename);
+
+    do {
+        if (*dictionarySize + blockSize > dictionaryBufferSize) {
+            dictionaryBufferSize *= 2;
```
Cyan4973: An LZ4 dictionary cannot be larger than 64KB
(or more precisely, in this case, only the last 64KB are loaded).
So it's not useful to load more than 64 KB.

If the dictionary file is larger than 64 KB (which can easily be the case, since zstd dictionary generator creates 110 KB files by default), load only the last 64 KB into memory buffer.

It would also be useful to set some kind of "limit" : 
if someone provides a 2 GB file as dictionary, it's likely an error.

programs/lz4io.c
```diff
@@ -412,13 +463,21 @@ static cRess_t LZ4IO_createCResources(void)
...
     return ress;
 }
 
 static void LZ4IO_freeCResources(cRess_t ress)
 {
     free(ress.srcBuffer);
     free(ress.dstBuffer);
+
+    if (ress.cdict) {
+        LZ4F_freeCDict(ress.cdict);
```
Cyan4973: minor : for information, `LZ4F_freeCDict()` is compatible with free on `NULL`, so you could simplify this.
Btw, maybe this capability should be mentioned in `lz4frame_static.h`.

programs/lz4io.c
```diff
@@ -472,7 +531,12 @@ static int LZ4IO_compressFilename_extRess(cRess_t ress, const char* srcFileName,
     /* single-block file */
     if (readSize < blockSize) {
         /* Compress in single pass */
-        size_t const cSize = LZ4F_compressFrame(dstBuffer, dstBufferSize, srcBuffer, readSize, &prefs);
+        size_t cSize;
+        if (ress.cdict) {
```
Cyan4973: minor : `LZ4F_compressFrame_usingCDict()` with a `NULL` cdict as argument is exactly the same as `LZ4F_compressFrame()` : 
https://github.com/lz4/lz4/blob/dev/lib/lz4frame_static.h#L102
It's implemented this way by the way : 
https://github.com/lz4/lz4/blob/dev/lib/lz4frame.c#L385

So it allows you to simplify this branch.

programs/lz4io.c
```diff
@@ -488,7 +552,12 @@ static int LZ4IO_compressFilename_extRess(cRess_t ress, const char* srcFileName,
     /* multiple-blocks file */
     {
         /* Write Archive Header */
-        size_t headerSize = LZ4F_compressBegin(ctx, dstBuffer, dstBufferSize, &prefs);
+        size_t headerSize;
+        if (ress.cdict) {
```
Cyan4973: Same idea here : 
`LZ4F_compressBegin_usingCDict()` with `cdict == NULL` is exactly the same as `LZ4F_compressBegin()` : 
https://github.com/lz4/lz4/blob/dev/lib/lz4frame.c#L598

If this property is not clear, maybe it could be added in comment in `lz4frame_static.h`.

programs/lz4io.c
```diff
@@ -786,7 +870,11 @@ static unsigned long long LZ4IO_decompressLZ4F(dRess_t ress, FILE* srcFile, FILE
     {   size_t inSize = MAGICNUMBER_SIZE;
         size_t outSize= 0;
         LZ4IO_writeLE32(ress.srcBuffer, LZ4IO_MAGICNUMBER);
-        nextToLoad = LZ4F_decompress(ress.dCtx, ress.dstBuffer, &outSize, ress.srcBuffer, &inSize, NULL);
+        if (g_useDictionary) {
```
Cyan4973: same idea here

programs/lz4io.c
```diff
@@ -805,7 +893,11 @@ static unsigned long long LZ4IO_decompressLZ4F(dRess_t ress, FILE* srcFile, FILE
             /* Decode Input (at least partially) */
             size_t remaining = readSize - pos;
             decodedBytes = ress.dstBufferSize;
-            nextToLoad = LZ4F_decompress(ress.dCtx, ress.dstBuffer, &decodedBytes, (char*)(ress.srcBuffer)+pos, &remaining, NULL);
+            if (g_useDictionary) {
```
Cyan4973: and here

tests/Makefile
```diff
@@ -253,6 +253,21 @@ test-lz4-basic: lz4 datagen unlz4 lz4cat
 	$(LZ4) -BX tmp-tlb-hw -c -q | $(LZ4) -tv  # test block checksum
 	@$(RM) tmp-tlb*
 
+test-lz4-dict: lz4 datagen
```
Cyan4973: 👍

programs/lz4io.c
felixhandte: I meant to ask, what is meant by the number in the `EXM_THROW` macro?

Cyan4973: This is the error code that the program will return (`exit(errorCode)`).
It will also be displayed in the error message.
It's a cheap way to find were the error was triggered, especially when 2 error messages are similar.

programs/lz4io.c
felixhandte: Yeah, I thought about doing that. So we don't necessarily know the size of the file we're reading. (It could be a pipe.) So we probably have to read the whole thing anyways.

But I guess I could read into a circular buffer instead.

programs/lz4io.c
```diff
@@ -395,8 +405,83 @@ typedef struct {
...
+        dictLen = LZ4_MAX_DICT_SIZE;
+    }
+
+    *dictSize = dictLen;
+
+    dictStart = (circularBufSize + dictEnd - dictLen) % circularBufSize;
+
+    if (dictStart == 0) {
+        // We're in the simple case where the dict starts at the beginning of our circular buffer.
+        dictBuf = realloc(circularBuf, dictLen ? dictLen : 1);
```
terrelln: We don't need this `realloc`, we can just say

```c
dictBuf = circularBuf;
circularBuf = NULL;
```

programs/lz4io.c
```diff
@@ -395,8 +405,83 @@ typedef struct {
...
+
+    dictFile = LZ4IO_openSrcFile(dictFilename);
+    if (!dictFile) EXM_THROW(25, "Dictionary error : could not open dictionary file");
+
+    if (strcmp(dictFilename, stdinmark)) {
+        UTIL_fseek(dictFile, -LZ4_MAX_DICT_SIZE, SEEK_END);
+    }
+
+    do {
+        readSize = fread((char*)circularBuf + dictEnd, (size_t)1, circularBufSize - dictEnd, dictFile);
```
terrelln: No need to cast the `1` to a size_t.

programs/lz4io.c
```diff
@@ -395,8 +405,83 @@ typedef struct {
...
+        dictEnd = (dictEnd + readSize) % LZ4_MAX_DICT_SIZE;
```
terrelln: Should this be `% circularBufSize`?

programs/lz4io.c
```diff
@@ -395,8 +405,83 @@ typedef struct {
...
+        dictBuf = realloc(circularBuf, dictLen ? dictLen : 1);
+        if (!dictBuf) EXM_THROW(25, "Allocation error : not enough memory");
+
+        circularBuf = NULL;
+    } else {
+        // Otherwise, we will alloc a new buffer and copy our dict into that.
+        dictBuf = malloc(dictLen ? dictLen : 1);
+        if (!dictBuf) EXM_THROW(25, "Allocation error : not enough memory");
+
+        if (dictStart + dictLen <= circularBufSize) {
```
terrelln: This isn't possible, right? If dictLen < circularBufferSize, then `dictStart` must be zero. And we already check for `dictStart == 0`.

programs/lz4io.c
```diff
@@ -395,8 +405,83 @@ typedef struct {
...
+
+    if (dictStart == 0) {
+        // We're in the simple case where the dict starts at the beginning of our circular buffer.
+        dictBuf = realloc(circularBuf, dictLen ? dictLen : 1);
+        if (!dictBuf) EXM_THROW(25, "Allocation error : not enough memory");
+
+        circularBuf = NULL;
+    } else {
+        // Otherwise, we will alloc a new buffer and copy our dict into that.
+        dictBuf = malloc(dictLen ? dictLen : 1);
```
terrelln: `dictLen == LZ4_MAX_DICT_SIZE` here right?

programs/lz4io.c
```diff
@@ -395,8 +405,83 @@ typedef struct {
...
 } cRess_t;
 
+static void* LZ4IO_createDict(const char* dictFilename, size_t *dictSize) {
+    size_t readSize;
+    size_t dictEnd = 0;
+    size_t dictLen = 0;
+    size_t dictStart;
+    size_t circularBufSize = LZ4_MAX_DICT_SIZE;
+    void* circularBuf;
+    void* dictBuf;
```
terrelln: You can make these `char*`, but still return a `void*`.

programs/lz4io.c
felixhandte: The `realloc` is to shrink the buffer if the dict is smaller than 64KB.

programs/lz4io.c
felixhandte: Yes, that must be the case.

programs/lz4io.c
felixhandte: Yeah, you're right.

programs/lz4io.c
felixhandte: Maybe that's not valuable, though?

terrelln: We free the buffer very soon after, when we create the CDict right?

In any case, I'm not too worried by over allocating by up to 64 KB.

felixhandte: Ah, that's a good point. We do on the compress path, but not on the decompress path. Ok, I'll axe it.

felixhandte:
The build is failing in gen_manual.cpp, which I did not touch, saying it can't find `#include<iostream>`. Not sure what to do about that.

programs/lz4io.c
```diff
@@ -395,8 +405,79 @@ typedef struct {
...
+    dictStart = (circularBufSize + dictEnd - dictLen) % circularBufSize;
+
+    if (dictStart == 0) {
+        /* We're in the simple case where the dict starts at the beginning of our circular buffer. */
+        dictBuf = circularBuf;
+        circularBuf = NULL;
+    } else {
+        /* Otherwise, we will alloc a new buffer and copy our dict into that. */
+        dictBuf = (char *) malloc(dictLen);
+        if (!dictBuf) EXM_THROW(25, "Allocation error : not enough memory");
```
terrelln:
```
lz4io.c:455:28: warning: Call to 'malloc' has an allocation size of 0 bytes
        dictBuf = (char *) malloc(dictLen);
                           ^~~~~~~~~~~~~~~
1 warning generated.
```

We should special case `dictLen == 0`, otherwise we may (or may not) trigger this allocation error.

felixhandte:
Ok, yeah. I put the ternary check back in. *shrug*

programs/lz4io.c
```diff
@@ -395,8 +405,79 @@ typedef struct {
...
+    *dictSize = dictLen;
+
+    dictStart = (circularBufSize + dictEnd - dictLen) % circularBufSize;
+
+    if (dictStart == 0) {
+        /* We're in the simple case where the dict starts at the beginning of our circular buffer. */
+        dictBuf = circularBuf;
+        circularBuf = NULL;
+    } else {
+        /* Otherwise, we will alloc a new buffer and copy our dict into that. */
```
Cyan4973: 👍

programs/lz4cli.c
```diff
@@ -399,6 +402,21 @@ int main(int argc, const char** argv)
...
+                        if (i + 1 == argc) {
+                            /* there is no next arg */
+                            badusage(exeName);
+                        }
+                        dictionary_filename = argv[++i];
+                    } else {
+                        /* path follows immediately */
+                        dictionary_filename = argument + 1;
+                    }
+                    break_inner_loop = 1;
```
Cyan4973: For the same effect, you could also move `argument` to the end of the string, using : 
`argument += strlen(argument) - 1;`
Outcome is the same (breaking inner loop, moving to next command line argument).
Difference is, it avoids introducing another mutable variable, which is one more state to track during debugging.

*Note* : `- 1` because `lz4cli` uses the convention of letting the main loop move argument forward by 1 character. This is is not a great choice btw : it has been changed in `zstdcli`. One day, it will be useful for both cli to use the same convention.

PR #406, opened 2017-10-10
Use the Dictionary Hash Table in Place

felixhandte:
## Motivation

The *status quo ante* workflows for dictionary compression in LZ4 are the following:

* For each compression, load the dictionary buffer into the working context via `LZ4_loadDict` and then call `LZ4_compress_fast_continue`.
* Load the dictionary buffer once into a separate context via `LZ4_loadDict`. For each compression, `memcpy` the dictionary context into the working context and then call `LZ4_compress_fast_continue`.

For compressing small blobs (where dictionary compression is most impactful), these ancillary operations (loading or even just copying the dictionary context) can come to dominate the cost of the compression.

In fact, for small compressions, whether dictionary based or not, any setup operation which involves touching the whole hashtable (whether copying it or zeroing it) will end up being the majority of time spent. It is therefore desirable to find a way to prepare the table and context for a new compression (dictionary based or not) without having to touch the whole table.

## Summary

This changeset makes two broad changes:

1. The `LZ4_stream_t` context no longer always needs to be reset before beginning a new compression. It can now be used "dirty".
2. Rather than being written into the working context, a dictionary context can be referred to from a working context and used where it lies.

In order to achieve these high level changes, this stack also makes a number of other internal changes concerning the management of the `LZ4_stream_t` compression context.

Here are the user visible ones:

* `LZ4F_compressFrame_usingCDict` now takes a `LZ4F_cctx`.
* In order to make room in the `LZ4_stream_t`, the `bufferStart` field was removed. Obsolete streaming functions that depended on `bufferStart`---`LZ4_init`, `LZ4_resetStreamState`, `LZ4_create`, and `LZ4_slideInputBuffer`---also had to be removed.

### Compressed Size

No changes should be observed. For all inputs and patterns of calls, LZ4 should produce the exact same compressed blob it used to. To the extent covered by tests, this appears to be the case.

#### Summary

* LZ4F performance for non-small data is uniformly improved.
* LZ4 performance for non-small data is, in general, very slightly improved.
* Some mid-range performance (4KB - 16KB) is worsened.
* Small data performance is greatly improved when reusing a context.

To provide detail, pictures are worth a thousand words. These chart before (red) and after (green) 75th percentile compression speed for each text in the silesia corpus, truncated to the various marked lengths.

Benchmarking was performed on an `Intel Xeon E5-2680 v4 @ 2.40GHz` with `clang version 6.0.0` and `gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-16)` with flags `-O3 -mtune=native`.

## Further Work

This work has many possible extensions, only a few of which have been investigated and rejected. A list is forthcoming...


lib/lz4.c
```diff
@@ -548,7 +548,13 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
 
                 match = LZ4_getPositionOnHash(h, cctx->hashTable, tableType, base);
                 if (dict==usingExtDict) {
-                    if (match < (const BYTE*)source) {
+                    if (match == 0) {
```
Cyan4973: OK, so according to this logic,
the code only looks for a match in the dictionary if it could not find one in the current data segment.
Well, I think it works.
For the speed / compression trade-off, we'll need a few test corpus to benchmark.

Cyan4973:
For dictionary compression benchmark,
we have been using a pack of public github records in the past for `zstd`.
Can be downloaded in attachment here : 
https://github.com/facebook/zstd/releases/tag/v1.1.3

lib/lz4.h
```diff
@@ -391,7 +393,7 @@ typedef struct {
  *        this definition is not API/ABI safe,
  *        it may change in a future version !
  */
-#define LZ4_STREAMSIZE_U64 ((1 << (LZ4_MEMORY_USAGE-3)) + 4)
+#define LZ4_STREAMSIZE_U64 ((1 << (LZ4_MEMORY_USAGE-3)) + 5)
```
Cyan4973: One way to avoid increasing the size of `LZ4_stream_t_internal`
is to take the slot currently occupied by `bufferStart`.
This pointer is no longer useful.
Removing it is likely going to trigger some side-effects in some obsolete functions,
but I believe it's fine, these functions have been obsolete for many years by now.
For discussion.

lib/lz4.c
```diff
@@ -461,11 +461,11 @@ LZ4_FORCE_INLINE void LZ4_putPosition(const BYTE* p, void* tableBase, tableType_
     LZ4_putPositionOnHash(p, h, tableBase, tableType, srcBase);
 }
 
-static const BYTE* LZ4_getPositionOnHash(U32 h, void* tableBase, tableType_t tableType, const BYTE* srcBase)
+static const BYTE* LZ4_getPositionOnHash(U32 h, const void* tableBase, tableType_t tableType, const BYTE* srcBase)
 {
-    if (tableType == byPtr) { const BYTE** hashTable = (const BYTE**) tableBase; return hashTable[h]; }
-    if (tableType == byU32) { const U32* const hashTable = (U32*) tableBase; return hashTable[h] + srcBase; }
-    { const U16* const hashTable = (U16*) tableBase; return hashTable[h] + srcBase; }   /* default, to ensure a return */
+    if (tableType == byPtr) { const BYTE* const* hashTable = (const BYTE* const*) tableBase; return hashTable[h]; }
```
felixhandte: This feels silly, but otherwise I get "`error: cast from 'const void *' to 'const unsigned char **' drops const qualifier`".

lib/lz4.h
felixhandte: If we feel we can remove those functions, then yeah, we could do that. I looked at them though and couldn't think of an easy way to remove `bufferStart` and retain their functionality.

Cyan4973: I don't think it's possible to retain their functionality.
I suspect no one uses these deprecated functions.

lib/lz4.c
Cyan4973: Something a bit complex : 
in `lz4.c`, the hash table can contain either : 
- 32 bits indexes (streaming)
- 16 bits indexes (small data, 64KB, no streaming)
- 32-bits pointers (32-bits mode only)

So it's a kind of template emulation (since template are not available in C).

Because of this property, you might have to change the type of `dictHashTable` into `const void*`,
in order to avoid potential strict aliasing issues.
I believe it's a minor change with no adverse consequence in your code.

Upside is : it's allowed to cast a `void*` pointer into any other type.
Any other conversion is more dangerous, as it can be transformed into a bug when compiler's optimisation decides to use strict aliasing for some dirty performance trick.

felixhandte:
CircleCI failure is "sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test" failing due to an expired GPG key. :disappointed:

Cyan4973:
Indeed, CI systems have occasional errors.
Let's see how to restart this test ...

Cyan4973:
Apparently restarting this test is only possible from your side.

lib/lz4.h
```diff
@@ -461,12 +470,6 @@ LZ4LIB_API LZ4_DEPRECATED("use LZ4_compress_fast_continue() instead") int LZ4_co
 LZ4LIB_API LZ4_DEPRECATED("use LZ4_decompress_fast() instead") int LZ4_uncompress (const char* source, char* dest, int outputSize);
 LZ4LIB_API LZ4_DEPRECATED("use LZ4_decompress_safe() instead") int LZ4_uncompress_unknownOutputSize (const char* source, char* dest, int isize, int maxOutputSize);
 
-/* Obsolete streaming functions; use new streaming interface whenever possible */
```
Cyan4973: It's always risky to remove API entry points.
I hope we won't get too many issues from removing these ones.
It will have to be clearly labelled in release documentation.

lib/lz4frame.c
```diff
@@ -316,23 +316,20 @@ size_t LZ4F_compressFrameBound(size_t srcSize, const LZ4F_preferences_t* prefere
...
-size_t LZ4F_compressFrame_usingCDict(void* dstBuffer, size_t dstCapacity,
+size_t LZ4F_compressFrame_usingCDict(LZ4F_cctx* cctx,
+                                     void* dstBuffer, size_t dstCapacity,
                                const void* srcBuffer, size_t srcSize,
                                const LZ4F_CDict* cdict,
                                const LZ4F_preferences_t* preferencesPtr)
 {
     LZ4F_cctx_t cctxI;
-    LZ4_stream_t lz4ctx;   /* pretty large on stack */
+    LZ4_stream_t lz4ctxI;
```
Cyan4973: This will use stack space even when a valid `cctx` is provided

lib/lz4frame.c
```diff
@@ -345,32 +342,37 @@ size_t LZ4F_compressFrame_usingCDict(void* dstBuffer, size_t dstCapacity,
     if (srcSize <= LZ4F_getBlockSize(prefs.frameInfo.blockSizeID))
         prefs.frameInfo.blockMode = LZ4F_blockIndependent;   /* only one block => no need for inter-block link */
 
-    if (prefs.compressionLevel < LZ4HC_CLEVEL_MIN) {
-        cctxI.lz4CtxPtr = &lz4ctx;
-        cctxI.lz4CtxLevel = 1;
-    }  /* fast compression context pre-created on stack */
+    if (cctx == NULL) {
```
Cyan4973: I believe this could be done differently : 
make sure `LZ4F_compressFrame()` provides a valid context to `LZ4F_compressFrame_usingCDict()`.

lib/lz4.h
```diff
@@ -354,9 +361,10 @@ LZ4LIB_API int LZ4_decompress_fast_usingDict (const char* src, char* dst, int or
 typedef struct {
     uint32_t hashTable[LZ4_HASH_SIZE_U32];
     uint32_t currentOffset;
-    uint32_t initCheck;
+    uint16_t initCheck;
+    uint16_t tableType;
     const uint8_t* dictionary;
-    uint8_t* bufferStart;   /* obsolete, used for slideInputBuffer */
+    const void* dictCtx;
```
Cyan4973: could the type be more precise than `void*` ?

lib/lz4frame.c
```diff
@@ -541,16 +543,34 @@ size_t LZ4F_compressBegin_usingCDict(LZ4F_cctx* cctxPtr,
         /* frame init only for blockLinked : blockIndependent will be init at each block */
         if (cdict) {
             if (cctxPtr->prefs.compressionLevel < LZ4HC_CLEVEL_MIN) {
-                memcpy(cctxPtr->lz4CtxPtr, cdict->fastCtx, sizeof(*cdict->fastCtx));
+                LZ4_stream_t_internal* internal_ctx = &((LZ4_stream_t*)cctxPtr->lz4CtxPtr)->internal_donotuse;
+                if (internal_ctx->initCheck || internal_ctx->currentOffset > 1 GB) {
```
Cyan4973: `initCheck` : it's unclear to me where this value was set.

tests/frametest.c
```diff
@@ -730,15 +733,17 @@ int basicTests(U32 seed, double compressibility)
 
 static void locateBuffDiff(const void* buff1, const void* buff2, size_t size, unsigned nonContiguous)
 {
-    int p=0;
+    size_t p=0;
     const BYTE* b1=(const BYTE*)buff1;
     const BYTE* b2=(const BYTE*)buff2;
     if (nonContiguous) {
-        DISPLAY("Non-contiguous output test (%i bytes)\n", (int)size);
+        DISPLAY("Non-contiguous output test (%zd bytes)\n", size);
```
Cyan4973: beware of the portability of `%zd`,
notably with Windows compilers.

tests/frametest.c
```diff
@@ -730,15 +733,17 @@ int basicTests(U32 seed, double compressibility)
...
         return;
     }
-    while (b1[p]==b2[p]) p++;
-    DISPLAY("Error at pos %i/%i : %02X != %02X \n", p, (int)size, b1[p], b2[p]);
+    while (p < size && b1[p]==b2[p]) p++;
+    if (p != size) {
```
Cyan4973: 👍  good point.
Also : display something if `p==size`

lib/lz4.c
```diff
@@ -537,46 +538,83 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
...
                  const U32 acceleration)
 {
     const BYTE* ip = (const BYTE*) source;
-    const BYTE* base;
+
+    /* If the table hasn't been used, it's guaranteed to be zeroed out, and is
+     * therefore safe to use no matter what mode we're in. Otherwise, we figure
+     * out if it's safe to leave as is or whether it needs to be reset.
+     */
+    const int resetTable = cctx->tableType != LZ4_unusedTable && (
```
Cyan4973: Try to indent nested test conditions for clarity

lib/lz4.c
```diff
@@ -537,46 +538,83 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
...
     {
     case noDict:
     default:
-        base = (const BYTE*)source;
         lowLimit = (const BYTE*)source;
         break;
     case withPrefix64k:
-        base = (const BYTE*)source - cctx->currentOffset;
-        lowLimit = (const BYTE*)source - cctx->dictSize;
+        lowLimit = (const BYTE*)source - dictSize;
```
Cyan4973: I guess this is now compatible with any dictSize, including < 64K ?

lib/lz4.c
```diff
@@ -537,46 +538,83 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
...
+        break;
+    case usingExtDictCtx:
         lowLimit = (const BYTE*)source;
         break;
     }
     if ((tableType == byU16) && (inputSize>=LZ4_64Klimit)) return 0;   /* Size too large (not within 64K limit) */
+
+    if (resetTable) {
+      DEBUGLOG(4, "Resetting table in %p", cctx);
+      MEM_INIT(cctx->hashTable, 0, LZ4_HASHTABLESIZE);
```
Cyan4973: `hashTable` used to be `memset()` outside of `LZ4_compress_generic()`.
This is a change of logic.
Are you sure you avoid any scenario involve either double `memset()`, or no `memset()` ?

lib/lz4.c
```diff
@@ -602,8 +640,19 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
                 if (unlikely(forwardIp > mflimit)) goto _last_literals;
 
                 match = LZ4_getPositionOnHash(h, cctx->hashTable, tableType, base);
-                if (dict==usingExtDict) {
+                if (dictDirective == usingExtDictCtx) {
                     if (match < (const BYTE*)source) {
+                        /* there was no match, try the dictionary */
+                        /* TODO: use precalc-ed hash? */
```
Cyan4973: That's an interesting optimization.
Actually, there are 2 scenarios : 
- use same hashing formula, hence re-use hash value
- use this opportunity to have _different_ hashing formula. For example, if dictionary uses a `byU16` table, while streaming data use `byU32` instead, dictionary hash values use one more bit. It's also possible deliberately target different `minmatch` sizes (LZ4 already automatically selects 4 bytes or 5 bytes depending on scenario).

lib/lz4.c
```diff
@@ -602,8 +640,19 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
...
+                        match = LZ4_getPosition(ip, dictCtx->hashTable, byU32, dictBase);
+                        refDelta = dictDelta;
+                        lowLimit = dictionary;
+                    } else {
+                        refDelta = 0;
+                        lowLimit = (const BYTE*)source;
+                    }
+                } else if (dictDirective==usingExtDict) {
+                    if (match < (const BYTE*)source && dictionary != NULL) {
```
Cyan4973: Is it possible to have both `(dictDirective==usingExtDict)` and `dictionary == NULL` ?
If not, this condition could be replaced by an `assert()`, in order to be removed from optimised binary (this is the hot loop, every branch matters).

lib/lz4.c
felixhandte: Yeah, this hasn't actually changed anything. `base` and `lowLimit` end up with the same values.

lib/lz4.c
felixhandte: The issue with this is that if we're going to copy the dict table into the compression ctx, we need it to be `tableType`-compatible (i.e., `byU32`). If we're going to keep them separate, it would be better if it were `byU16`. We don't know which scenario we're in though until we know the size of the blob we're compressing, which is not known at dict load time. So I'm not sure there's a good solution.

lib/lz4.c
felixhandte: Apparently it happens, though I don't know when/why.

lib/lz4.c
Cyan4973: Is there still a size threshold after which it's preferable to copy the table rather than using it in-place ?

lib/lz4.c
Cyan4973: This should be analysed then.
If this combination can happen, it should be a legitimate construction.
There might be an optimisation opportunity left there.

lib/lz4.c
felixhandte: Right now, I'm being pretty conservative and copying the table over if the blob to compress is larger than 2KB. I don't think there's a clear line, but it's somewhere between 1KB and 32KB.

Cyan4973: We will have to measure again this threshold once optimisation pass is completed.

Cyan4973:
try to merge recent changes from `dev` into your feature branch, to fix conflicts

lib/lz4.c
```diff
@@ -692,16 +747,27 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
...
+            if (match < (const BYTE*)source) {
+                refDelta = dictDelta;
+                lowLimit = dictLowLimit;
             } else {
                 refDelta = 0;
                 lowLimit = (const BYTE*)source;
         }   }
         LZ4_putPosition(ip, cctx->hashTable, tableType, base);
-        if ( ((dictIssue==dictSmall) ? (match>=lowRefLimit) : 1)
+        if ( match >= lowRefLimit
```
Cyan4973: I guess this test is now run at each loop,
while previously, it was only active when `(dictIssue == dictSmall)`.
This can make a difference.

felixhandte: Yes, unfortunately, by not resetting the table as frequently (and by starting the context with non-zero `currentOffset`), there are pretty limited situations in which this optimization can be restored. The branch `https://github.com/felixhandte/lz4/tree/ref-dict-table-test-2` tries to restore the check, but benchmarking didn't show an improvement.

lib/lz4.c
```diff
@@ -602,18 +646,29 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
                 if (unlikely(forwardIp > mflimit)) goto _last_literals;
 
                 match = LZ4_getPositionOnHash(h, cctx->hashTable, tableType, base);
-                if (dict==usingExtDict) {
+                if (dictDirective == usingExtDictCtx) {
+                    if (match < (const BYTE*)source) {
```
terrelln: Can we invert these conditions to avoid repetition in the inner `else`?

lib/lz4.c
```diff
@@ -692,16 +747,27 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
 
         /* Test next position */
         match = LZ4_getPosition(ip, cctx->hashTable, tableType, base);
-        if (dict==usingExtDict) {
+        if (dictDirective == usingExtDictCtx) {
             if (match < (const BYTE*)source) {
+                /* there was no match, try the dictionary */
+                /* TODO: use precalc-ed hash? */
```
terrelln: Same 2 comments here as above.

lib/lz4frame.c
```diff
@@ -394,9 +395,41 @@ size_t LZ4F_compressFrame(void* dstBuffer, size_t dstCapacity,
                     const void* srcBuffer, size_t srcSize,
                     const LZ4F_preferences_t* preferencesPtr)
 {
-    return LZ4F_compressFrame_usingCDict(dstBuffer, dstCapacity,
-                                         srcBuffer, srcSize,
-                                         NULL, preferencesPtr);
+    size_t result;
+#if (LZ4_HEAPMODE)
+    LZ4F_cctx_t *cctxPtr;
+    LZ4F_createCompressionContext(&cctxPtr, LZ4F_VERSION);
```
Cyan4973: check if context creation is actually successful

lib/lz4frame.c
```diff
@@ -46,11 +46,25 @@ You can contact the author at :
...
+/*-************************************
+*  Tuning parameters
+**************************************/
+/*
+ * LZ4_HEAPMODE :
+ * Select how default compression functions will allocate memory for their hash table,
+ * in memory stack (0:default, fastest), or in memory heap (1:requires malloc()).
+ */
+#ifndef LZ4_HEAPMODE
+#  define LZ4_HEAPMODE 0
```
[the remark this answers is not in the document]
Cyan4973: Good point.
There is one small advantage for keeping the name `LZ4_HEAPMODE`  : 
It becomes possible to compile with `-DLZ4_HEAPMODE=1`, and it's applicable to the entire `lz4` library / binary.
Otherwise, one would need to set both `-DLZ4_HEAPMODE=1 -DLZ4F_HEAPMODE=1`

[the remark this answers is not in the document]
terrelln: I slightly prefer `LZ4F_`, since thats what everything in the API is prefixed with, but I don't have a strong preference.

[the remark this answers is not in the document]
Cyan4973: OK, there is a narrow advantage to `LZ4F_`, due to consistency.
So let's switch to `LZ4F_HEAPMODE`.

lib/lz4frame.c
Cyan4973: This comment is actually pretty important.
You really have to check that `cctxPtr != NULL`,
otherwise, the next instruction, which is `LZ4F_compressFrame_usingCDict()`, will segfault

lib/lz4.c
```diff
@@ -723,41 +794,76 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
...
 
 
 int LZ4_compress_fast_extState(void* state, const char* source, char* dest, int inputSize, int maxOutputSize, int acceleration)
+{
+  LZ4_resetStream((LZ4_stream_t*)state);
+  return LZ4_compress_fast_safeExtState(state, source, dest, inputSize, maxOutputSize, acceleration);
+}
+
+
+int LZ4_compress_fast_safeExtState(void* state, const char* source, char* dest, int inputSize, int maxOutputSize, int acceleration)
```
Cyan4973: `_safeExtState()` ?
Can you explain the logic behind the name ?

Cyan4973: Note : I could read the associated code comment in `lz4.h`.
Note that you can copy/paste code comments from `.h` into `.c` when they make sense to explain a function objective. It's not required either, but it can make sense when it's not obvious.

In this case, the naming does not make the scope of this function obvious, so it needs a comment to explain what it does.

It's also possible to change the name of the function.

For example, `LZ4_compress_fast_noReset()` ? or `_noStateReset` ?

lib/lz4.c
```diff
@@ -508,20 +509,48 @@ LZ4_FORCE_INLINE void LZ4_putPosition(const BYTE* p, void* tableBase, tableType_
...
+LZ4_FORCE_INLINE void LZ4_resetTable(
+        LZ4_stream_t_internal* const cctx,
+        const int inputSize,
+        const tableType_t tableType,
+        const dict_directive dictDirective) {
+  /* If the table hasn't been used, it's guaranteed to be zeroed out, and is
+   * therefore safe to use no matter what mode we're in. Otherwise, we figure
+   * out if it's safe to leave as is or whether it needs to be reset.
+   */
+  if (cctx->tableType != unusedTable && (
```
Cyan4973: This is a long `if` condition.

Try to arrange your tests in a way which is easier to read.
I would recommend : 
- Use indentation for each level of `( )`, so we can follow which test is paired with which one
- Prefer combination signs `&&` `||` at beginning of statement (when applicable)

For readability, you might even prefer to name some of these tests, 
just because a descriptive name is easier to read and combine.

lib/lz4.c
```diff
@@ -723,41 +794,76 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
...
     }
 }
 
 
 int LZ4_compress_fast(const char* source, char* dest, int inputSize, int maxOutputSize, int acceleration)
 {
+    int result;
 #if (LZ4_HEAPMODE)
-    void* ctxPtr = ALLOCATOR(1, sizeof(LZ4_stream_t));   /* malloc-calloc always properly aligned */
+    LZ4_stream_t* ctxPtr = ALLOC(sizeof(LZ4_stream_t));   /* malloc-calloc always properly aligned */
```
Cyan4973: Ensure that `ctxPtr!=NULL`

lib/lz4.c
```diff
@@ -723,41 +794,76 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
...
 #else
     LZ4_stream_t ctx;
-    void* const ctxPtr = &ctx;
+    LZ4_stream_t* const ctxPtr = &ctx;
 #endif
-
-    int const result = LZ4_compress_fast_extState(ctxPtr, source, dest, inputSize, maxOutputSize, acceleration);
+    ctxPtr->internal_donotuse.initCheck = 0;
+    ctxPtr->internal_donotuse.tableType = byPtr; /* always triggers a reset */
```
Cyan4973: If the intention is to trigger a reset, why not using `LZ4_compress_fast_extState()`, which ensures it ?

lib/lz4.h
```diff
@@ -102,8 +102,8 @@ extern "C" {
 #define LZ4_EXPAND_AND_QUOTE(str) LZ4_QUOTE(str)
 #define LZ4_VERSION_STRING LZ4_EXPAND_AND_QUOTE(LZ4_LIB_VERSION)
 
-LZ4LIB_API int LZ4_versionNumber (void);  /**< library version number; useful to check dll version */
-LZ4LIB_API const char* LZ4_versionString (void);   /**< library version string; unseful to check dll version */
+LZ4LIB_API int LZ4_versionNumber (void);  /**< library version number; to be used when checking dll version */
```
Cyan4973: was `lz4.h` updated from `dev` ?

Some of these comments look like older version comments.

lib/lz4frame.c
```diff
@@ -188,7 +202,15 @@ typedef struct LZ4F_cctx_s
...
     XXH32_state_t xxh;
     void*  lz4CtxPtr;
-    U32    lz4CtxLevel;   /* 0: unallocated;  1: LZ4_stream_t;  3: LZ4_streamHC_t */
+    /* lz4CtxLevel records both what size of memory has been allocated into the
+     * ctx pointer field and what it's being used as.
+     * - The low bit (& 1) value indicates whether the ctx is being used for
+     *   hc (1) or not (0).
+     * - The next bit (& 2) indicates whether the allocated memory is big
+     *   enough for a non-hc context.
+     * - The next bit (& 4) indicates whether the allocated memory is big
```
Cyan4973: Don't we have `&4` -> implies `&2` ?

tests/fuzzer.c
```diff
@@ -572,16 +572,22 @@ static int FUZ_test(U32 seed, U32 nbCycles, const U32 startCycle, const double c
         FUZ_DISPLAYTEST;
         dict -= (FUZ_rand(&randState) & 0xF) + 1;   /* Separation, so it is an ExtDict */
         if (dict < (char*)CNBuffer) dict = (char*)CNBuffer;
+        /* we depend on exactly reproducing the compression operation,
```
Cyan4973: For my understanding : 
why would a difference in `resetStream()` produce a difference in compression result ?

felixhandte: That's an interesting question. Here's the best theory I have (with @terrelln's help):

The fundamental difference is that when you start with an empty stream, hash table misses cause match checks against the beginning of the stream, while if you have a used stream--with a non-zero `currentOffset`--no match check occurs. If a hash miss points to the beginning of the stream and the bytes actually match, you will get a different output compared to a non-zero `currentOffset` stream where the hash miss doesn't trigger a match check.

This almost doesn't matter, since the starting position is hashed and written into the table at the beginning of the compression. However, because the hash is on 5 bytes and the initial match check is on 4 bytes, any position in which the first 4 bytes match the beginning of the stream but the fifth byte doesn't may hit a hash miss which hit this difference in behavior--the 0-offset stream will check against the beginning of the stream and find a match, which the non-0 stream won't.

lib/lz4frame.c
felixhandte: Yes. I can add a comment to that effect.

lib/lz4.h
felixhandte: Yeah, it looks like there was a rebase problem. I'll fix it.

tests/fuzzer.c
terrelln: We can verify if this is the case by adding a loop like this to the beginning of the compression function and see if the test failure goes away. This will put position 0 on every value that could hash to it and actually match.

```c
if (ip == base) {
  BYTE bytes[5];
  memcpy(bytes, ip, 4); // before ip++
  for (U32 b = 0; b < 256; ++b) {
    bytes[4] = (BYTE)b;
    LZ4_putPosition(bytes, cctx->hashTable, tableType, bytes);
  }
}
```

lib/lz4frame.c
Cyan4973: Or, you could use an assignment scheme which makes it impossible to generate a bogus statement by accident, such as "large enough for hc, but not large enough for fast".

What about a 3-states value `{0,1,2}` ?

lib/lz4.c
```diff
@@ -602,18 +646,29 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
                 if (unlikely(forwardIp > mflimit)) goto _last_literals;
 
                 match = LZ4_getPositionOnHash(h, cctx->hashTable, tableType, base);
-                if (dict==usingExtDict) {
+                if (dictDirective == usingExtDictCtx) {
+                    if (match < (const BYTE*)source) {
+                        /* there was no match, try the dictionary */
+                        /* TODO: use precalc-ed hash? */
```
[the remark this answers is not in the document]
felixhandte: I tried explicitly using `LZ4_getPositionOnHash` when `tableType == byU32` and saw no improvement. So the compiler must already be deduping the operations in that case.

lib/lz4.c
felixhandte: That means performing the `match > source` branch even when unnecessary, though.

lib/lz4.c
```diff
@@ -446,9 +447,9 @@ static const U32 LZ4_skipTrigger = 6;  /* Increase this value ==> compression ru
 *  Local Structures and types
 **************************************/
 typedef enum { notLimited = 0, limitedOutput = 1 } limitedOutput_directive;
-typedef enum { byPtr, byU32, byU16 } tableType_t;
+typedef enum { unusedTable = 0, byPtr = 1, byU32 = 2, byU16 = 3 } tableType_t;
```
Cyan4973: minor detail : 
My understanding is that `unusedTable` is actually used to describe a table which content has been reset to `0`. I'm wondering if `unusedTable` is best wording to convey this message.

I believe you want to track when a table is in 0-state because in such case it can be converted into anything. In contrast, if it's already filled with something, special care is needed to ensure next usage interprets data correctly.

So, maybe `clearedTable` ?

lib/lz4.c
```diff
@@ -446,9 +447,9 @@ static const U32 LZ4_skipTrigger = 6;  /* Increase this value ==> compression ru
...
 
-typedef enum { noDict = 0, withPrefix64k, usingExtDict } dict_directive;
+typedef enum { noDict = 0, withPrefix64k, usingExtDict, usingExtDictCtx } dict_directive;
```
Cyan4973: Wording of the last 2 directives feel a bit too close, which introduces some confusion.

My understanding is that your patch introduces a new member within compression context, called `cctx->dictCtx`. It's itself a pointer to another compression context. When the new directive is set, all information related to external dictionary must be read from `dictCtx`.

I was expecting that, for familiarity with `zstd` API, you would use something like `CDict`. It's true it's not a separate type though, so maybe `dictCtx` is more appropriate.
In which case, `usingDictCtx` feel like a better name.

Also : try commenting what each directive means / does.

lib/lz4frame.c
Cyan4973: ping

tests/Makefile
```diff
@@ -109,7 +109,7 @@ clean:
         fullbench$(EXT) fullbench32$(EXT) \
         fuzzer$(EXT) fuzzer32$(EXT) \
         frametest$(EXT) frametest32$(EXT) \
-        fasttest$(EXT) datagen$(EXT) checkTag$(EXT)
+        fasttest$(EXT) datagen$(EXT)
```
Cyan4973: add `checkTag`

lib/lz4frame.c
Cyan4973: Also : mixing flag bits into a pointer is a near trick to save memory, which can be handy when the structure size is constrained.
But it doesn't seem to be the case here.
In contrast with LZ4 context type,
`struct LZ4F_cctx_s` is an incomplete type from a public API perspective,
so there is no way to allocate it statically,
so there is no structure size to preserve.

It's likely better for clarity to have dedicated field names.

felixhandte: Ok, yeah, if we don't care about continuity of struct layout, I'll split `lz4CtxLevel` into two fields, one indicating the allocated size and one indicating what it's being used for now.

Cyan4973:
As a general comment, this looks like a fairly good patch.
Technical choices are sound, and code quality is good.

Remaining comments are generally about improving code readability, 
for improved maintenance and reviewer experience.

Also, debugging is not completely done yet,
`valgrind` has apparently found something it did not like : 
https://travis-ci.org/lz4/lz4/jobs/352590664#L1273

lib/lz4.h
```diff
@@ -466,10 +478,7 @@ LZ4_DEPRECATED("use LZ4_decompress_fast() instead") LZ4LIB_API int LZ4_uncompres
 LZ4_DEPRECATED("use LZ4_decompress_safe() instead") LZ4LIB_API int LZ4_uncompress_unknownOutputSize (const char* source, char* dest, int isize, int maxOutputSize);
 
 /* Obsolete streaming functions; use new streaming interface whenever possible */
-LZ4_DEPRECATED("use LZ4_createStream() instead") LZ4LIB_API void* LZ4_create (char* inputBuffer);
```
Cyan4973: Removing support for these functions could, in theory, force us to update the major number of library version. I would prefer not to.

Now, I understand that these functions, and more specifically `LZ4_slideInputBuffer()`, cannot work anymore, due to refurbishing of the pointer required for this action.

Also, this is a fairly complex (and badly specified) API, which never had a lot of clients, had a limited lifetime, was deprecated long ago, so the impact should be fairly limited by now.
Nevertheless, it doesn't take long to find [reference of these functions somewhere over Internet](https://www.google.com/search?q=LZ4_slideInputBuffer). It doesn't prove that they are still active, just that they are not forgotten.

As a consequence, I'm considering a mild intermediate scenario : 
Maybe these prototypes could continue to exist, for linking purpose (imagine that, for obscure reasons, these symbols are still linked within a deeply nested dependency that no-one triggers anyway), but always generate a documented error code if someone tries to use them ?

This is an opened question, feel free to comment.

lib/lz4.h
```diff
@@ -179,13 +179,21 @@ LZ4LIB_API int LZ4_compress_fast (const char* src, char* dst, int srcSize, int d
...
     Then, provide it as 'void* state' to compression function.
+
+    Use the _noReset variant if LZ4_resetStream() was called on the state
+    buffer before being used for the first time (calls to both extState
+    functions leave the state in a safe state, so zeroing is not required
+    between calls). Otherwise, using the legacy _extState requires LZ4 to
+    reinitialize the state internally for every call.
 */
 LZ4LIB_API int LZ4_sizeofState(void);
+LZ4LIB_API int LZ4_compress_fast_extState_noReset (void* state, const char* src, char* dst, int srcSize, int dstCapacity, int acceleration);
```
Cyan4973: This is a new prototype. In theory, we are now bound to support it forever (at least, until the major library version changes).

I was wondering : would it be better to first declare it within a "staging area", so that we get some time to use it and make sure it fulfill its role correctly ? (Something equivalent to `ZSTD_STATIC_LINKING_ONLY` ?)

A recent example where such staging area proved useful is `LZ4F_compressFrame_usingCDict()`, which prototype is going to change now as your experiments proved it would be better to add a `CCtx` as argument.

tests/frametest.c
Cyan4973: if `p==size`, `locateBuffDiff()` will not display any information.
I suspect it would be better to know this unexpected outcome happened.

lib/lz4.h
felixhandte: Yes, I see your point. I don't know if that would be an improvement necessarily.

An alternative is to get clever and pack the context better so we don't have to remove `bufferStart`. We never use the `uint8_t* dictionary` and `LZ4_stream_t_internal* dictCtx` fields at the same time, so they could be a `union` together, or replaced by a single `void*`.

lib/lz4.h
felixhandte: Yes, good call. It needs to be accessible to lz4frame.c, and it would be nice to make it accessible externally, since it does provide better performance. But I agree it should be marked unstable.

lib/lz4.h
felixhandte: Hmm, no, that's not quite true. We only ever use one of them at a time, but there are contexts where we don't have any other information about which mode we're in, `usingExtDict` or `usingDictCtx`. So we would have to get pretty tricky. We could split `uint16_t tableType` into two `uint8_t`s, a `tableType` and `dictDirective`...

What do you think?

Cyan4973: Well, these functions are mostly dead.
They may come back haunting us as zombies, but we should not get out of our way to keep them alive.
Select a simple solution.
Even removing these functions feels better than introducing complex and risky mechanisms (that's still an opened possibility).
I initially liked the "return always error" one on the ground that it felt simple enough to implement, and reduce any potential interaction with the rest of the code.

lib/lz4.h
Cyan4973: We have the [end of `lz4.h`](https://github.com/lz4/lz4/blob/dev/lib/lz4.h#L344) used for such purpose. There is no protection macro defined though, for historical reasons (programs which already use those definitions should not break now). But we could introduce one for newer candidates.

lib/lz4.h
```diff
@@ -465,11 +492,15 @@ LZ4_DEPRECATED("use LZ4_compress_fast_continue() instead") LZ4LIB_API int LZ4_co
...
-LZ4_DEPRECATED("use LZ4_createStream() instead") LZ4LIB_API void* LZ4_create (char* inputBuffer);
-LZ4_DEPRECATED("use LZ4_createStream() instead") LZ4LIB_API int   LZ4_sizeofStreamState(void);
-LZ4_DEPRECATED("use LZ4_resetStream() instead") LZ4LIB_API  int   LZ4_resetStreamState(void* state, char* inputBuffer);
-LZ4_DEPRECATED("use LZ4_saveDict() instead") LZ4LIB_API     char* LZ4_slideInputBuffer (void* state);
+/* Broken, obsolete streaming functions; do not use!
+ *
+ * These functions depended on data that is no longer tracked in the state. They
+ * are therefore broken--they don't retain any history across compressions.
+ */
+LZ4_DEPRECATED("Broken!!! Use LZ4_createStream() instead") LZ4LIB_API void* LZ4_create (char* inputBuffer);
```
Cyan4973: minor : it seems that, we latest changes we have been discussing, these functions are no longer "broken". Compression ratio might be impact a little, but that's a very minor side effect compared to "broken".

felixhandte:
Great!

I also succeeded in getting 32 bit testing working and it all seemed happy.

lib/lz4.c
```diff
@@ -1516,36 +1703,27 @@ They are only provided here for compatibility with older user programs.
...
+    return LZ4_createStream();
 }
 
-char* LZ4_slideInputBuffer (void* LZ4_Data)
+char* LZ4_slideInputBuffer (void* state)
 {
-    LZ4_stream_t_internal* ctx = &((LZ4_stream_t*)LZ4_Data)->internal_donotuse;
-    int dictSize = LZ4_saveDict((LZ4_stream_t*)LZ4_Data, (char*)ctx->bufferStart, 64 KB);
-    return (char*)(ctx->bufferStart + dictSize);
+    // avoid const char * -> char * conversion warning
```
Cyan4973: OK, I missed that one.
Please avoid `//` comment style in `lz4`.
Prefer `/*   */` instead to increase compatibility with older compilers.

felixhandte: Ah, indeed.

Fixed in #487.

PR #407, opened 2017-10-13
Use O2 for the decompression functions on ppc64le with gcc

odaira:
This PR is to propose using the optimization level of O2, instead of O3, for the decompression functions in lz4.c on ppc64le with gcc.

>  Gcc on ppc64le generates an unrolled SIMDized loop for LZ4_wildCopy,
>  together with a simple 8-byte copy loop as a fall-back path.
>  However, this optimization hurts the decompression speed by >30%,
>  because the execution does not go to the optimized loop
>  for typical compressible data, and all of the preamble checks
>  before going to the fall-back path become useless overhead.
>  This optimization happens only with the -O3 flag, and -O2 generates
>  a simple 8-byte copy loop.
>  With gcc on ppc64le, all of the LZ4_decompress_* and LZ4_wildCopy
>  functions are annotated with __attribute__((optimize("O2"))),
>  and also LZ4_wildCopy is forcibly inlined, so that the O2 attribute
>  of LZ4_wildCopy does not affect the compression speed.

I ran the fullbench test for the Silesia and Calgary corpora and observed almost consistent decompression speed-ups throughout the data files on two different ppc64le environments (POWER8 8335-GCA/Ubuntu 17.04/gcc 6.3.0 and POWER8 8247-42L/Ubuntu 16.04.2/gcc 5.4.0). Here are the results of LZ4_decompression_fast:
https://github.com/odaira/lz4/blob/useO2ppc64leFigures/figure_lz4fullbench_LZ4_decompress_fast_P8-8335-GCA.png
https://github.com/odaira/lz4/blob/useO2ppc64leFigures/figure_lz4fullbench_LZ4_decompress_fast_P8-8247-42L.png
I saw similar speed-ups in the other decompression functions like LZ4_decompres_safe, LZ4_decompress_safe_usingDict, etc. The proposed change slows down the decompression of the non-compressible files (silesia/x-ray and calgary/geo) and the highly-compressible file (calgary/pic) by 6% or less, but it pays off for all the other files, which show typical compression ratios of 30%-90%.

I also confirmed that the proposed change does not slow down the compression.
https://github.com/odaira/lz4/blob/useO2ppc64leFigures/figure_lz4fullbench_LZ4_compress_default_P8-8335-GCA.png
https://github.com/odaira/lz4/blob/useO2ppc64leFigures/figure_lz4fullbench_LZ4_compress_default_P8-8247-42L.png

FYI, I have not seen the same effect on x86 Linux.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -117,6 +117,28 @@
 #  endif  /* _MSC_VER */
 #endif /* LZ4_FORCE_INLINE */
 
+/* LZ4_FORCE_O2_GCC_PPC64LE and LZ4_FORCE_O2_INLINE_GCC_PPC64LE
+ * Gcc on ppc64le generates an unrolled SIMDized loop for LZ4_wildCopy,
+ * together with a simple 8-byte copy loop as a fall-back path.
+ * However, this optimization hurts the decompression speed by >30%,
+ * because the execution does not go to the optimized loop
+ * for typical compressible data, and all of the preamble checks
+ * before going to the fall-back path become useless overhead.
+ * This optimization happens only with the -O3 flag, and -O2 generates
+ * a simple 8-byte copy loop.
+ * With gcc on ppc64le, all of the LZ4_decompress_* and LZ4_wildCopy
+ * functions are annotated with __attribute__((optimize("O2"))),
+ * and also LZ4_wildCopy is forcibly inlined, so that the O2 attribute
+ * of LZ4_wildCopy does not affect the compression speed.
+ */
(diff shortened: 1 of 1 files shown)


Cyan4973:
Just waiting for test completion before merge

Cyan4973:
As a side comment : 
I should make a proper investigation to see how much speed is gained from `-O3` compared to `-O2`. It used to matter, but in latest updates, the difference between the 2 has been much reduced, and maybe there is not much difference left.
In which case, defaulting to `-O2` would make sense.

PR #408, opened 2017-10-17
[bench] Use higher resolution timer on POSIX

terrelln:
The timer used was only accurate up to 0.01 seconds. This timer is accurate up to 1 ns.
It is a monotonic timer that measures the real time difference, not on CPU time.

Copied the benchmark code from https://github.com/facebook/zstd/commit/6ab4d5e9041aba962a810ffee191f95897c6208e

CC @felixhandte


Cyan4973:
one issue found : 
https://travis-ci.org/lz4/lz4/jobs/289195047#L725

PR #415, opened 2017-10-31
Faster decoding xp

Cyan4973:
This modification improves decoding speed by 0-10%, depending on source file.
It was successfully tested on x64 and x86 (32-bits).
I haven't been able to test it on other architectures yet.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1146,7 +1146,7 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
                  int partialDecoding,    /* full, partial */
                  int targetOutputSize,   /* only used if partialDecoding==partial */
                  int dict,               /* noDict, withPrefix64k, usingExtDict */
-                 const BYTE* const lowPrefix,  /* == dst when no prefix */
+                 const BYTE* const lowPrefix,  /* always <= dst, == dst when no prefix */
                  const BYTE* const dictStart,  /* only if dict==usingExtDict */
                  const size_t dictSize         /* note : = 0 if noDict */
                  )
@@ -1168,7 +1168,7 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
 
 
     /* Special cases */
-    if ((partialDecoding) && (oexit > oend-MFLIMIT)) oexit = oend-MFLIMIT;                        /* targetOutputSize too high => decode everything */
+    if ((partialDecoding) && (oexit > oend-MFLIMIT)) oexit = oend-MFLIMIT;                      /* targetOutputSize too high => just decode everything */
     if ((endOnInput) && (unlikely(outputSize==0))) return ((srcSize==1) && (*ip==0)) ? 0 : -1;  /* Empty output buffer */
     if ((!endOnInput) && (unlikely(outputSize==0))) return (*ip==0?1:-1);
 
@@ -1178,24 +1178,25 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
         const BYTE* match;
         size_t offset;
 
-        /* get literal length */
         unsigned const token = *ip++;
 
-        if (1)
-        if (ip + 14 + 2 <= iend)
-        if ((token < 15*16) & ((token & 0xF) <= 12)) {
+        /* shortcut for common case */
+        /* this shortcut was tested on x86 and x64, where it improves decoding speed.
+         * it has not yet been benchmarked on ARM, Power, mips, etc. */
+        if (((ip + 14 + 2 <= iend) & (op + 14 + 16 <= oend))
+          & ((token < 15*16) & ((token & 0xF) <= 12)) ) {
             size_t const ll = token >> ML_BITS;
-            size_t const off = LZ4_readLE16(ip+ll);   /* check input validity */
-            if (off >= 16) {
+            size_t const off = LZ4_readLE16(ip+ll);
+            const BYTE* const matchPtr = op + ll - off;  /* pointer underflow risk ? */
+            if ((off >= 16) & (matchPtr >= lowPrefix)) {
                 size_t const ml = (token & 0xF) + MINMATCH;
-                DEBUGLOG(2, "rest:%u, ll:%2u, ml:%2u, off:%u",
-                    (U32)(oend-op), (U32)ll, (U32)ml, (U32)off);
-                memcpy(op, ip, 16); op += ll; ip += ll + 2 /* offset */;
-                memcpy(op, op - off, 16); op += ml;
+                memcpy(op, ip, 16); op += ll; ip += ll + 2 /*offset*/;
+                memcpy(op, matchPtr, 16); op += ml;
                 continue;
             }
         }
 
+        /* decode literal length */
         if ((length=(token>>ML_BITS)) == RUN_MASK) {
             unsigned s;
             do {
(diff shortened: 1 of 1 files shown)


lib/lz4.c
```diff
@@ -1178,8 +1178,25 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
...
         size_t offset;
 
-        /* get literal length */
         unsigned const token = *ip++;
+
+        /* shortcut for common case */
+        /* this shortcut was tested on x86 and x64, where it improves decoding speed.
+         * it has not yet been benchmarked on ARM, Power, mips, etc. */
+        if (((ip + 14 + 2 <= iend) & (op + 14 + 16 <= oend))
+          & ((token < 15*16) & ((token & 0xF) <= 12)) ) {
```
terrelln: Can you use `(token < (15 << 4))`, so it is more clear that we are looking at the top 4 bits.

lib/lz4.c
```diff
@@ -1178,8 +1178,25 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
...
+            size_t const ll = token >> ML_BITS;
+            size_t const off = LZ4_readLE16(ip+ll);
+            const BYTE* const matchPtr = op + ll - off;  /* pointer underflow risk ? */
+            if ((off >= 16) & (matchPtr >= lowPrefix)) {
+                size_t const ml = (token & 0xF) + MINMATCH;
+                memcpy(op, ip, 16); op += ll; ip += ll + 2 /*offset*/;
```
terrelln: How does the speed compare to wildcopy, or just two `memcpy(op, ip, 8);`?

lib/lz4.c
```diff
@@ -1178,8 +1178,25 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
         const BYTE* match;
         size_t offset;
 
-        /* get literal length */
         unsigned const token = *ip++;
+
+        /* shortcut for common case */
```
terrelln: Can you add a comment explaining what the common case is?

lib/lz4.c
terrelln: Was it faster to only copy 16 byte matches, instead of the possible 19?

Cyan4973: 16 bytes is meant to match an SSE register,
or two 64-bits registers.

The maximum possible match length in the token is 18 bytes, since 19 requires reading additional bytes.
I haven't tested speed for 18 bytes.
I expect it to perform worse as it requires an additional register copy all the time, while a match length of 17 or 18 is likely marginal.

lib/lz4.c
Cyan4973: not tested yet.
`memcpy(,,16)` gives a chance to use an SSE register.
I expect it to nicely degrade to two `memcpy(,,8)` when such register is not available.

On systems without 16-bytes registers, `memcpy(,,8)` for literals could prove a better choice, though it highly depends on the distribution of literal lengths values.
If we assume that most literal lengths are `<= 2`, an 8-bytes copy is likely enough, and costs less work for system limited to 64-bits memory transfers.
But if the distribution is a bit wider, literal lengths from `9` to `14` become no longer eligible to shortcut, making the shortcut test less predictable. The impact on the branch predictor might be more problematic than the extra 8-bytes copy.

lib/lz4.c
Cyan4973: I just tested a variant using `matchLength <= 18` on x64.
Performance was similar on my small sample set (`silesia`, `calgary`, `enwik8`), with differences within margin error.

Cyan4973: Tested on godbolt : 
- clang trunk uses SSE registers : 
- gcc 7.2 doesn't, its assembly output looks more complex

https://godbolt.org/g/5cn7t4

Cyan4973: `memcpy(,,18)` introduces a mix of SSE and normal registers : 
https://godbolt.org/g/fiWFgM

lib/lz4.c
Cyan4973: I just tested a variant limiting literal lengths to 8, in order to use a `memcpy(,,8)` for literals.
Its speed is _significantly_ worse, even worse than without the patch.

It seems that making the branch predictor reliable is paramount to performance.

Cyan4973:
Added comments as requested

lib/lz4.c
terrelln: Ahh sorry, I meant `memcpy(op, ip, 8); memcpy(op + 8, ip + 8, 8);` instead of `memcpy(op, ip, 16);`. Looking at the godbolt output, the difference will be using SSE vs not using SSE. I suspect SSE will win, especially since that is what the compiler outputs.

If 16 does win by a good amount, could wildcopy be using 16 bytes? Or would the increased overlap range where we have to special case the copy be a large slowdown.

Cyan4973: moving all `wildcopy()` operations `16` is more complex to implement.
there are a number of tests which depend on `wildcopy()` over-writing to be limited to 8,
and the overlapping match code also depends on it.

I made a test on this scenario a long time ago, the result was so-so, not bad but not terribly good, and seemed not worth added code complexity.
It could also be that the modifications were not done well enough, masking some of the potential benefits.

Cyan4973:
Made some complementary tests : 
compressed file using `lz4 -12`, instead of default `lz4 -1`.

The impact of this patch on decompression speed becomes _negative_,
though by a small amount (0-2%).

Quite a difference ...

Cyan4973:
For data compressed with `lz4 -12`, changing shortcut's maximum match length from 16 to 18 actually improves decoding speed, in the range of 3-4 %. Not huge but definitely beyond error noise.

Since it doesn't negatively impact decoding speed produced at level 1, I'll update the code to use match length 18 bytes instead.

terrelln:
I gathered some stats on silesia

| level | ML | Ratio |
|---|---|---|
| 1 | 16 | .778 |
| 1 | 18 | .781 |
| 12 | 16 | .817 |
| 12 | 18 | .847 |

Cyan4973:
That's a lower % than I would expect.
Anyway, in spite of this relatively poor %, decompression speed is nonetheless improved.
It implies that savings go beyond branch prediction : there is simply much less work to do in the shortcut.

Cyan4973:
I tried a more complex strategy, decoding matchlength >= 19 by reading an extra byte.
The operation is branchless, so decoding matchlength is more complex but predictable.
The idea was to elect larger match lengths in shortcut, such as 24, 32, etc.

The variant failed to produce any benefit.
It was in fact slower.
I suspect that's due to the additional logic to decode match length, which is still in the critical path to determine shortcut.

Cyan4973:
In loosely related topic :  

In order to improve decompression speed, I tried a few different strategies to deal with overlap matches. The intention was to reduce the amount of store forwarding failures due to misalignment.

I tried multiple variants, all based on the same core concept : create a "reference" pattern to copy, so that the `src` ptr doesn't move during the match copy.

Unfortunately, this is no better than current strategy. Differences are small, and in general, performance is close enough, though slightly negative for the best variant (minor exception for `enwik8`, on which I can get a "slightly positive" outcome with best variant).

This seems not worth extra code complexity, especially given that I've not thoroughly checked that all boundaries checks were still correct (they probably need adaptation, potentially mode branches, etc.)

PR #416, opened 2017-11-03
Improve Optimal parser

### LINKED ISSUE #339: High compression time for specific data
Usually LZ4(HC) compresses small data chunks within a few milliseconds, even with LZ4HC_CLEVEL_MAX.
The attached 100kb file takes around 4 seconds though. It's extracted from a 3 MB file which takes 60 seconds. On LZ4HC_CLEVEL_OPT_MIN and below even the 3 MB file just takes a few milliseconds as expected. Is this a fixable edge-case?
I'm using the 1.7.5 64 bit release and LZ4_compress_HC with properly set buffer size.

[dump.txt](https://github.com/lz4/lz4/files/867217/dump.txt)
(Not a text file, but I got an error when trying to upload it zipped and/or with other extensions)

Cyan4973:
The match finder has been changed from binary tree to hash train, requiring a full rewrite of optimal parser. The outcome is a seamless transition between regular HC parser and the new optimal parser.
It saves a bunch of incompatibility issues (`cdict`, `_destSize`), also unlocking ability to change compression level on the fly between blocks.

New optimal parser is faster and compress a little better than v1.8.0.
It also fixes #339.
See https://phabricator.intern.facebook.com/P58578327 for details.

### DIFF
--- a/lib/lz4hc.c
+++ b/lib/lz4hc.c
@@ -661,7 +661,6 @@ static int LZ4HC_compress_generic (
             case 11:
                 return LZ4HC_compress_optimal(ctx, src, dst, *srcSizePtr, dstCapacity, limit, 512, 128, 0);
             default:
-                cLevel = 12;
                 /* fall-through */
             case 12:
                 return LZ4HC_compress_optimal(ctx, src, dst, *srcSizePtr, dstCapacity, limit, 1<<13, LZ4_OPT_NUM, 1);
--- a/lib/lz4hc.c
+++ b/lib/lz4hc.c
@@ -188,12 +188,15 @@ LZ4_FORCE_INLINE int LZ4HC_InsertAndGetWiderMatch (
     repeat_state_e repeat = rep_untested;
     size_t srcPatternLength = 0;
 
-
+    DEBUGLOG(7, "LZ4HC_InsertAndGetWiderMatch");
     /* First Match */
     LZ4HC_Insert(hc4, ip);
     matchIndex = HashTable[LZ4HC_hashPtr(ip)];
+    DEBUGLOG(7, "First match at index %u / %u (lowLimit)",
+                matchIndex, lowLimit);
 
     while ((matchIndex>=lowLimit) && (nbAttempts)) {
+        DEBUGLOG(7, "remaining attempts : %i", nbAttempts);
         nbAttempts--;
         if (matchIndex >= dictLimit) {
             const BYTE* const matchPtr = base + matchIndex;
@@ -360,16 +363,18 @@ LZ4_FORCE_INLINE int LZ4HC_encodeSequence (
 #if defined(LZ4_DEBUG) && (LZ4_DEBUG >= 2)
     static const BYTE* start = NULL;
     static U32 totalCost = 0;
+    U32 const pos = (U32)(*anchor - start);
     U32 const ll = (U32)(*ip - *anchor);
     U32 const llAdd = (ll>=15) ? ((ll-15) / 255) + 1 : 0;
     U32 const mlAdd = (matchLength>=19) ? ((matchLength-19) / 255) + 1 : 0;
     U32 const cost = 1 + llAdd + ll + 2 + mlAdd;
     if (start==NULL) start = *anchor;  /* only works for single segment */
-    totalCost += cost;
-    DEBUGLOG(2, "pos:%7u -- literals:%3u, match:%4i, offset:%5u, cost:%3u/%7u",
-                (U32)(*anchor - start),
+    //g_debuglog_enable = (pos >= 112705) & (pos <= 112760);
+    DEBUGLOG(2, "pos:%7u -- literals:%3u, match:%4i, offset:%5u, cost:%3u / %u",
+                pos,
                 (U32)(*ip - *anchor), matchLength, (U32)(*ip-match),
                 cost, totalCost);
+    totalCost += cost;
 #endif
 
     /* Encode Literal length */
--- a/lib/lz4hc.c
+++ b/lib/lz4hc.c
@@ -183,7 +183,7 @@ LZ4_FORCE_INLINE int LZ4HC_InsertAndGetWiderMatch (
     const BYTE* const dictBase = hc4->dictBase;
     int const delta = (int)(ip-iLowLimit);
     int nbAttempts = maxNbAttempts;
-    reg_t const pattern = LZ4_read_ARCH(ip);
+    U32 const pattern = LZ4_read32(ip);
     U32 matchIndex;
     repeat_state_e repeat = rep_untested;
     size_t srcPatternLength = 0;
@@ -195,10 +195,16 @@ LZ4_FORCE_INLINE int LZ4HC_InsertAndGetWiderMatch (
 
     while ((matchIndex>=lowLimit) && (nbAttempts)) {
         nbAttempts--;
+        int trace = 0;
+        if (nbAttempts==0) {
+            trace = 1;
+            DEBUGLOG(2, "reached max nb of attempts ! : %08X => %08X ",
+                pattern, HASH_FUNCTION(pattern));
+        }
         if (matchIndex >= dictLimit) {
             const BYTE* const matchPtr = base + matchIndex;
             if (*(iLowLimit + longest) == *(matchPtr - delta + longest)) {
-                if (LZ4_read32(matchPtr) == (U32)pattern) {
+                if (LZ4_read32(matchPtr) == pattern) {
                     int mlt = MINMATCH + LZ4_count(ip+MINMATCH, matchPtr+MINMATCH, iHighLimit);
     #if 0
                     /* more generic but unfortunately slower ... */
@@ -221,9 +227,9 @@ LZ4_FORCE_INLINE int LZ4HC_InsertAndGetWiderMatch (
             }
         } else {   /* matchIndex < dictLimit */
             const BYTE* const matchPtr = dictBase + matchIndex;
-            if (LZ4_read32(matchPtr) == (U32)pattern) {
+            if (LZ4_read32(matchPtr) == pattern) {
                 int mlt;
-                int back=0;
+                int back = 0;
                 const BYTE* vLimit = ip + (dictLimit - matchIndex);
                 if (vLimit > iHighLimit) vLimit = iHighLimit;
                 mlt = LZ4_count(ip+MINMATCH, matchPtr+MINMATCH, vLimit) + MINMATCH;
@@ -243,22 +249,25 @@ LZ4_FORCE_INLINE int LZ4HC_InsertAndGetWiderMatch (
         {   U32 const nextOffset = DELTANEXTU16(chainTable, matchIndex);
             matchIndex -= nextOffset;
             if (1 && (nextOffset==1)) {
+                if (trace) DEBUGLOG(2, "check repeat mode : %u", repeat);
                 /* may be a repeated pattern */
                 if (repeat == rep_untested) {
-                    if (LZ4_read32(ip+4) == (U32)pattern) {  /* should check ip limit */
+                    if ((pattern & 0xFFFF) == (pattern >> 16)) {   /* is it enough ? */
                         repeat = rep_confirmed;
-                        srcPatternLength = LZ4HC_countPattern(ip+8, iHighLimit, pattern) + 8;
+                        srcPatternLength = LZ4HC_countPattern(ip+4, iHighLimit, pattern) + 4;
                     } else {
                         repeat = rep_not;
                 }   }
                 if ( (repeat == rep_confirmed)   /* proven repeated pattern (1-2-4) */
                   && (matchIndex >= dictLimit) ) {   /* same segment only */
                     const BYTE* const matchPtr = base + matchIndex;
-                    if (LZ4_read_ARCH(matchPtr) == pattern) {  /* good candidate */
+                    if (trace) DEBUGLOG(2, "search direct pattern position");
+                    if (LZ4_read32(matchPtr) == pattern) {  /* good candidate */
                         size_t const forwardPatternLength = LZ4HC_countPattern(matchPtr+sizeof(pattern), iHighLimit, pattern) + sizeof(pattern);
                         const BYTE* const maxLowPtr = (lowPrefixPtr + MAX_DISTANCE >= ip) ? lowPrefixPtr : ip - MAX_DISTANCE;
-                        size_t const backLength = LZ4HC_reverseCountPattern(matchPtr, maxLowPtr, (U32)pattern);
+                        size_t const backLength = LZ4HC_reverseCountPattern(matchPtr, maxLowPtr, pattern);
                         size_t const currentSegmentLength = backLength + forwardPatternLength;
+                        if (trace) DEBUGLOG(2, "good start position (match == pattern)");
 
                         if ( (currentSegmentLength >= srcPatternLength)   /* current pattern segment large enough to contain full srcPatternLength */
                           && (forwardPatternLength <= srcPatternLength) ) { /* haven't reached this position yet */
@@ -633,7 +642,7 @@ static int LZ4HC_getSearchNum(int compressionLevel)
     switch (compressionLevel) {
         default: return 0; /* unused */
         case 11: return 128;
-        case 12: return 1<<10;
+        case 12: return 1<<13;
     }
 }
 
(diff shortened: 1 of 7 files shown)


lib/lz4hc.c
```diff
@@ -180,59 +180,164 @@ LZ4_FORCE_INLINE int LZ4HC_InsertAndGetWiderMatch (
...
-                if (mlt > longest) { longest = mlt; *matchpos = base + matchIndex + back; *startpos = ip+back; }
+                if (mlt > longest) {
+                    longest = mlt;
+                    *matchpos = base + matchIndex + back;
+                    *startpos = ip + back;
+        }   }   }
+
+        {   U32 const nextOffset = DELTANEXTU16(chainTable, matchIndex);
+            matchIndex -= nextOffset;
+            if (1 && (nextOffset==1)) {
```
terrelln: Remove the `1 && `

lib/lz4hc.c
```diff
@@ -180,59 +180,164 @@ LZ4_FORCE_INLINE int LZ4HC_InsertAndGetWiderMatch (
...
+                /* may be a repeated pattern */
+                if (repeat == rep_untested) {
+                    if ((pattern & 0xFFFF) == (pattern >> 16)) {   /* is it enough ? */
```
terrelln: Does this guarantee that `pattern` is a byte repeated 4 times? It seems like it will work given that `nextOffset == 1`, unless `nextOffset` is a collision. If it is just a collision, I think this only guarantees that it is 2 bytes repeated twice.

lib/lz4hc.c
```diff
@@ -180,59 +180,164 @@ LZ4_FORCE_INLINE int LZ4HC_InsertAndGetWiderMatch (
...
+                        repeat = rep_confirmed;
+                        srcPatternLength = LZ4HC_countPattern(ip+4, iHighLimit, pattern) + 4;
```
terrelln: `pattern` here is a `U32`, which will be zero extended to a `reg_t`. In 64-bit mode I think we need `(U64)pattern << 32 & pattern`. It should probably be handled inside of `LZ4HC_countPattern()`.

lib/lz4hc.c
```diff
@@ -180,59 +180,164 @@ LZ4_FORCE_INLINE int LZ4HC_InsertAndGetWiderMatch (
...
+                    } else {
+                        repeat = rep_not;
+                }   }
+                if ( (repeat == rep_confirmed)   /* proven repeated pattern (1-2-4) */
```
terrelln: The comment is out of date, the check only works for `1`.

lib/lz4hc.c
```diff
@@ -116,54 +117,53 @@ LZ4_FORCE_INLINE void LZ4HC_Insert (LZ4HC_CCtx_internal* hc4, const BYTE* ip)
...
+         && (ip[back-1] == match[back-1]))
+            back--;
+    return back;
+}
 
-LZ4_FORCE_INLINE int LZ4HC_InsertAndFindBestMatch (LZ4HC_CCtx_internal* const hc4,   /* Index table will be updated */
-                                               const BYTE* const ip, const BYTE* const iLimit,
-                                               const BYTE** matchpos,
-                                               const int maxNbAttempts)
+static unsigned LZ4HC_countPattern(const BYTE* ip, const BYTE* const iEnd, reg_t pattern)
```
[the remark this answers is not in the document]
Cyan4973: This interface is wrong.
`pattern` should have `U32` type.

I believe it remained unnoticed because most repetitive patterns tested are a bunch of zero, so `0x00000000` becomes `0x0000000000000000`, hence same pattern.
But that doesn't work with a different byte value.

To be fixed.

Cyan4973:
`LZ4HC_countPattern()` fixed

lib/lz4hc.c
```diff
@@ -180,59 +180,164 @@ LZ4_FORCE_INLINE int LZ4HC_InsertAndGetWiderMatch (
...
         nbAttempts--;
         if (matchIndex >= dictLimit) {
             const BYTE* const matchPtr = base + matchIndex;
             if (*(iLowLimit + longest) == *(matchPtr - delta + longest)) {
-                if (LZ4_read32(matchPtr) == LZ4_read32(ip)) {
+                if (LZ4_read32(matchPtr) == pattern) {
                     int mlt = MINMATCH + LZ4_count(ip+MINMATCH, matchPtr+MINMATCH, iHighLimit);
+    #if 0
+                    /* more generic but unfortunately slower ... */
+                    int const back = LZ4HC_countBack(ip, matchPtr, iLowLimit, lowPrefixPtr);
```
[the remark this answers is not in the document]
Cyan4973: Unfortunately, not in my tests.
Speed is noticeably slower, even with `LZ4_FORCE_INLINE`.
Could it be a compiler version difference ?

lib/lz4hc.c
```diff
@@ -180,59 +180,164 @@ LZ4_FORCE_INLINE int LZ4HC_InsertAndGetWiderMatch (
...
                 int mlt;
-                int back=0;
+                int back = 0;
                 const BYTE* vLimit = ip + (dictLimit - matchIndex);
                 if (vLimit > iHighLimit) vLimit = iHighLimit;
                 mlt = LZ4_count(ip+MINMATCH, matchPtr+MINMATCH, vLimit) + MINMATCH;
                 if ((ip+mlt == vLimit) && (vLimit < iHighLimit))
                     mlt += LZ4_count(ip+mlt, base+dictLimit, iHighLimit);
-                while ((ip+back > iLowLimit) && (matchIndex+back > lowLimit) && (ip[back-1] == matchPtr[back-1])) back--;
+                while ( (ip+back > iLowLimit)
```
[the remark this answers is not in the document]
Cyan4973: If we can make it produce same performance with `clang` compiler, then yes.

lib/lz4hc.c
Cyan4973: Right, this is a left over when testing the feature (easy disabling by turning this value to `0`)

lib/lz4hc.c
Cyan4973: Yes, good point, it can be a collision

lib/lz4hc.c
```diff
@@ -180,59 +180,164 @@ LZ4_FORCE_INLINE int LZ4HC_InsertAndGetWiderMatch (
...
+                                const int maxNbAttempts)
+{
+    const BYTE* uselessPtr = ip;
+    return LZ4HC_InsertAndGetWiderMatch(hc4, ip, ip, iLimit, MINMATCH-1, matchpos, &uselessPtr, maxNbAttempts);
+}
+
+#else
+
+LZ4_FORCE_INLINE
+int LZ4HC_InsertAndFindBestMatch (LZ4HC_CCtx_internal* const hc4,   /* Index table will be updated */
```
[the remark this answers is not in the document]
Cyan4973: No longer.
This was useful to compare performance of older version, which is also simpler to read, with newer one.
According to benchmark measurements, performance is preserved.

lib/lz4hc.c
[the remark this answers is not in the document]
terrelln: Thats a bummer. I was testing with gcc-6.

lib/lz4hc.c
```diff
@@ -161,17 +161,21 @@ static unsigned LZ4HC_countPattern(const BYTE* ip, const BYTE* const iEnd, U32 c
...
 
-    while (likely(ip>=iLow+4)) {
+    while (likely(ip >= iLow+4)) {
         if (LZ4_read32(ip-4) != pattern) break;
         ip -= 4;
     }
     while (likely(ip>iLow)) {
-        if (ip[-1] != (BYTE)pattern) break;
-        ip--;
+        const BYTE* bytePtr = (const BYTE*)(&pattern) + 3; /* works for any endianess */
```
terrelln: This needs to be outside the loop, right? Can the same trick be used in `LZ4HC_countPattern()`?

lib/lz4hc.c
```diff
@@ -180,71 +199,121 @@ LZ4_FORCE_INLINE int LZ4HC_InsertAndGetWiderMatch (
...
     return longest;
 }
 
+LZ4_FORCE_INLINE
+int LZ4HC_InsertAndFindBestMatch(LZ4HC_CCtx_internal* const hc4,   /* Index table will be updated */
+                                const BYTE* const ip, const BYTE* const iLimit,
+                                const BYTE** matchpos,
+                                const int maxNbAttempts)
+{
+    const BYTE* uselessPtr = ip;
```
terrelln: `LZ4HC_InsertAndFindBestMatch()` didn't used to move `startpos` back. If it does get updated, I think we will end up with corrupted data.

lib/lz4hc.c
```diff
@@ -353,7 +434,7 @@ static int LZ4HC_compress_hashChain (
     /* Main Loop */
     while (ip < mflimit) {
         ml = LZ4HC_InsertAndFindBestMatch (ctx, ip, matchlimit, (&ref), maxNbAttempts);
```
terrelln: I think we need to pass `start0` to this call to be updated.

lib/lz4hc.c
Cyan4973: Indeed ! It's a mistake.
It works because all bytes are the same, but would break for 2-bytes or 4-bytes patterns.

Cyan4973: I tried the same ending routine in `LZ4HC_countPattern()`, but speed was decreased by a measurable amount.

lib/lz4hc.c
Cyan4973: `startpos` should not be updated, because the call to `LZ4HC_InsertAndGetWiderMatch()` defines `iLowLimit==ip`, so it cannot look back further.

I can ensure it by an `assert()`.

lib/lz4hc.c
Cyan4973: Not sure to follow this one.
`start0` is updated _after_ a successful call to `LZ4HC_InsertAndFindBestMatch()`, in which case it is set to `ip`.
It does not seem useful to update `start0` before, as the function is just looping over `LZ4HC_InsertAndFindBestMatch()` until it finds at least one match.

lib/lz4hc.c
terrelln: Ok I see now. Can you add a comment explaining that?

lib/lz4hc.c
terrelln: I was incorrect because `LZ4HC_InsertAndFindBestMatch()` won't update `startpos`.

lib/lz4hc.c
Cyan4973: Sure, code comments added

lib/lz4opt.h
```diff
@@ -50,317 +44,289 @@ typedef struct {
...
+                opt[mlen].mlen = mlen;
+                opt[mlen].off = offset;
+                opt[mlen].litlen = llen;
+                opt[mlen].price = cost;
+                DEBUGLOG(7, "rPos:%3i => price:%3i (matchlen=%i) -- initial setup",
+                            mlen, cost, mlen);
+        }   }
+        last_match_pos = firstMatch.len;
+        {   int addLit;
+            for (addLit = 1; addLit <= 3; addLit ++) {
```
terrelln: Should this be `< MINMATCH`, or is it for a different reason?

lib/lz4opt.h
```diff
@@ -50,317 +44,289 @@ typedef struct {
...
-                }
-
-                if (price < (size_t)opt[cur].price)
-                    SET_PRICE(cur, 1 /*mlen*/, 0 /*off*/, litlen, price);   /* note : increases last_pos */
+            LZ4HC_match_t newMatch;
+
+            if (curPtr >= mflimit) break;
+            DEBUGLOG(7, "rPos:%u[%u] vs [%u]%u",
+                    cur, opt[cur].price, opt[cur+1].price, cur+1);
+            if (fullUpdate) {
```
terrelln: Can you add a comment explaining how these heuristics were constructed?

lib/lz4opt.h
```diff
@@ -50,317 +44,289 @@ typedef struct {
...
-        /* encode all recorded sequences */
-        cur = 0;
-        while (cur < last_pos) {
-            int const ml = opt[cur].mlen;
-            int const offset = opt[cur].off;
-            if (ml == 1) { ip++; cur++; continue; }
-            cur += ml;
-            if ( LZ4HC_encodeSequence(&ip, &op, &anchor, ml, ip - offset, limit, oend) ) return 0;
-        }
+                    if (pos > last_match_pos+3 || price <= opt[pos].price) {
```
terrelln: Where does the `3` come from?

lib/lz4opt.h
```diff
@@ -50,317 +44,289 @@ typedef struct {
...
+                          && (last_match_pos < pos) )
+                            last_match_pos = pos;
+                        opt[pos].mlen = ml;
+                        opt[pos].off = offset;
+                        opt[pos].litlen = ll;
+                        opt[pos].price = price;
+            }   }   }
+            /* complete following positions with literals */
+            {   int addLit;
+                for (addLit = 1; addLit <= 3; addLit ++) {
```
terrelln: Is this related to `MINMATCH`? (as above)

lib/lz4opt.h
```diff
@@ -50,317 +44,289 @@ typedef struct {
...
+typedef struct {
+    int off;
+    int len;
+} LZ4HC_match_t;
 
-/** Tree updater, providing best match */
-LZ4_FORCE_INLINE int LZ4HC_BinTree_GetAllMatches (
-                        LZ4HC_CCtx_internal* ctx,
+LZ4_FORCE_INLINE
+LZ4HC_match_t LZ4HC_HashChain_GetAllMatches (LZ4HC_CCtx_internal* const ctx,
```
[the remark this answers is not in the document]
Cyan4973: Logic modified to call `LZ4HC_findLongerMatch()` directly

lib/lz4opt.h
Cyan4973: It's for a different reason.

The purpose of this code is to fill the cost for the next 3 positions after current best match.
Since we have no more match that far, these positions are filled with literals.

So why 3 ?
Well, this part is a bit more complex to explain.

Let's assume there is an initial match of length 17 (cost 4).
Next 3 positions 18, 19, 20 are filled with literals (respective cost 5, 6 and 7).
Let's now assume another match is found, at position 14 (cost 3).
At position 18, this match is not interesting, because total cost is 3+3=6, which is > cost(18)==5.
It starts to be interesting at position 19, as here cost(19)==6, so it's equal, and in this case, it's always beneficial to use the match (from a compression ratio perspective, decoding speed would prefer the literals).  That also means this second match must have a length >= 5 to be interesting. If its length is only 4, it's not an interesting match.

If the next 3 positions were not pre-filled with literal cost, all I would know is that the new match can reach position 18 > 17. In absence of competition, should it be selected blindly for position 18, it would be a bad decision, leading to worse cost.
Now, it's also possible to "calculate the cost of previous match + literal" on the fly, but for simplicity reason, I found it easier to simply pre-fill these 3 slots (and keep the `last_match_pos` to the real end position of the last useful match).

Pre-filling beyond 3 slots is not useful, because at such distance, a new match is necessarily better than 4+ literals.

So that's the reason for number `3`, and it is present in several places.
Maybe I could use a constant instead, in order to name it properly.

Cyan4973: Added constant `TRAILING_LITERALS`

lib/lz4hc.c
```diff
@@ -180,71 +198,124 @@ LZ4_FORCE_INLINE int LZ4HC_InsertAndGetWiderMatch (
...
     const BYTE* const lowPrefixPtr = base + dictLimit;
-    const U32 lowLimit = (hc4->lowLimit + 64 KB > (U32)(ip-base)) ? hc4->lowLimit : (U32)(ip - base) - (64 KB - 1);
+    const U32 lowLimit = (hc4->lowLimit + 64 KB > (U32)(ip-base)) ? hc4->lowLimit : (U32)(ip - base) - MAX_DISTANCE;
     const BYTE* const dictBase = hc4->dictBase;
     int const delta = (int)(ip-iLowLimit);
     int nbAttempts = maxNbAttempts;
+    U32 const pattern = LZ4_read32(ip);
     U32 matchIndex;
+    repeat_state_e repeat = rep_untested;
+    size_t srcPatternLength = 0;
```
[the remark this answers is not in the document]
Cyan4973: `srcPatternLength` and `repeat` are only set once, and then re-used on next searches. 
Their values must survive loop scope.

PR #417, opened 2017-11-05
When building with a C++ compiler, remove the 'register' keyword to silent a warning

sylvestre:
For example, with clang:

```
lz4.c:XXX:36: error: 'register' storage class specifier is deprecated and incompatible with C++17 [-Werror,-Wdeprecated-register]
static unsigned LZ4_NbCommonBytes (register reg_t val)
                                   ^~~~~~~~~
```

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -85,6 +85,17 @@
 #endif
 
 
+/*
+ * register is ignored when the code built with a C++ compiler
+ * Remove the keyword when built with C++ to silent the warning
+ */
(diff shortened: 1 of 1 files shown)


Cyan4973:
Good point

If the problem is `C++17`, could the patch be more accurate in the filtering out ?
Right now, it removes `register` for _all_ `C++` versions.

sylvestre:
Sure, I will update my PR!

Cyan4973:
Thanks for the patch !

PR #419, opened 2017-11-07
-Wdeprecated-register: Silent the warning when building with a non C++17 compiler

sylvestre:
Now, fixing the warning!

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -88,9 +88,17 @@
 /*
  * register is ignored when the code built with a C++-17 compiler
  * Remove the keyword when built with C++-17 to silent the warning
+ * Ignore the warning with other C++ versions
  */
-#if defined(__cplusplus) &&  __cplusplus > 201402L
+#ifdef __cplusplus
+#if __cplusplus > 201402L
+/* C++ 17 */
 #  define REGISTER
+#else
+/* Before C++ 17 */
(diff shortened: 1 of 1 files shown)


lib/lz4.c
```diff
@@ -88,10 +88,18 @@
 /*
  * register is ignored when the code built with a C++-17 compiler
  * Remove the keyword when built with C++-17 to silent the warning
+ * Ignore the warning with other C++ versions
  */
-#if defined(__cplusplus) &&  __cplusplus > 201402L
+#ifdef __cplusplus
+#if __cplusplus > 201402L
```
Cyan4973: minor: try to indent inner `#if`

sylvestre:
I saw the CI failure. 
Another solution would be to disable the register keyword for all C++ . I am not sure it is useful with C++:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/n4340.html
http://en.cppreference.com/w/cpp/keyword/register

Cyan4973:
That's probably true.
To be complete, this hypothesis should be checked (benchmarked) against a variety of compilers to ensure it's not detrimental for some targets.

Cyan4973:
OK, let's remove `register` symbol,
it's likely the simplest thing to do at this stage.

PR #426, opened 2017-12-04
Fix cli sometimes calling UTIL_setFileStat on /dev/null

aliceatlas:
LZ4IO_compressFilename_extRess does not check if dstFileName is /dev/null (nulmark) before calling UTIL_setFileStat on it, which can result in /dev/null being chmodded 644, 600, etc when it's the output file and lz4 is running as root. Fixed that here


Cyan4973:
Thanks for the patch @aliceatlas

PR #427, opened 2017-12-05
remove `register` keyword

Cyan4973:
deprecated starting C++17

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -85,16 +85,6 @@
 #endif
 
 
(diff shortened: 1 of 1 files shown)


Godzil:
But not in C which is the language LZ4 is using.

Cyan4973:
Right @Godzil.
It's just that this keyword has limited usage and impacted, to say the least, in `lz4` code base.
So removing it look like the simpler choice nowadays.

PR #430, opened 2017-12-20
Add failing emscripten test

jscheid:
Tests compression and decompression with buffers at very low memory
addresses, triggering issues to do with unsigned pointer arithmetic.

See #397

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ bin/
 # Mac
 .DS_Store
 *.dSYM
+
+# Vagrant
+.vagrant/
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,100 @@
+Vagrant.configure('2') do |config|
+  config.vm.box = 'velocity42/xenial64'
+  config.vm.box_version = '1.1.0'
+  config.vm.provision 'shell', privileged: false, inline: <<SCRIPT
+
+  set -e
+  set -x
+  set -o pipefail
+
+  # Install apt source for yarn
+  curl -sSL https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
+  echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
+
+  # Install apt source for node
+  curl -sSL https://deb.nodesource.com/setup_8.x | sudo bash
+
+  # Install node and yarn, and other dependencies
+  # See https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md
+  sudo apt-get -y install \
+    nodejs \
+    yarn \
+    gconf-service \
+    libasound2 \
+    libatk1.0-0 \
+    libc6 \
+    libcairo2 \
+    libcups2 \
+    libdbus-1-3 \
+    libexpat1 \
+    libfontconfig1 \
+    libgcc1 \
+    libgconf-2-4 \
+    libgdk-pixbuf2.0-0 \
+    libglib2.0-0 \
+    libgtk-3-0 \
+    libnspr4 \
+    libpango-1.0-0 \
+    libpangocairo-1.0-0 \
+    libstdc++6 \
+    libx11-6 \
+    libx11-xcb1 \
+    libxcb1 \
+    libxcomposite1 \
+    libxcursor1 \
+    libxdamage1 \
+    libxext6 \
+    libxfixes3 \
+    libxi6 \
+    libxrandr2 \
+    libxrender1 \
+    libxss1 \
+    libxtst6 \
+    ca-certificates \
+    fonts-liberation \
+    libappindicator1 \
+    libnss3 \
+    lsb-release \
+    xdg-utils \
+    wget
+
+  # Install puppeteer
+  sudo yarn global add puppeteer@0.13
+
+  # Not clear why this is needed, perhaps a bug in yarn?
+  sudo chmod a+r /usr/local/share/.config/yarn/global/node_modules/puppeteer/.local-chromium/linux-515411/chrome-linux/*
+  sudo chmod a+x /usr/local/share/.config/yarn/global/node_modules/puppeteer/.local-chromium/linux-515411/chrome-linux/chrome
+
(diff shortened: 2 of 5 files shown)


jscheid:
This can be run as follows (needs emscripten environment, tested with 1.37.xx):

```
make -C tests test-emscripten
```

To run in an alternate browser, set the `EMRUN_FLAGS` env var, e.g.:

```
make -C tests test-emscripten EMRUN_FLAGS=--browser=chrome
make -C tests test-emscripten EMRUN_FLAGS=--browser=firefox
```

See also http://kripken.github.io/emscripten-site/docs/compiling/Running-html-files-with-emrun.html

Cyan4973:
I've been trying to test the feature branch, unfortunately without success so far.

- On Mac OS-X, `emcc` complains about an llvm version mismatch, and stops
- On Ubuntu LTS : compilation seems to work, but running test doesn't (`emrun` is missing). Note that emscripten version available on Ubuntu LTS is v1.22.1.

Being able to run the test on Ubuntu LTS would be great, as many CI environments ship this version (or a close derivative).

jscheid:
emscripten config is a bit finicky unfortunately. Are you using emscripten with Homebrew, are you at the latest version (1.37.25?) I remember that the last emscripten upgrade printed a message that told me to change something in `~/.emscripten`, did you see that? FWIW it works here on Mac with Homebrew and 1.37.25.

I'll set up Ubuntu LTS in VMware and see if I can get it to run there. You're talking about 16.04, correct?

Cyan4973:
> Are you using emscripten with Homebrew

yes

> are you at the latest version (1.37.25?)

seems so

> I remember that the last emscripten upgrade printed a message that told me to change something in ~/.emscripten, did you see that?

I saw it on Ubuntu, but not on Mac

> I'll set up Ubuntu LTS in VMware and see if I can get it to run there. You're talking about 16.04, correct?

correct

jscheid:
@Cyan4973 I've pushed a Vagrantfile that you can use for running the test. It's tested to work with Vagrant 2.0.1 and VMware 8.5.9, although slightly older versions (and Virtualbox) should also be fine. You can use it like so:

```shell
vagrant up # might take a while...
vagrant ssh -c 'cd /vagrant && make clean && make && make -C tests test-emscripten'
```

Most of the Vagrantfile is concerned with running on Chrome Headless. I don't think that will be needed on TravisCI because they have an X11 server running and so you can just use Firefox or something.

The only part that I think would be needed is installing emsdk portable:

```shell
  curl -sSL https://s3.amazonaws.com/mozilla-games/emscripten/releases/emsdk-portable.tar.gz \
      | tar xz
  (cd emsdk-portable && \
    ./emsdk install emscripten-1.37.22 && \
    ./emsdk install clang-e1.37.22-64bit && \
    ./emsdk activate emscripten-1.37.22 && \
    ./emsdk activate clang-e1.37.22-64bit)
```

and making sure that `~/emsdk-portable/emsdk_env.sh` is sourced prior to running the tests.

You're right that the default emscripten version in 16.04 (1.22) doesn't include `emrun` and, while theoretically we could run the scripts without it, that would be quite a bit of extra work.

Let me know how you go with this.

Cyan4973:
I tested vagrant on the Linux Ubuntu LTS 16.04 VM.
It required installing vagrant and virtualBox, and enabling VT-x.
It produces the following result : 
```
vagrant up
(...)
The SSH connection was unexpectedly closed by the remote end. This
usually indicates that SSH within the guest machine was unable to
properly start up. Please boot the VM in GUI mode to check whether
it is booting properly.
```

Since I don't control the guest machine (virtualbox was installed verbatim, no VM was created, so I suspect it's provided through vagrant), I'm unsure what's going on and what's next.

jscheid:
Vagrant is intended to be run on your host, not guest. It's really just a way to automate starting up and provisioning a VM. You could also run the commands in the provisioning script manually on a Ubuntu VM.

Cyan4973:
I'm currently using a Max OS-X laptop.
VMware is installed, it's used to run an Ubuntu LTS VM.

My understanding is as follows : 
- vagrant must be run from Ubuntu.
- vagrant needs a "provider". Therefore, inside the Ubuntu VM, virtualbox is also installed.
- From a vagrant perspective, Ubuntu is the host, and virtualbox is the "provider" through which the guest is created and launched.
- I suspect the guest system is defined within `Vagrantfile`

jscheid:
Sorry for any confusion. I made the Vagrantfile so that we have a reproducible way to set up Ubuntu for testing this. Theoretically you _could_ run Vagrant inside a guest, thereby having a VM inside a VM, but that was not the intention. I suggest you either install Vagrant on your Mac laptop and use it to create a new VM for testing, or -- if for some reason you don't want to create a new VM -- copy the shell commands from the provisioning script and run them inside your existing Ubuntu guest. Creating a new VM with Vagrant would be better because it's a clean slate and I've tested it to work, on your existing VM there is a chance that something doesn't work with whatever packages you've already installed on it or any configuration changes you've already made. I hope this makes more sense now?

Cyan4973:
Clear enough.
I tried installing vagrant on Mac through a hopeful `brew install vagrant`, unsuccessfully.
So I guess installation is going to be a bit more involved.

Cyan4973:
I could install `vagrant` for OS-X, using the [official installer](https://www.vagrantup.com/downloads.html).
then start `vagrant` from console : 
```
vagrant up
No usable default provider could be found for your system.
```
Strange thing is, VMware is installed on this laptop.

jscheid:
Oh, I forgot. if you’re using VMware you need to install a separate Vagrant provider which is not free. Sorry, I forgot about that, it’s been such a long time since I’ve set it all up here.

In that case, since you presumably don’t want to buy the vagrant provider for VMware ($79) you could either use Virtualbox or run the script manually after tweaking the paths.

Cyan4973:
In branch __[lowAddr](https://github.com/lz4/lz4/tree/lowAddr)__ , `fuzzer` tests have been enhanced with a method to generate buffer at low address space.
This covers the same need as #430, and makes it possible to further analyse #397.
So I'll close this PR.

PR #432, opened 2017-12-21
Fix bug with buffers at very low memory addresses

jscheid:
See #397 

See also #430 for corresponding test cases.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -603,7 +603,7 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
 
                 match = LZ4_getPositionOnHash(h, cctx->hashTable, tableType, base);
                 if (dict==usingExtDict) {
-                    if (match < (const BYTE*)source) {
+                    if ((ptrdiff_t)match < (ptrdiff_t)source) {
                         refDelta = dictDelta;
                         lowLimit = dictionary;
                     } else {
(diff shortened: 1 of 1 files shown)


Cyan4973:
Sorry, I did not reviewed this patch before, as I wanted to reproduce #430 first,
and then start an analysis from there. But as I could not reproduce #430 ...

Anyway, verdict is, this patch is not merged because I suspect it introduces some new problems (which your use case doesn't trigger).

The most important idea is that you are replacing several pointer comparisons with their corresponding signed value comparison, through a cast to `(ptrdiff_t)`.
I'm sure it solves your issue, when data is loaded into low memory region. But it introduces some new ones.

First, not all platform have `sizeof(ptrdiff_t) == sizeof(void*)`. This is a reasonable expectation for many systems, but exceptions exist, such as openVMS64 (where `sizeof(void*)==8`, but `sizeof(ptrdiff_t)==sizeof(size_t)==4`. See https://groups.google.com/forum/#!msg/comp.os.vms/UuK6cEBjEgQ/F9_1SLDocPoJ for example.
So it becomes a portability issue, because on those platforms, the cast will truncate upper bits, resulting in wrong comparison.

A more important issue is that `ptrdiff_t` indeed avoids overflow issues at address `0x00000000`, but creates a new one at address `0x80000000`. 
For `(ptrdiff_t)match < (ptrdiff_t)source)`, one could be unlucky, and have `match==0x7FFFFFFE` and `source==0x80000001`. In which case, `match` is indeed `< source`.  But, the cast to `ptrdiff_t` will make `match` look positive and `source` negative, so the test would be reversed.

To be clear, I believe the problem described in #397 is real and need a solution.
But it will have to be another one.

I still believe a correct first step would be to create a use case which triggers this problem, so I will spend some more time on this action.

Issue #441, opened 2018-01-04
lz4 -d --rm does not preserve timestamp


expressoCoder:
Using lz4 v1.8.0 on CentOS 6.8 x86_64

Compressing with `lz4` or `lz4 --rm` will preserve timestamp
Decompressing with `lz4 -d` will also preserve timestamp.

However `lz4 -d --rm` will not preserve timestamp

terrelln:
I can reproduce the issue, and will look into it shortly.

Cyan4973:
fixed by @terrelln

PR #442, opened 2018-01-05
[lz4io] Fix decompression file stat with --rm

terrelln:
Fixes Issue #441.


johan-boule (commented):
utime only has a resolution of 1 second, so this doesn't work.

PR #448, opened 2018-01-13
Fix lz4 version

sunpoet:


Cyan4973:
Thanks for the patch @sunpoet !
Indeed, I managed to miss this one ...

PR #454, opened 2018-01-17
Change file format back to ASCII (from UTF-8)

sunpoet:
- Replace U+00A0 by space
- Fix build failure of archivers/py-borgbackup in FreeBSD

Reference: https://bugs.FreeBSD.org/225235


Cyan4973:
Wow, pretty much invisible, I wonder how it managed to get there.
Thanks for the patch @sunpoet !

Issue #445, opened 2018-01-09
what is the frame footer?


psteinb:
Hi,
I am just trying to follow the example on frame based compression, the example defines a preprocessor symbol by the name of `LZ4_FOOTER_SIZE`. 

https://github.com/lz4/lz4/blob/0b203b04f6b5085f9f98c8c78c6535399264d755/examples/frameCompress.c#L13

The `lz4frame.h` doesn't define anything alike. the documentation for the frame format doesn't cite a footer at all.

So I am wondering, what this is?

On a related note, the [frame format docs](https://github.com/lz4/lz4/blob/dev/doc/lz4_Frame_format.md) do not define what the header is as well. Only the header checksum is defined.

psteinb:
Cool, thanks a bunch. I'd fancy to submit a PR so that is documented in 
the example as well as in the design docs for the frame API.

I'd love to use lz4 in a multi-threaded fashion. In that context, is the 
frame API thread-safe? For this I am considering the following 2 cases:

a) call the following sequence
LZ4F_createCompressionContext+LZ4F_compressBegin+LZ4F_compressUpdate+LZ4F_compressEnd+LZ4F_freeCompressionContext
in each thread separately and also create the relevant structs once per 
thread?

b) call LZ4F_createCompressionContext once, then start the parallel 
region and call the sequence 
LZ4F_compressBegin+LZ4F_compressUpdate+LZ4F_compressEnd per thread, 
after ending the parallel processing, I'd call LZ4F_freeCompressionContext.

thanks for your help so far.
P

Cyan4973:
- You will need one context per thread
- contexts can be re-used by successive compression jobs

Which gives : 

For each thread : 
```
LZ4F_createCompressionContext()
(.. do some works ...)
LZ4F_freeCompressionContext
```

For each work : 
```
LZ4F_compressBegin()
LZ4F_compressUpdate()  // as many times as necessary
LZ4F_compressEnd()
```

psteinb:
thanks for all your answers and for picking it up in #458. Question is how to deal with a global macro specifying this as [frameCompess.c](https://github.com/lz4/lz4/blob/0b203b04f6b5085f9f98c8c78c6535399264d755/examples/frameCompress.c#L13) does?

Cyan4973:
[...]

PR #457, opened 2018-01-30
inserted information from github issue #445

psteinb:


Cyan4973:
Thanks for the proposed update.
I understand the need from #445, but these new terms must be introduced with care into existing specification. I'll give it a try, and see what I can do.

psteinb:
fine with me, as long as they are well documented

PR #474, opened 2018-02-19
It seems more cache-friendly for offset=1 and offset=2

hsiangkao:
Not quite sure but it seems more aligned and cache-friendly

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1185,7 +1185,7 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
     BYTE* oexit = op + targetOutputSize;
 
     const BYTE* const dictEnd = (const BYTE*)dictStart + dictSize;
-    const unsigned inc32table[8] = {0, 1, 2,  1,  0,  4, 4, 4};
+    const unsigned inc32table[8] = {0, 0, 0,  1,  0,  4, 4, 4};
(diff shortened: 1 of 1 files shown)


Cyan4973:
I believe the existing setting is actually better.

Here is why : 
on starting the overlapping sequence, "origin" is 1 byte before "destination" (for `offset==1`).
After the initial byte-by-byte copy, "origin" hasn't moved, while "destination" is 4 bytes farther.
So now distance between "origin" and "destination" is 5 bytes.
By adding 1 byte to "origin", the distance is shrunk to 4 bytes, 
which is just right for the next 4-bytes copy.
Advanced CPU will recognise this pattern, as there is no aliasing, so data from "origin" will be fetched directly from pipeline cache, without a need to actually commit data to memory.

Same logic when initial distance is 2 bytes.

In any case, if you believe you found an improvement to the algorithm, which is perfectly possible, please ensure your statement with benchmark measurements, and publish them. There are many examples where a modification that "looks like" an improvement ends up being actually detrimental or neutral. Optimization can be hard.

hsiangkao:
Hi, thanks for your kindly detailed explanation.
I need to look up more texts and do some experiments to confirm that, which is some different from my understanding of modern CPU architecture through.

Thanks.

PR #475, opened 2018-02-20
Better bench measurements for small inputs

Cyan4973:
`time()` invocation is not free,
and when a unit of work is very small (like compressing / decompressing a small file)
`time()` invocation can actually represent a sizable fraction of workload measured.

The new strategy is to use `time()` only once per batch,
and dynamically resize batch size so that each round lasts approximately 1 second,
starting from a conservative (low) estimate.

This change only matters for small inputs.
Measurement for large files (such as silesia.tar) are much less impacted.
(though decoding speed is so fast that even medium-size files will notice an improvement).

### DIFF
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -221,8 +221,8 @@ static int BMK_benchMem(const void* srcBuffer, size_t srcSize,
         U64 const crcOrig = XXH64(srcBuffer, srcSize, 0);
         UTIL_time_t coolTime;
         U64 const maxTime = (g_nbSeconds * TIMELOOP_NANOSEC) + 100;
-        U32 nbCompressionLoops = ((5 MB) / (srcSize+1)) + 1;  /* conservative initial compression speed estimate */
-        U32 nbDecodeLoops = ((200 MB) / (srcSize+1)) + 1;  /* conservative initial decode speed estimate */
+        U32 nbCompressionLoops = (U32)((5 MB) / (srcSize+1)) + 1;  /* conservative initial compression speed estimate */
+        U32 nbDecodeLoops = (U32)((200 MB) / (srcSize+1)) + 1;  /* conservative initial decode speed estimate */
         U64 totalCTime=0, totalDTime=0;
         U32 cCompleted=0, dCompleted=0;
 #       define NB_MARKS 4
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -67,6 +67,7 @@ static int LZ4_compress_local(const char* src, char* dst, int srcSize, int dstSi
 
 #define NBSECONDS             3
 #define TIMELOOP_MICROSEC     1*1000000ULL /* 1 second */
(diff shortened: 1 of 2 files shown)


programs/bench.c
```diff
@@ -41,6 +41,7 @@
 #include <string.h>      /* memset */
 #include <stdio.h>       /* fprintf, fopen, ftello */
 #include <time.h>        /* clock_t, clock, CLOCKS_PER_SEC */
+#include <assert.h>      /* assert */
```
terrelln: Does lz4 have a `LZ4_DEBUG=X` parameter, or does it set `-DNDEBUG` for optimized builds?

programs/bench.c
```diff
@@ -272,44 +279,48 @@ static int BMK_benchMem(const void* srcBuffer, size_t srcSize,
...
                         }
-
                         blockTable[blockNb].resSize = regenSize;
+                }   }
+                {   U64 const clockSpan = UTIL_clockSpanNano(clockStart);
+                    if (clockSpan > 0) {
+                        if (clockSpan < fastestD * nbDecodeLoops)
+                            fastestD = clockSpan / nbDecodeLoops;
+                        assert(fastestD > 0);
+                        nbDecodeLoops = (U32)(1000000000/*1sec*/ / fastestD) + 1;  /* aim for ~1sec */
```
terrelln: Can you use `TIMELOOP_NANOSEC` here?

programs/bench.c
```diff
@@ -246,21 +247,27 @@ static int BMK_benchMem(const void* srcBuffer, size_t srcSize,
...
                         size_t const rSize = compP.compressionFunction(blockTable[blockNb].srcPtr, blockTable[blockNb].cPtr, (int)blockTable[blockNb].srcSize, (int)blockTable[blockNb].cRoom, cLevel);
                         if (LZ4_isError(rSize)) EXM_THROW(1, "LZ4_compress() failed");
                         blockTable[blockNb].cSize = rSize;
+                }   }
+                {   U64 const clockSpan = UTIL_clockSpanNano(clockStart);
+                    if (clockSpan > 0) {
+                        if (clockSpan < fastestC * nbCompressionLoops)
+                            fastestC = clockSpan / nbCompressionLoops;
+                        assert(fastestC > 0);
+                        nbCompressionLoops = (U32)(1000000000/*1sec*/ / fastestC) + 1;  /* aim for ~1sec */
```
terrelln: Can you use `TIMELOOP_NANOSEC` here, or introdude a new macro like `SECOND_IN_NANOSEC`?

programs/bench.c
Cyan4973: lz4 has a `LZ4_DEBUG=X` parameter, but that's only for `/lib`.

Modules in `/programs` just enable `assert()` all the time.
That's because they are not performance sensitive.
Even in `bench.c`, `assert()` is used in few places and outside of timing loops.

PR #486, opened 2018-03-21
Add Dependency to Fix Parallel `make test` Runs

felixhandte:
When run with `-jN`, the `rm tmp*` can run in the middle of the `test-lz4-dict`
job, which will then fail, finding its files to have been axed. This adds a
dependency between the two.


Cyan4973:
Nice catch @felixhandte !

PR #488, opened 2018-03-26
Use Dictionary In-Place in HC Mode

felixhandte:
I'll include a nicer presentation of the performance implications of these changes in the release notes, but in summary:

- Compression which safely reuses a context (via `LZ4_resetStreamHC_fast()`, which includes LZ4F) is enormously faster for very small inputs (~10 MB/s -> ~100 MB/s for 100B inputs, ~40 MB/s -> ~100 MB/s for 1KB inputs).
- Compression which doesn't reuse a context is unaffected.
- Compression on large inputs is unaffected (or sub-1% faster).
- Optimal compression levels (10-12) are 1-3% slower, due to their not being inlined.


Cyan4973:
created branch `HC_dict` for this PR

lib/lz4hc.h
```diff
@@ -153,11 +154,13 @@ typedef struct
     uint32_t   lowLimit;        /* below that point, no more dict */
     uint32_t   nextToUpdate;    /* index from which to continue dictionary update */
     int        compressionLevel;
-} LZ4HC_CCtx_internal;
+    const LZ4HC_CCtx_internal* dictCtx;
```
Cyan4973: This increases the size of `LZ4HC_CCtx_internal`.
The risk is, should someone link to this version of the library, 
while allocating a static memory area based on the size of a previous version, 
it may generate out-of-bound read/write when accessing `dictCtx`.

I thought the plan was to refurbish `uint8_t* inputBuffer` instead, preserving the structure size ?

tests/fuzzer.c
```diff
@@ -626,6 +632,97 @@ static int FUZ_test(U32 seed, U32 nbCycles, const U32 startCycle, const double c
                 FUZ_CHECKTEST(decodedBuffer[blockSize-missingBytes], "LZ4_decompress_safe_usingDict overrun specified output buffer size (-%u byte) (blockSize=%i)", missingBytes, blockSize);
         }   }
 
+        /* Compress using external dictionary stream */
```
Cyan4973: We will need tests for LZ4 HC functions too.

lib/lz4hc.c
```diff
@@ -287,6 +304,41 @@ LZ4HC_InsertAndGetWiderMatch (
         }   }   }   }
     }  /* while ((matchIndex>=lowLimit) && (nbAttempts)) */
 
+    if (dict == usingDictCtx && nbAttempts && ip - base - lowLimit < MAX_DISTANCE) {
+        ptrdiff_t dictIndexDelta = dictCtx->base - dictCtx->end + lowLimit;
```
Cyan4973: seems to be a `const` value.

lib/lz4hc.c
```diff
@@ -287,6 +304,41 @@ LZ4HC_InsertAndGetWiderMatch (
...
+        dictMatchIndex = dictCtx->hashTable[LZ4HC_hashPtr(ip)];
+        matchIndex = dictMatchIndex + (int)dictIndexDelta;
```
Cyan4973: minor : `dictIndexDelta` is casted here, from `ptrdiff_t` to `int`.
In 64-bit mode, `ptrdiff_t` is a 64-bit value, so it can, in theory, overflow `int`.

Considering the initialization of `dictIndexDelta`, the value shouldn't be that large.
Box it with an `assert()` then.

felixhandte: Yeah, should be safe since indexes in both contexts can't grow past 1 GB. I'll add an assert.

lib/lz4hc.h
felixhandte: Ah, very good. Yes. Let me investigate how to do that.

lib/lz4hc.c
```diff
@@ -287,6 +304,41 @@ LZ4HC_InsertAndGetWiderMatch (
...
+        while ((ptrdiff_t) dictMatchIndex + MAX_DISTANCE > ip - base - dictIndexDelta && nbAttempts--) {
```
Cyan4973: minor : `matchIndex` already includes `dictIndexDelta `,
it could be used instead of `dictMatchIndex` to simplify this test.

lib/lz4hc.c
```diff
@@ -287,6 +304,41 @@ LZ4HC_InsertAndGetWiderMatch (
...
+            const BYTE* const matchPtr = dictCtx->base + dictMatchIndex;
+
+            if (LZ4_read32(matchPtr) == pattern) {
+                int mlt;
+                int back = 0;
+                const BYTE* vLimit = ip + (dictCtx->end - matchPtr);
+                if (vLimit > iHighLimit) vLimit = iHighLimit;
+                mlt = LZ4_count(ip+MINMATCH, matchPtr+MINMATCH, vLimit) + MINMATCH;
+                /*
```
Cyan4973: yeah, not continuing the match after dictionary end is a good bet.
You can remove that commented code.

lib/lz4hc.c
```diff
@@ -653,29 +711,96 @@ static int LZ4HC_compress_generic (
...
+    LZ4HC_CCtx_internal* const ctx,
+    const char* const src,
+    char* const dst,
+    int* const srcSizePtr,
+    int const dstCapacity,
+    int cLevel,
+    limitedOutput_directive limit
+    )
+{
+    size_t position = ctx->end - ctx->base - ctx->lowLimit;
```
Cyan4973: minor: `const`

lib/lz4hc.c
```diff
@@ -653,29 +711,96 @@ static int LZ4HC_compress_generic (
...
+    if (position >= 64 KB) {
+        ctx->dictCtx = NULL;
+        return LZ4HC_compress_generic_noDictCtx(ctx, src, dst, srcSizePtr, dstCapacity, cLevel, limit);
+    } else if (position == 0 && *srcSizePtr > 4 KB) {
+        memcpy(ctx, ctx->dictCtx, sizeof(LZ4HC_CCtx_internal));
+        LZ4HC_setExternalDict(ctx, (const BYTE *)src);
+        ctx->compressionLevel = cLevel;
+        return LZ4HC_compress_generic_noDictCtx(ctx, src, dst, srcSizePtr, dstCapacity, cLevel, limit);
+    } else {
+        int result = LZ4HC_compress_generic_internal(ctx, src, dst, srcSizePtr, dstCapacity, cLevel, limit, usingDictCtx);
```
Cyan4973: minor : `return` directly

lib/lz4hc.c
```diff
@@ -684,7 +809,7 @@ int LZ4_compress_HC(const char* src, char* dst, int srcSize, int dstCapacity, in
     LZ4_streamHC_t state;
     LZ4_streamHC_t* const statePtr = &state;
 #endif
-    int const cSize = LZ4_compress_HC_extStateHC(statePtr, src, dst, srcSize, dstCapacity, compressionLevel);
+    int cSize = LZ4_compress_HC_extStateHC(statePtr, src, dst, srcSize, dstCapacity, compressionLevel);
```
Cyan4973: why removing `const` ?

lib/lz4hc.c
```diff
@@ -287,6 +302,39 @@ LZ4HC_InsertAndGetWiderMatch (
         }   }   }   }
     }  /* while ((matchIndex>=lowLimit) && (nbAttempts)) */
 
+    if (dict == usingDictCtx && nbAttempts && ip - base - lowLimit < MAX_DISTANCE) {
+        const ptrdiff_t dictIndexDelta = dictCtx->base - dictCtx->end + lowLimit;
+        /* bounds check, since we need to downcast */
```
Cyan4973: 1 GB is unlikely to be large enough.

I suspect `dictCtx->base - dictCtx->end` could be correctly bounded (<= 64KB ?),
but `lowLimit` can be much larger, and becomes your real concern.

Cyan4973: (concept)
```
size_t const dictMaxOffset_st = dictCtx->end - dictCtx->base;
U32 const dictMaxOffset = (U32)dictMaxOffset_st;   // <--- assert();
U32 const dictIndexDelta = lowLimit - dictMaxOffset;   // defined
matchIndex = dictMatchIndex + dictIndexDelta;  // U32 = U32 + U32
```

lib/lz4hc.c
```diff
@@ -79,21 +79,33 @@
...
     MEM_INIT((void*)hc4->hashTable, 0, sizeof(hc4->hashTable));
     MEM_INIT(hc4->chainTable, 0xFF, sizeof(hc4->chainTable));
-    hc4->nextToUpdate = 64 KB;
-    hc4->base = start - 64 KB;
+}
+
+static void LZ4HC_init (LZ4HC_CCtx_internal* hc4, const BYTE* start)
+{
+    uptrval startingOffset = hc4->end - hc4->base;
+    if (startingOffset > 1 GB || startingOffset > (uptrval)start) {
```
Cyan4973: It seems that the objective of this test is to avoid some overflow later in the code ?

lib/lz4hc.c
```diff
@@ -79,21 +79,33 @@
...
+        LZ4HC_clearTables(hc4);
+        startingOffset = 0;
+    }
+    startingOffset += 64 KB;
```
Cyan4973: What happens if `startingOffset == (uptrval)start - 63 KB` ?

lib/lz4hc.c
```diff
@@ -79,21 +79,33 @@
...
+    hc4->nextToUpdate = (U32) startingOffset;
+    hc4->base = start - startingOffset;
     hc4->end = start;
-    hc4->dictBase = start - 64 KB;
-    hc4->dictLimit = 64 KB;
-    hc4->lowLimit = 64 KB;
+    hc4->dictBase = start - startingOffset;
```
Cyan4973: `dictBase = base` then

lib/lz4hc.c
felixhandte: Yes, the intent is to avoid selecting a new "`currentOffset`" such that our calculation of `base` underflows. However, looking at how `base` is used, I believe it's actually safe for it to underflow, so I'm going to remove this check.

lib/lz4hc.c
```diff
@@ -287,6 +303,37 @@ LZ4HC_InsertAndGetWiderMatch (
...
+        assert(dictEndOffset <= 1 GB);
+        dictMatchIndex = dictCtx->hashTable[LZ4HC_hashPtr(ip)];
+        matchIndex = dictMatchIndex + lowLimit - (U32)dictEndOffset;
+        while (ipIndex - matchIndex <= MAX_DISTANCE && nbAttempts--) {
+            const BYTE* const matchPtr = dictCtx->base + dictMatchIndex;
+
+            if (LZ4_read32(matchPtr) == pattern) {
+                int mlt;
+                int back = 0;
+                const BYTE* vLimit = ip + (dictCtx->end - matchPtr);
```
Cyan4973: minor, more food for thought : 
`dictCtx->end - matchPtr == dictEndOffset - dictMatchIndex`.

The differences are : 
- 2nd calculation is done in `U32` space, rather than as a pointer difference. Ptr difference may more easily invoke undefined behaviors (though, in this case, it should completely fine, since both `end` and `matchPtr` point to the same object).
- arguments for 2nd calculation are available before `matchPtr` is calculated. In theory it's better for performance, but in this case, I don't expect it to make any tangible benefit : there are tons of mis-predictions in this code, which cost a lot more.

lib/lz4hc.c
```diff
@@ -718,7 +843,18 @@ int             LZ4_freeStreamHC (LZ4_streamHC_t* LZ4_streamHCPtr) {
...
     LZ4_STATIC_ASSERT(sizeof(LZ4HC_CCtx_internal) <= sizeof(size_t) * LZ4_STREAMHCSIZE_SIZET);   /* if compilation fails here, LZ4_STREAMHCSIZE must be increased */
+    LZ4_streamHCPtr->internal_donotuse.end = (const BYTE *)(ptrdiff_t)-1;
+    LZ4_streamHCPtr->internal_donotuse.base = NULL;
+    LZ4_streamHCPtr->internal_donotuse.dictCtx = NULL;
+    LZ4_setCompressionLevel(LZ4_streamHCPtr, compressionLevel);
+}
+
+void LZ4_resetStreamHC_fast (LZ4_streamHC_t* LZ4_streamHCPtr, int compressionLevel)
+{
+    LZ4_STATIC_ASSERT(sizeof(LZ4HC_CCtx_internal) <= sizeof(size_t) * LZ4_STREAMHCSIZE_SIZET);   /* if compilation fails here, LZ4_STREAMHCSIZE must be increased */
```
Cyan4973: A single invocation of `LZ4_STATIC_ASSERT()` should be enough. It's only triggered at compile-time.

lib/lz4hc.c
```diff
@@ -718,7 +843,18 @@ int             LZ4_freeStreamHC (LZ4_streamHC_t* LZ4_streamHCPtr) {
...
+    LZ4_streamHCPtr->internal_donotuse.end -= (uptrval)LZ4_streamHCPtr->internal_donotuse.base;
     LZ4_streamHCPtr->internal_donotuse.base = NULL;
```
Cyan4973: Could you explain why you need to move `base` to `NULL` position ?

lib/lz4hc.h
```diff
@@ -141,39 +141,41 @@ LZ4LIB_API int LZ4_saveDictHC (LZ4_streamHC_t* streamHCPtr, char* safeBuffer, in
...
     unsigned int   nextToUpdate;     /* index from which to continue dictionary update */
     int            compressionLevel;
-} LZ4HC_CCtx_internal;
+    const LZ4HC_CCtx_internal* dictCtx;
+};
 
 #endif
 
-#define LZ4_STREAMHCSIZE       (4*LZ4HC_HASHTABLESIZE + 2*LZ4HC_MAXD + 56) /* 262200 */
+#define LZ4_STREAMHCSIZE       (4*LZ4HC_HASHTABLESIZE + 2*LZ4HC_MAXD + 64) /* 262200 */
```
Cyan4973: Since the size of `struct LZ4HC_CCtx_internal` has been preserved, 
I guess it's not useful to increase `LZ4_STREAMHCSIZE` ?
I'm unsure of what kind of side effect this could create.

Cyan4973:
I only have one last comment on `LZ4_STREAMHCSIZE `,
but apart from that, it looks very good.

If you feel it's ready, we can proceed with the merge.

lib/lz4hc.h
felixhandte: Ah, good catch. I'll revert.

PR #494, opened 2018-04-06
Return to Allowing Early Returns in LZ4_compress_generic()

felixhandte:
Or: `goto` Considered Harmful

Or: https://xkcd.com/292/


Cyan4973:
Note that I'm not especially opposed to `goto`.
Sometimes, they do make sense.

It's just that this one seems easy to get rid off.
And the issue solved is less the code of today, which works fine and does the same thing,
than potential mistakes a future contributor may trigger by (typically) issuing an early exit without paying attention to the `_clean_up` tag.

felixhandte:
Yes, for sure.

Issue #485, opened 2018-03-21
Question regarding the  ptr arithmetic in lz4.c


JoachimSchneider:
Dear Developers,
while compiling the lz4 library on OS/400 (the OS running on the IBM midrange AS/400 computers) we found a questionable pointer arithmetic construct in lib/lz4.c:

We guess that line 1046 of lz4/lib/lz4.c should read
  (LZ4_dict->dictionary + LZ4_dict->currentOffset > src)) {   /* address space overflow */
instead of
       ((uptrval)LZ4_dict->currentOffset > (uptrval)src)) {   /* address space overflow */

Attached is a patch file (lz4.c.diff.zip)
[lz4.c.diff.zip](https://github.com/lz4/lz4/files/1833361/lz4.c.diff.zip)

After applying the modification w successfully ran all the tests contained in the lz4/tests directory (make test).

Best Regards,
Joachim Kern and Joachim Schneider, SAP SE

Cyan4973:
Thanks for your report @JoachimSchneider !

I understand a compiler might feel this comparison is questionable, since this test is comparing an `U32` index to a pointer, which are 2 things of different nature.
This is valid though, since both are casted to `uptrval` prior to the comparison.

And the test is correct too.
The issue it solves is the following : `currentOffset` will be used to [calculate `dictBase`](https://github.com/lz4/lz4/blob/dev/lib/lz4.c#L620), which is defined as `src - currentOffset`. Therefore, if `src` < `currentOffset`, the resulting `dictBase` will underflow, and become a high address value. This _may_ cause troubles later on as there are branch decisions based on ptr comparisons, with the assumption that a `match` is necessarily < `currentPtr`. 
Note that problems are not "guaranteed", it also depends on bad luck, distribution of values in the hash table, etc.

Anyway, the main idea here is to make sure `src - currentOffset = dictBase` remains a positive number, so that we can then guarantee that `dictBase < src`.
This test achieve this goal.

However, your compiler complains, maybe because the underlying types, prior casting to `uptrval`, are of different nature.

Maybe something like that could work then : 
`(uptrval)((BYTE*)NULL + LZ4_dict->currentOffset) > (uptrval)src`

It's the same test, but now `(BYTE*)NULL + LZ4_dict->currentOffset` is officially a pointer before its conversion to `uptrval`, so it might calm down compiler warning.

Another potential variant is : 
`(BYTE*)(uptrval)LZ4_dict->currentOffset > src`
[...]

Issue #463, opened 2018-02-04
Question: tableType==byPtr, dict=noDict, dictIssue=noDictIssue


MiloszKrajewski:
Hi,

This is not as issue as everything works fine, I'm just trying to understand how it really works.

In `lz4.c`, in `/* Find a match */` section, the condition in `do { ... } while` loop is:

```
((dictIssue==dictSmall) ? (match < lowRefLimit) : 0)
|| ((tableType==byU16) ? 0 : (match + MAX_DISTANCE < ip))
|| (LZ4_read32(match+refDelta) != LZ4_read32(ip))
```

which looks a little bit mysterious to me.
If `dictIssue==noDictIssue` then first line is evaluated to false.
As `tableType==byPtr` the second line evaluates to `match + MAX_DISTANCE < ip`.
Now there comes the question is it possible, that `ip` is low enough that this expression is actually `false` if match is `null`? 
IN such case third line come into play testing `LZ4_read32(match+regDelta)` which is potentally reading memory at `0x00000000`.

Am I missing something?

MiloszKrajewski:
In my case `refDelta` is `0` as `refDelta = dictDelta` is triggered only if `dict==usingExtDict` (it is `noDict` in my case). The read is (potentially) exactly @ `0x00000000`.

As I said I have no bug because my `ip` is above 64k threshold but I'm thinging about, for example, IoT devices.

BTW: I just use `LZ4_compress_fast((char*)src, (char*)dst, src_len, dst_len, 1);` to compress 1MB block.

PR #497, opened 2018-04-12
Compatibility with low memory addresses

### LINKED ISSUE #397: Streaming decompression issue w/emscripten

Cyan4973:
Current lz4 fast compression implementation has a weakness
when input buffer is provided at a low memory address < 64K.

Several checks are performed as pointer comparisons, but their result depend on the condition that input memory address is > 64K, which is true on most systems.

This patch makes lz4 compatible with low memory addresses.
It adds a test (within `fuzzer`) to generate and test such condition, as part of CI suite.

The patch protects `byU32` mode against such scenario, by making checks based on index values, instead of pointers. This also remove a ptr overflow check, which is known to cause trouble on some IBM systems (see #485).
`byPtr` mode still needs pointer comparisons. It is only available for 32-bits systems, and when there is no dictionary and no streaming. It is now also restricted to buffers starting >= address 64KB. When any of these conditions is not respected, the implementation switches to `byU32`.

Should fix #397 , #463 and #485.


terrelln:
Looks like you accidentally? changed your git name.

lib/lz4hc.c
```diff
@@ -885,7 +886,7 @@ char* LZ4_slideInputBufferHC(void* LZ4HC_Data)
 {
     LZ4HC_CCtx_internal* const hc4 = &((LZ4_streamHC_t*)LZ4HC_Data)->internal_donotuse;
     int const dictSize = LZ4_saveDictHC((LZ4_streamHC_t*)LZ4HC_Data, (char*)(hc4->inputBuffer), 64 KB);
-    return (char*)(hc4->inputBuffer + dictSize);
+    return (char*)(hc4->inputBuffer) + dictSize;
```
terrelln: Should it return a `const char*`, or do we not want to change the API?

lib/lz4.c
```diff
@@ -773,34 +846,66 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
         /* Fill table */
         LZ4_putPosition(ip-2, cctx->hashTable, tableType, base);
 
+#if 1
```
terrelln: Is there a reason to leave the `#if 1` around?

lib/lz4.c
```diff
@@ -721,30 +783,41 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
             /* Copy Literals */
             LZ4_wildCopy(op, anchor, op+litLength);
             op+=litLength;
+            DEBUGLOG(6, "seq.start:%zi, literals=%u, match.start:%zi", anchor-(const BYTE*)source, litLength, ip-(const BYTE*)source);
         }
 
 _next_match:
```
terrelln: Can we list the variables required by `_next_match` in a comment?

Cyan4973:
> Looks like you accidentally? changed your git name.

Right, I had to develop this patch on a Linux VM in order to access capability to generate buffer at low memory addresses. Since it's a new system, it has different settings.

To be fixed...

lib/lz4hc.c
Cyan4973: It must return a non-`const` pointer, since it is the position at which user is supposed to resume filling the buffer with some new input.

It's a pretty poor interface, and that's why it is deprecated.

lib/lz4.c
Cyan4973: I was making some speed comparison tests.
You're right, it shall be removed before merge.

terrelln: I've tried removing it as well, but it was a pretty large loss for me, though I don't remember the exact numbers, maybe 5%.

Cyan4973: yes, I had about the same experience

Cyan4973:
The speed regression was introduced in 64a3e41acaf9e186937d32c9dd2dc104e5bc4a72.

It's important to note that previous patch, 6d931b6a93100fe2cb4dafc032ff486c814d1fed, wasn't correct, as it could not support "advanced" scenarios.
But for the straighforward `lz4 -b` scenario, it would work correctly, and introduce no speed regression.

It might be that 64a3e41acaf9e186937d32c9dd2dc104e5bc4a72 introduced too many changes in a single patch, as the goal was to support remaining scenarios. But the result is that it's difficult to associate the speed regression to one specific element.

So maybe a possible solution is to restart from 6d931b6a93100fe2cb4dafc032ff486c814d1fed, and introduce changes slowly, to see which one trigger any speed regression.

PR #502, opened 2018-04-18
Wrap likely/unlikely macroses with #ifndef

lhacc1:
It prevent redefine error when project which uses lz4 has its own likely/unlikely
macroses.


Cyan4973:
It's a good point @lhacc1 ,
though the implicit assumption made here is that pre-existing `likely`/`unlikely` macros have same meaning. Does that sound, well, likely ?

PR #504, opened 2018-04-19
lib: allow to disable shared libraries

baruchsiach:
Just like BUILD_STATIC=no disables static libraries, BUILD_SHARED=no
disabled shared libraries. This is useful to support toolchains that do
not support shared libraries.


Cyan4973:
Good point @baruchsiach !

I was thinking : we now have a few variables which can influence the build process.
It would probably be a good idea to document them, 
for example in a `README.md` in the same directory.

PR #510, opened 2018-04-23
Fix input size validation edge cases

### LINKED ISSUE #508: poor validation in the 1.8.1 shortcut
The main loop in `LZ4_decompress_generic` conceptually starts as follows:
```c
    /* Main Loop : decode sequences */
    while (1) {
        assert(ip < iend);
        unsigned const token = *ip++;
```
I added the assertion which I claim can be triggered (with a malicious input).
The loop proceeds with the shortcut:
```c
        if (((ip + 14 /*maxLL*/ + 2 /*offset*/ <= iend)
            ...
                ip += ll + 2 /*offset*/;
                continue;
```
When `ll` is `14`, it seems that nothing prevents the shortcut from setting `ip` to `iend` and reentering the loop.

terrelln:
The bug is a read up to 2 bytes past the end of the buffer.
There are three cases for this bug, one for each test case added.

* An empty input causes `token = *ip++` to read one byte too far.
* A one byte input with `(token >> ML_BITS) == RUN_MASK` causes
  one extra byte to be read without validation. This could be
  combined with the first bug to cause 2 extra bytes to be read.
* The case pointed out in issue #508, where `ip == iend` at the
  beginning of the loop after taking the shortcut.

Benchmarks show no regressions on clang or gcc-7 on both my mac
and devserver.

Fixes #508.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1520,6 +1520,7 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
     if ((partialDecoding) && (oexit > oend-MFLIMIT)) oexit = oend-MFLIMIT;                      /* targetOutputSize too high => just decode everything */
     if ((endOnInput) && (unlikely(outputSize==0))) return ((srcSize==1) && (*ip==0)) ? 0 : -1;  /* Empty output buffer */
     if ((!endOnInput) && (unlikely(outputSize==0))) return (*ip==0?1:-1);
+    if ((endOnInput) && unlikely(srcSize==0)) return -1;
 
     /* Main Loop : decode sequences */
     while (1) {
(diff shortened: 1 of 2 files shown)


Cyan4973:
perfect fix, with all relevant test cases included in fuzzer test suite.

terrelln:
The new assertion is still failing in https://ci.appveyor.com/project/YannCollet/lz4-1lndh/build/1.0.676/job/pm7sa9krbun6nf45#L174, to be investigated.

terrelln:
The assert is fixed, just forgot to check `endOnInput` in the assert.

I fuzzed using this fuzzer:

```c
#include "lz4.h"
#include <stdint.h>
#include <stddef.h>

static char out[8192];
static size_t const outsize = sizeof(out);

extern "C" int LLVMFuzzerTestOneInput(uint8_t const* data, size_t size)
{
  LZ4_decompress_safe((char const*)data, out, size, outsize);
  return 0;
}
```

It found all 3 bugs immediately, and hasn't found anything since the fix.

Cyan4973:
I guess one key property to catch these bugs
is that `data` buffer must be sized to be _exactly_ `size_t size`
in order for sanitizers to catch the read-after-input-buffer bug.

Issue #506, opened 2018-04-21
the 1.8.1 shortcut slows down decompress_fast


svpv:
The shortcut introduced in cc4a109b0df7d12cfd0f26960ba3548d9b7915ba only benefits the `decompress_safe` routines, but slows down the `decompress_fast` routines (this is because the input length is not known, so the condition `ip + ... <= iend` never holds). In fact, `decompress_fast` is now **slower by about 9%** than `decompress_safe`. And if you remove the shortcut from `decompress_fast`, but keep the shortcut in `decompress_safe`, the latter is still faster. At least with the kind of data I'm using. If the same holds true for other kinds of data, this may leave no chance for the `decompress_fast` family after all.

Cyan4973:
I confirm the issue.
Using latest `dev` branch and test program `fullbench`, I measure on my laptop : 
```
 4-LZ4_decompress_safe           : 211948032 ->  2518.2 MB/s
```
versus
```
 1-LZ4_decompress_fast           : 211948032 ->  2302.9 MB/s
```

Your analysis is also spot on : the shortcut is effectively disabled for `_fast` variant, due to input size being unknown.

In feature branch `fasterFastDec`, I simply enabled the shortcut for `_fast` variant, by setting input size arbitrarily to `LZ4_compressBound(originalSize)`. It seems to work : 
```
 1-LZ4_decompress_fast           : 211948032 ->  2573.6 MB/s
```
Now the `_fast` variant is once again faster, if only by very little.

Generally speaking, we don't recommend the `_fast` variant, due to security concerns.
The only valid scenario is when data can be "trusted", which is generally limited to data being produced and consumed on the same machine.

svpv:
Oh, I didn't realize that the shortcut could be enabled in the `decompress_fast` mode, exactly because the input size is unknown and assumptions about the size can work to our advantage. So indeed the code works faster on the `fasterFastDec` branch.  Although still not as fast as `decompress_safe`, with the kind of data that I'm using.

However, there seems to be a better way out.  If we can make assumptions about the input size which work to our advantage, we can further eliminate the input size check altogether, by propagating the `safeDecode` constant:
```c
        if ((safeDecode ? (ip + 14 /*maxLL*/ + 2 /*offset*/ <= iend) : 1)
          &               (op + 14 /*maxLL*/ + 18 /*maxML*/ <= oend)
          & ((token < (15<<ML_BITS)) & ((token & ML_MASK) != 15)) ) {
```
[...]

PR #511, opened 2018-04-24
Fixed performance issue with LZ4_decompress_fast()

Cyan4973:
by disabling the shortcut, which it cannot use anyway
(see #506 )

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1536,9 +1536,10 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
          * this shortcut was tested on x86 and x64, where it improves decoding speed.
          * it has not yet been benchmarked on ARM, Power, mips, etc.
          * NOTE: The loop begins with a read, so we must have one byte left at the end. */
-        if (((ip + 14 /*maxLL*/ + 2 /*offset*/ < iend)
-          & (op + 14 /*maxLL*/ + 18 /*maxML*/ <= oend))
-          & ((token < (15<<ML_BITS)) & ((token & ML_MASK) != 15)) ) {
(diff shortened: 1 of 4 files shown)


lib/lz4.c
```diff
@@ -1536,9 +1536,10 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
          * this shortcut was tested on x86 and x64, where it improves decoding speed.
          * it has not yet been benchmarked on ARM, Power, mips, etc.
          * NOTE: The loop begins with a read, so we must have one byte left at the end. */
-        if (((ip + 14 /*maxLL*/ + 2 /*offset*/ < iend)
-          & (op + 14 /*maxLL*/ + 18 /*maxML*/ <= oend))
-          & ((token < (15<<ML_BITS)) & ((token & ML_MASK) != 15)) ) {
+        if (endOnInput
+          && ((ip + 14 /*maxLL*/ + 2 /*offset*/ < iend)
```
terrelln: Can you add extra parens to the rest of the statement, so we aren't mixing `&&` and `&`?

Cyan4973:
As a minor note,
this patch makes a good work at improving performance of `LZ4_decompress_fast()`,
but as a side effect, it decreases a bit performance of `LZ4_decompress_safe()` on gcc 4.8.5.
Which does not make a lot of sense : performance should remain the same.

However, performance impact on `LZ4_decompress_safe()` remains neutral for `clang` and `gcc-7`. So it could be a weakness of `gcc` older version. As gcc 4.8 is unmaintained, I guess it's not a large concern, and good results on maintained compiler versions matters more.

PR #513, opened 2018-04-24
Integrate Contents of `lz4frame_static.h` into `lz4frame.h`

felixhandte:
This PR moves all of the declarations in `lz4frame_static.h` into `lz4frame.h`, protected by a macro, `LZ4F_STATIC_LINKING_ONLY`. This brings LZ4F into using the same pattern as LZ4 and Zstd. This diff safely preserves functionality for existing includes of `lz4frame_static.h` and in general should be safe for multiple inclusion with and without the guard macro defined.

### DIFF
--- a/lib/lz4frame.c
+++ b/lib/lz4frame.c
@@ -73,7 +73,8 @@ You can contact the author at :
 /*-************************************
 *  Includes
 **************************************/
-#include "lz4frame_static.h"
+#define LZ4F_STATIC_LINKING_ONLY
+#include "lz4frame.h"
 #define LZ4_STATIC_LINKING_ONLY
 #include "lz4.h"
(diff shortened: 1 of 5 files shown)


Cyan4973:
Looks good to me !
Great presentation, nice clarity.

The only minor point I was considering is related to this sentence : 
> in general should be safe for multiple inclusion with and without the guard macro defined.

How are we going to test that ?

For example, multi-include within `frametest.c` ?

felixhandte:
I believe it's safe now because the PR is two commits, one in which I move all the declarations, and one in which I change other file's includes, and the first commit tests successfully. If you want that assurance going forward though, I can add some redundancies to the include statements in `frametest.c`, sure.

PR #515, opened 2018-04-25
lz4.c: refactor the decoding routines

svpv:
I noticed that LZ4_decompress_generic is sometimes instantiated with
identical set of parameters, or (what's worse) with a subtly different
sets of parameters.  For example, LZ4_decompress_fast_withPrefix64k is
instantiated as follows:

    return LZ4_decompress_generic(source, dest, 0, originalSize, endOnOutputSize,
		full, 0, withPrefix64k, (BYTE*)dest - 64 KB, NULL, 64 KB);

while the equivalent withPrefix64k call in LZ4_decompress_usingDict_generic
passes 0 for the last argument instead of 64 KB.  It turns out that there
is no difference in this case: if you change 64 KB to 0 KB in
LZ4_decompress_fast_withPrefix64k, you get the same binary code.

Moreover, because it's been clarified that LZ4_decompress_fast doesn't
check match offsets, it is now obvious that both of these fast/withPrefix64k
instantiations are simply redundant.  Exactly because LZ4_decompress_fast
doesn't check offsets, it serves well with any prefixed dictionary.

There's a difference, though, with LZ4_decompress_safe_withPrefix64k.
It also passes 64 KB as the last argument, and if you change that to 0,
as in LZ4_decompress_usingDict_generic, you get a completely different
binary code.  It seems that passing 0 enables offset checking:

    const int checkOffset = ((safeDecode) && (dictSize < (int)(64 KB)));

However, the resulting code seems to run a bit faster.  How come
enabling extra checks can make the code run faster?  Curiouser and
curiouser!  This needs extra study.  Currently I take the view that
the dictSize should be set to non-zero when nothing else will do,
i.e. when passing the external dictionary via dictStart.  Otherwise,
lowPrefix betrays just enough information about the dictionary.

    * * *

Anyway, with this change, I instantiate all the necessary cases as
functions with distinctive names, which also take fewer arguments and
are therefore less error-prone.  I also make the functions non-inline.
(The compiler won't inline the functions because they are used more than
once.  Hence I attach LZ4_FORCE_O2_GCC_PPC64LE to the instances while
removing from the callers.)  The number of instances is now is reduced
from 18 (safe+fast+partial+4*continue+4*prefix+4*dict+2*prefix64+forceExtDict)
down to 7 (safe+fast+partial+2*prefix+2*dict).  The size of the code is
not the only issue here.  Separate helper function are much more
amenable to profile-guided optimization: it is enough to profile only
a few basic functions, while the other less-often used functions, such
as LZ4_decompress_*_continue, will benefit automatically.

This is the list of LZ4_decompress* functions in liblz4.so, sorted by size.
Exported functions are marked with a capital T.
```
$ nm -S lib/liblz4.so |grep -wi T |grep LZ4_decompress |sort -k2
0000000000016260 0000000000000005 T LZ4_decompress_fast_withPrefix64k
0000000000016dc0 0000000000000025 T LZ4_decompress_fast_usingDict
0000000000016d80 0000000000000040 T LZ4_decompress_safe_usingDict
0000000000016d10 000000000000006b T LZ4_decompress_fast_continue
0000000000016c70 000000000000009f T LZ4_decompress_safe_continue
00000000000156c0 000000000000059c T LZ4_decompress_fast
0000000000014a90 00000000000005fa T LZ4_decompress_safe
0000000000015c60 00000000000005fa T LZ4_decompress_safe_withPrefix64k
0000000000002280 00000000000005fa t LZ4_decompress_safe_withSmallPrefix
0000000000015090 000000000000062f T LZ4_decompress_safe_partial
0000000000002880 00000000000008ea t LZ4_decompress_fast_extDict
0000000000016270 0000000000000993 t LZ4_decompress_safe_forceExtDict
```

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -92,6 +92,7 @@
 *  Dependency
 **************************************/
 #define LZ4_STATIC_LINKING_ONLY
+#define LZ4_DISABLE_DEPRECATE_WARNINGS /* due to LZ4_decompress_safe_withPrefix64k */
 #include "lz4.h"
 /* see also "memory routines" below */
 
@@ -1666,6 +1667,8 @@ _output_error:
 }
 
 
+/*===== Instantiate the API decoding functions. =====*/
+
 LZ4_FORCE_O2_GCC_PPC64LE
 int LZ4_decompress_safe(const char* source, char* dest, int compressedSize, int maxDecompressedSize)
 {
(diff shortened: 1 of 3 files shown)


Cyan4973:
Thanks @svpv .
This is a great proposition.

Indeed, it helps clarify and partially tame the combinatorial explosion of decompression variants.
I believe it's a great starting point.

Note though that test failures suggest it's going a bit too far, removing ring buffer support in the process. I'll look deeper in the code to see where it may happen.

Finally, we'll need some benchmark figures to ensure performance is preserved.
You can use `tests/fullbench -d $FILE` for this purpose.

svpv:
I think I can see now what my fault was. I didn't realize that the "double dictionary mode" was possible, i.e. that the decoding routine can look both at the first part of the dictionary passed as prefix and the second part passed via dictStart+dictSize.  So let me fix it, if you haven't looked into it yet. I think I'll post another commit on this branch / pull request.

Cyan4973:
That looks pretty good !
Do you have some benchmark figures, comparing the decompression speed at different modes before and after your changes ?
See above note on usage of `tests/fullbench -d $FILE`

svpv:
The branch is now in a consistent state, but unfortunately I have another idea. :-) Let's look closer at these two instantiations:
```c
// LZ4_decompress_safe
    return LZ4_decompress_generic(source, dest, compressedSize, maxDecompressedSize,
                                  endOnInputSize, full, 0, noDict,
                                  (BYTE*)dest, NULL, 0);
// LZ4_decompress_safe_withSmallPrefix
    return LZ4_decompress_generic(source, dest, compressedSize, maxOutputSize,
                                  endOnInputSize, full, 0, noDict,
                                  (BYTE*)dest-prefixSize, NULL, 0);
```
The only difference is in the `lowPrefix` parameter to `LZ4_decompress_generic`&#8239;&mdash;&thinsp;whether it's  just `dest` or `dest-prefixSize`&#8239;&mdash;&thinsp;and `dest` is already passed as the second parameter. So, first I thought, like, the compiler can save a register because it can spot that two parameters are the same (and this justifies a separate case for `LZ4_decompress_safe`).

Or can it? You see, saving a register is only important in a loop. Saving a register in the prologue is unimportant.  In the loop, the offsets are checked against `lowPrefix` but the `dst` parameter is only used to initialize the `op` pointer, which gets advanced in the loop; `dst` is otherwise used only in successful return:
```c
       return (int) (((char*)op)-dst);     /* Nb of output bytes decoded */
```
So, with a decent compiler, the loop should have the same register pressure/spilling. So let's think a bit whether it's worth to reduce the number of base cases/instantiations even further.

Cyan4973:
We are converging towards a new release soon,
so it would be good to consolidate benefits before that point,
further improvements can still be delivered afterwards.

For this PR, I merely need 2 metrics : 
- What's the benefit ? I expect the binary size to be improved, as a result of generating less variants. Code clarity is also a benefit, though it's more difficult to associate a figure
- What's the impact ? Well, in this case, the goal is to show it has a _neutral_ impact, hence no adverse performance on decompression speed.

Cyan4973:
I could merge your patch into a test branch based on current `dev`.
The speed looks good enough, with all measures within noise level.

But there is one measure which looks _too_ good : 
`LZ4_decompress_fast_usingDict()` has basically the same speed as `LZ4_decompress_fast()` with your patch, while it takes a noticeable hit within `dev` branch.

I'm now wondering if `LZ4_decompress_fast_usingDict()` is really doing its job, or has been simplified to become just the same as `LZ4_decompress_fast()`, which would be incorrect.

Cyan4973:
Indeed,
the test in `fullbench` measuring speed of `LZ4_decompress_fast_usingDict()`
lays the 64 KB dictionary right before the output buffer, 
so it's technically a 64KB prefix, 
in which case your implementation makes it behave like `LZ4_decompress_fast()`.
Hence same speed.

I added another test which separates the dictionary from output buffer by just one byte.
It's enough to observe a performance difference, as expected.

So this is a nice property of your patch : it also improves performance for some corner cases it can detect and deal with.

Nice job !

PR #516, opened 2018-04-25
Merge _destSize Compress Variant into LZ4_compress_generic()

felixhandte:


lib/lz4.c
```diff
@@ -795,9 +797,14 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
         /* Encode Literals */
         {   unsigned const litLength = (unsigned)(ip - anchor);
             token = op++;
-            if ((outputLimited) &&  /* Check output buffer overflow */
+            if ((outputLimited == limitedOutput) &&  /* Check output buffer overflow */
                 (unlikely(op + litLength + (2 + 1 + LASTLITERALS) + (litLength/255) > olimit)))
                 return 0;
+            if ((outputLimited == fillOutput) &&
+                (unlikely(op + (litLength+240)/255 /* litlen */ + litLength /* literals */ + 2 /* offset */ + 1 /* token */ + MFLIMIT - MINMATCH /* min last literals so last match is <= end - MFLIMIT */ > olimit))) {
```
felixhandte: Technically, all we need to check here is that there's enough space for the literals to avoid a buffer overrun before we get to the check below. Not sure if it would be beneficial to perform a simpler check.

lib/lz4.c
```diff
@@ -822,6 +829,14 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
          * - token and *token : position to write 4-bits for match length; higher 4-bits for literal length supposed already written
          */
 
+        if ((outputLimited == fillOutput) &&
+            (op + 2 /* offset */ + 1 /* token */ + MFLIMIT - MINMATCH /* min last literals so last match is <= end - MFLIMIT */ > olimit)) {
```
felixhandte: One comment about this test. Technically this is more cautious than it needs to be. You only need to reserve space for `MFLIMIT - MINMATCH` literals at the end in the case that the last match is `MINMATCH` long. For long matches, you only need space for `LASTLITERALS` bytes. However, we don't yet know the length of the match here, and we do have to do some kind of check to make sure it's safe to write the offset, so we can't just move this down to where we know the match length. It's only a couple bytes difference, so I suspect it's not worth trying to engineer a way to be perfect.

Cyan4973:
> I realized neither `fullbench` nor my testing code tests the `destSize` variants.

Yes, that's a good point !
It's definitely worth adding !

lib/lz4.c
Cyan4973: The check looks good to me

lib/lz4.c
Cyan4973: I fully agree.
That looks good enough to me.

felixhandte:
The new (very basic) test in `fullbench` looks good! ~355MB/s on dev vs ~375MB/s on this branch.

Cyan4973:
Looks great to me !

`_destSize()` variant will now automatically benefit from maintenance of `LZ4_compress_generic()`.
And it's good to finally have the ability to speed test this variant !

PR #521, opened 2018-04-27
fix lz4hc -BD non-determinism

### LINKED ISSUE #230: Parameter -BD can be non-deterministic
(tested on lz4 r131)
Running `lz4 -9 enwik8 - > /dev/null` always shows, independent of CPU/OS:
`Compressed 100000000 bytes into 42283904 bytes ==> 42.28%`

Adding the `-BD` parameter produces different filesizes:
CentOS6/lz4 (x64) from EPEL repository:

```
Compressed 100000000 bytes into 42210479 bytes ==> 42.21%
Compressed 100000000 bytes into 42210483 bytes ==> 42.21%
Compressed 100000000 bytes into 42210476 bytes ==> 42.21%
```

=> varying

Ubuntu 15.10 (Odroid C1 / ARM cpu / lz4 git HEAD g++ 5.2.1):
`Compressed 100000000 bytes into 42210493 bytes ==> 42.21%` (always same size)

Debian 6.0.10 (Xeon X3323 / lz4 git HEAD g++ 4.4.5):
`Compressed 100000000 bytes into 42210461 bytes ==> 42.21%` (always same size)

Cyan4973:
fix #230.

This is a non-trivial bug,
related to the way chain table is updated at end of block, in `_continue()` mode.
Depending on ptr address overflow, it could accidentally trigger partial update skip, 
hence different evolution of the chain table,
hence different compressed file.

Compressed files were always valid.
It's just they missed a few opportunities, and the missed part was basically random.
Skipped sections are generally quite small, and far apart,
so the impact to compression was very small (a few bytes on files > 100 MB).

Still, this patch fixes it, now producing always same compressed file, even on CentOS6.

### DIFF
--- a/lib/lz4hc.c
+++ b/lib/lz4hc.c
@@ -699,8 +699,6 @@ LZ4_FORCE_INLINE int LZ4HC_compress_generic_internal (
     ctx->end += *srcSizePtr;
     if (cLevel < 1) cLevel = LZ4HC_CLEVEL_DEFAULT;   /* note : convention is different from lz4frame, maybe something to review */
     cLevel = MIN(LZ4HC_CLEVEL_MAX, cLevel);
-    assert(cLevel >= 0);
-    assert(cLevel <= LZ4HC_CLEVEL_MAX);
     {   cParams_t const cParam = clTable[cLevel];
         if (cParam.strat == lz4hc)
             return LZ4HC_compress_hashChain(ctx,
@@ -892,7 +890,8 @@ void LZ4_attach_HC_dictionary(LZ4_streamHC_t *working_stream, const LZ4_streamHC
 static void LZ4HC_setExternalDict(LZ4HC_CCtx_internal* ctxPtr, const BYTE* newBlock)
 {
     DEBUGLOG(4, "LZ4HC_setExternalDict(%p, %p)", ctxPtr, newBlock);
-    if (ctxPtr->end >= ctxPtr->base + 4) LZ4HC_Insert (ctxPtr, ctxPtr->end-3);   /* Referencing remaining dictionary content */
+    if (ctxPtr->end >= ctxPtr->base + ctxPtr->dictLimit + 4)
+        LZ4HC_Insert (ctxPtr, ctxPtr->end-3);   /* Referencing remaining dictionary content */
 
     /* Only one memory segment for extDict, so any previous extDict is lost at this stage */
     ctxPtr->lowLimit  = ctxPtr->dictLimit;
--- a/lib/lz4hc.c
+++ b/lib/lz4hc.c
@@ -223,7 +223,7 @@ LZ4HC_InsertAndGetWiderMatch (
     const U32 dictLimit = hc4->dictLimit;
     const BYTE* const lowPrefixPtr = base + dictLimit;
     const U32 ipIndex = (U32)(ip - base);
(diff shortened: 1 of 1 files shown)


lib/lz4hc.c
```diff
@@ -247,34 +248,35 @@ LZ4HC_InsertAndGetWiderMatch (
...
                         longest = mlt;
                         *matchpos = matchPtr+back;
                         *startpos = ip+back;
-                }   }
-            }
+            }   }   }
         } else {   /* matchIndex < dictLimit */
             const BYTE* const matchPtr = dictBase + matchIndex;
             if (LZ4_read32(matchPtr) == pattern) {
+                const BYTE* const dictLowLimit = dictBase + hc4->lowLimit;
```
terrelln: I think this should be given a different name or use `lowLimit` directly. `dictLowLimit` implices it is related to `lowLimit`, which it sort of is. However, `dictLowLimit` might include space that is below `lowLimit`. It is okay in this scenario, since we are only using it for backwards counting, so we already have a fixed offset, but someone might confuse it in the future.

lib/lz4hc.c
```diff
@@ -583,11 +582,12 @@ LZ4_FORCE_INLINE int LZ4HC_compress_hashChain (
...
         /*
-        * OK, now we have 3 ascending matches; let's write at least the first one
-        * ip & ref are known; Now for ml
+        * OK, now we have 3 ascending matches;
+        * let's write the first one ML1.
+        * ip & ref are known; Now decide ml.
         */
         if (start2 < ip+ml) {
-            if ((start2 - ip) < (int)ML_MASK) {
+            if ((start2 - ip) < OPTIMAL_ML) {
```
terrelln: Why change this?

Cyan4973: I'm pretty sure that's the correct intention : 
the algorithm wants to shorten matches > `OPTIMAL_ML`, in order to save an additional length byte. When length `< OPTIMAL_ML`, it's not necessary to trigger the shortener.

Note that, strangely enough, this change doesn't impact compression result, 
presumably because this check is actually redundant with some previous check, completed multiple lines above.
Which also leads me to believe this code could be better optimized.
But that's a much more complex task, and I did not want to make this patch any more complex.

lib/lz4hc.c
Cyan4973: I agree, the issue here is using `lowLimit` to mean different things.
Unfortunately, I failed to find a clear enough naming convention to fix that situation...

terrelln: Is it still correct to use `lowLimit` instead of `hc4->lowLimit` here? That way we can keep the dual-meaning consistent.

Cyan4973: I believe it's not.
`lowLimit` is for initial distance check, from `ip`.
`hc4->lowLimit` is the place we must stop looking backward because there is no more data.
Sometimes, both correspond. But in many cases, they do not.

Note that, the compression impact is extremely limited : one would need an initial match very close to the distance limit (65534 bytes for example) and then multiple bytes of look back. In which case, using` lowLimit` instead of `hc4->lowLimit`, the look-back stage would be limited, and potentially miss a few bytes of compression.

So using `lowLimit` still works, and negatively impact compression ratio by a negligible amount.
But it's not correct. Hence the proposed fix.

Cyan4973: Changes variables name to `dictStart`

Cyan4973: Also renamed `lowLimit` to become `lowestMatchIndex`
which seems more accurate.

PR #524, opened 2018-04-28
lz4.c: two-stage shortcut for LZ4_decompress_generic

svpv:
This is an early version of a two-stage shortcut that can improve the decoding speed **by 10%**. The shortcut is not as simple as it could be, because I encountered another issue with LZ4 bound checking.  The question is: if `LZ4_decompress_safe` returns `uncompressedSize <= dstCapacity`, can it clobber the data in between `uncompressedSize` and `dstCapacity`?  It's undocumented!  And it determines what the shortcut can do and what it can't, in the worst-case scenario.

So I found that the shortcut can't always write to `dstCapacity` because this can clobber part of the dictionary in the ring buffer.  This is pretty annoying.  The shortcut could otherwise be just that:
```diff
-    /* Can we copy the literals with a single memcpy invocation?  Sometimes we can't
-     * copy 16 bytes, because they can clobber the dictionary in the ring buffer. */
-    if (!endOnInput /* only 8 bytes */ || /* nothing to clobber */ dict != usingExtDict) {
+    if (1) {

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1397,6 +1397,9 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
     const int safeDecode = (endOnInput==endOnInputSize);
     const int checkOffset = ((safeDecode) && (dictSize < (int)(64 KB)));
 
+    /* Set up the "end" pointers for the shortcut. */
+    const BYTE* const shortiend = iend - (endOnInput ? 14 : 8) /*maxLL*/ - 2 /*offset*/;
+    const BYTE* const shortoend = oend - (endOnInput ? 14 : 8) /*maxLL*/ - 18 /*maxML*/;
 
     /* Special cases */
     if ((partialDecoding) && (oexit > oend-MFLIMIT)) oexit = oend-MFLIMIT;                      /* targetOutputSize too high => just decode everything */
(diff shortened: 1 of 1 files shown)


svpv:
What I initially diagnosed as writing past `uncompressedSize` was probably wrong. It is more likely that writing the literals can clobber its own match.

Cyan4973:
Indeed,
in a situation where the ring buffer is <= 64KB,
writing literals unconditionally overwrites a few bytes,
and as a consequence, potentially corrupt a match with sufficiently large offset (`ringBufferSize - fewBytes`).

Note that a similar situation can happen with matches : writing unconditionally > 8 bytes
can corrupt data beyond next match (if both literal length and match length are small enough).

There are a number of conditions required to trigger the issue, so probability is small, hence it might be difficult to produce, but it's definitely not null.

There are solutions to this issues : 
- Increase ring buffer size. For example, `LZ4F` uses ring buffer size of 128 KB, which prevents such situation to happen.
- Use double buffer strategy : it ensures that matches can only come from currently decoded data or from previous buffer, so overwriting a few bytes within current buffer is not an issue. It's an encoder side restriction though.

Cyan4973:
A few more comments on this PR : 

- first, 10% decompression speed improvement is an impressive outcome, and it's definitely interesting to investigate.
- `LZ4_decompress_safe()` API only guarantees to never write beyond `dstCapacity`. It doesn't attempt to provide any such objective beyond `uncompressedSize`, which it only discovers during decompression. So indeed, it can write a few bytes beyond `uncompressedSize`. And you can simplify your patch.
- In case of a ring buffer, a user must respect a set of conditions, [described in code comments](https://github.com/lz4/lz4/blob/dev/lib/lz4.h#L319). It's pretty complex. The general idea is : provide a ring buffer which size is large enough to guarantee that any match at maximum distance is not corrupted, _or_, guarantee that compression and decompression are synchronous (same buffer size, same update position rule), and each block knows its exact boundaries (decompressed size), preventing any risk of modified bytes beyond block size.

I suspect your patch would work with these scenarios, as long as they respect the conditions.

Cyan4973:
As we are getting close to release deadline, this topic needs to progress.
It will be integrated through a feature branch,
so that we have enough time and latitude to test and modify before merge.

svpv:
I think the patch is good to go, it tries hard not to break anything, and that's why it is more complicated than necessary. Reckless 16-byte copying of literals (as opposed to just 8-byte copying) can clobber something the ring buffer, that's why I handle the 16-byte coping with `usingExtDict` specially.  But we had better figure out the tight bound on literal copying without clobbering.  I'll look into this soon.

Cyan4973:
This patch is transferred to #528,
where it's paired with clarifications on ring buffer scenarios,
completed with new function and new test case.

I can confirm that it brings all the performance benefits expected, which is plain impressive.

PR #528, opened 2018-05-02
Faster decoding speed

Cyan4973:
This patch introduces the 2-stage shortcut, proposed by @svpv,
which improves decoding speed by about +10%.

While reviewing this patch, it became clear that 
clarifying ring buffer scenarios was a necessity.
To this end, documentation is updated,
and a new function is proposed : `LZ4_decoderRingBufferSize(maxBlockSize)`.
This is the minimum size above which LZ4 decompression is guaranteed to work 
(provided source is an LZ4 format compliant block of size `<= maxBlockSize`).
Ring buffer tests cases are also updated to use the new function.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1431,7 +1431,7 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
             /* strictly "less than" on input, to re-enter the loop with at least one byte */
             likely((endOnInput ? ip < shortiend : 1) & (op <= shortoend)))
         {
-            /* Copy the literals. */
+            /* Copy the literals */
             memcpy(op, ip, endOnInput ? 16 : 8);
             op += length; ip += length;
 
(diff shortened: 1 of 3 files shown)


lib/lz4.c
```diff
@@ -1407,39 +1410,57 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
...
+            op += length; ip += length;
+
+            /* The second stage: prepare for match copying, decode full info.
+             * If it doesn't work out, the info won't be wasted. */
+            length = token & ML_MASK; /* match length */
+            offset = LZ4_readLE16(ip); ip += 2;
+            match = op - offset;
+
+            /* Do not deal with overlapping matches. */
+            if ((length != 15) && (offset >= 8) &&
```
terrelln: Use `RUN_MASK` here.

lib/lz4.c
```diff
@@ -1407,39 +1410,57 @@ LZ4_FORCE_INLINE int LZ4_decompress_generic(
...
+         * (in the fast mode, only 8 bytes can be safely copied this way).
+         * 2) Further if the match length is 4..18, copy 18 bytes in a similar
+         * manner; but we ensure that there's enough space in the output for
+         * those 18 bytes earlier, upon entering the shortcut (in other words,
+         * there is a combined check for both stages).
+         */
+        if ((endOnInput ? length != RUN_MASK : length <= 8) &&
+            /* strictly "less than" on input, to re-enter the loop with at least one byte */
+            likely((endOnInput ? ip < shortiend : 1) & (op <= shortoend)))
+        {
```
terrelln: Can we braces style?

lib/lz4.c
```diff
@@ -1681,6 +1704,26 @@ int LZ4_setStreamDecode (LZ4_streamDecode_t* LZ4_streamDecode, const char* dicti
     return 1;
 }
 
+/*! LZ4_decoderRingBufferSize() :
+ *  when setting a ring buffer for streaming decompression (optional scenario),
+ *  provides the minimum size of this ring buffer
+ *  to be compatible with any source respecting maxBlockSize condition.
```
terrelln: nit: can we fix the line wrapping?

lib/lz4.c
Cyan4973: Since it's about match length, `ML_MASK` is likely more appropriate.

PR #529, opened 2018-05-03
LZ4F: Only Reset the LZ4_stream_t when Init'ing a Streaming Block

felixhandte:
`LZ4_resetStream_fast()` will do a full reset if the table isn't in `byU32` mode, since it's built to prep the stream for streaming compression, i.e., `LZ4_compress_fast_continue()`. This can lead to unnecessary resets, such as when the stream is in `byU16` mode and will next be used in `byU16`, which can happen when performing sequential, independent, small compressions and which shouldn't require a reset.

This PR also clarifies some comments to be more explicit about how fastResets work.

### DIFF
--- a/lib/lz4.h
+++ b/lib/lz4.h
@@ -358,6 +358,15 @@ LZ4LIB_API int LZ4_decompress_fast_usingDict (const char* src, char* dst, int or
 #ifdef LZ4_STATIC_LINKING_ONLY
 
 /*! LZ4_resetStream_fast() :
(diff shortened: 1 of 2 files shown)


Cyan4973:
Any performance figure, showing the benefit of this change ?

felixhandte:
Here's a rough comparison:

montage-dickens-gcc-00-c32e031-5406c2e-all-cls

The base revision had a lot more time to stabilize, so I don't believe anything that appears as a regression actually exists.

Cyan4973:
Thanks @felixhandte , it's definitely significant

PR #531, opened 2018-05-03
Preparing v1.8.2

Cyan4973:
Summarized list of changes : 

perf: *much* faster dictionary compression on small files, by @felixhandte
perf: improved decompression speed and binary size, by Alexey Tourbin (@svpv)
perf: slightly faster HC compression and decompression speed
perf: very small compression ratio improvement
fix : compression compatible with low memory addresses (< 0xFFFF)
fix : decompression segfault when provided with NULL input, by @terrelln
cli : new command `--favor-decSpeed`
cli : benchmark mode more accurate for small inputs
fullbench : can bench `_destSize()` variants, by @felixhandte
doc : clarified block format parsing restrictions, by Alexey Tourbin (@svpv)


felixhandte:
Clarification: compression is faster with or without a dictionary.

PR #535, opened 2018-05-04
Correctly deal with set MANDIR variable

paulmenzel:


programs/Makefile
```diff
@@ -131,11 +131,13 @@ mandir      ?= $(datarootdir)/man
 man1dir     ?= $(mandir)/man1
```
Cyan4973: A consequence of this patch is to make `man1dir` effectively inactive.
However, according to https://www.gnu.org/prep/standards/html_node/Directory-Variables.html : 

```
‘man1dir’
The directory for installing section 1 man pages. Write it as $(mandir)/man1.
```

Note that the variable is defined as _lowercase_.

So someone using : 
`make install man1dir=somePath`
won't be able to influence man installation directory.

paulmenzel: So, `MAN1DIR ?= $(MANDIR)/man1` should only be conditionally set.

paulmenzel: Looking at [1], automake doesn’t seem to allow to override it, if I am not mistaken.

[1]: http://www.rpi.edu/dept/acm/packages/fluxbox/fluxbox-0.1.7/doc/Makefile.in

programs/Makefile
```diff
@@ -131,11 +131,13 @@ mandir      ?= $(datarootdir)/man
...
 
 ifneq (,$(filter $(shell uname),OpenBSD FreeBSD NetBSD DragonFly SunOS))
-MANDIR  ?= $(PREFIX)/man/man1
```
Cyan4973: This patch changes the meaning of `MANDIR` in FreeBSD context.

As I have no experience on the topic, and the [FreeBSD documentation on Man Pages](https://www.freebsd.org/doc/en/books/porters-handbook/makefile-manpages.html) is a bit too short, I have no idea what should be done.

As a consequence, I've retained earlier patch from FreeBSD contributor, which makes `MANDIR` the final destination path of the man page.
This patch would change that.

And I guess there is a name conflict here : `MANDIR` has 2 different meanings, depending on your context, or the FreeBSD context.

programs/Makefile
Cyan4973: I don't see any `MANDIR` variable in the automake example.

programs/Makefile
Cyan4973: It's actually possible to change `MANDIR` definition without breaking FreeBSD.
So that part is fine.

Cyan4973:
The only remaining comment is to make sure that `man1dir` is still an active variable, that users can modify on command line if need be.

PR #537, opened 2018-05-05
Speed optimization for optimal parser

Cyan4973:
Finally found a way to integrate
a new speed optimization for LZ4 optimal parser, named Chain Swap,
into the generic match finder,
thus supporting all search scenarios (extDict and CDict).

Results are pretty good : 
benchmark on `entropy`, `gcc-7`, highest compression mode `-12` :

| Name | -12 (dev) | -12 (this patch) | Improvement |
| --- | --- | --- | --- |
| calgary.tar | 6.2 MB/s | 16.2 MB/s | +160 % |
| enwik7 | 15.6 MB/s | 18.7 MB/s | +20 % |
| silesia.tar | 5.8 MB/s | 13.8 MB/s | +140 % |

Performance improvement vary depending on file, but good news is,
slowest files see the biggest improvements.

Benefits are mostly present at level 12.
Level 11 still see some benefit, but much lower, in the ~20% range.
Level 10 see a strange reversal, with same speed or slightly slower, but better compression ratio.
That's because, instead of reducing the nb of searches, which are now pretty low, the optimization makes each search more useful, but also more complex.

This optimization is not enabled at levels 9 and below,
as the trade-off between better ratio vs slower speed becomes less interesting.

### DIFF
--- a/lib/lz4hc.c
+++ b/lib/lz4hc.c
@@ -382,7 +382,7 @@ int LZ4HC_InsertAndFindBestMatch(LZ4HC_CCtx_internal* const hc4,   /* Index tabl
     /* note : LZ4HC_InsertAndGetWiderMatch() is able to modify the starting position of a match (*startpos),
      * but this won't be the case here, as we define iLowLimit==ip,
      * so LZ4HC_InsertAndGetWiderMatch() won't be allowed to search past ip */
-    return LZ4HC_InsertAndGetWiderMatch(hc4, ip, ip, iLimit, MINMATCH-1, matchpos, &uselessPtr, maxNbAttempts, 0 /* chainSwap */, patternAnalysis, dict, favorCompressionRatio);
+    return LZ4HC_InsertAndGetWiderMatch(hc4, ip, ip, iLimit, MINMATCH-1, matchpos, &uselessPtr, maxNbAttempts, patternAnalysis, 0 /*chainSwap*/, dict, favorCompressionRatio);
 }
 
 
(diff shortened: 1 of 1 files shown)


lib/lz4hc.c
```diff
@@ -241,31 +243,30 @@ LZ4HC_InsertAndGetWiderMatch (
                 matchIndex, lowestMatchIndex);
 
     while ((matchIndex>=lowestMatchIndex) && (nbAttempts)) {
-        DEBUGLOG(7, "remaining attempts : %i", nbAttempts);
+        int mlt=0;
```
terrelln: Now that this variable has a larger scope, can we give it a better name?

lib/lz4hc.c
```diff
@@ -280,34 +281,68 @@ LZ4HC_InsertAndGetWiderMatch (
...
+        {   U32 const distNextMatch = DELTANEXTU16(chainTable, matchIndex);
+            if (patternAnalysis && distNextMatch==1 && matchChainPos==0) {
+                U32 const matchCandidateIdx = matchIndex-1;
                 /* may be a repeated pattern */
                 if (repeat == rep_untested) {
                     if ( ((pattern & 0xFFFF) == (pattern >> 16))
                       &  ((pattern & 0xFF)   == (pattern >> 24)) ) {
                         repeat = rep_confirmed;
-                        srcPatternLength = LZ4HC_countPattern(ip+4, iHighLimit, pattern) + 4;
+                        srcPatternLength = LZ4HC_countPattern(ip+sizeof(pattern), iHighLimit, pattern) + sizeof(pattern);
```
terrelln: nit: consistent spacing around `+`.

PR #540, opened 2018-05-17
Add Haiku as a validated target.

fbrosson:
lz4 1.8.2 works fine on Haiku and passes all tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ clean:
 #-----------------------------------------------------------------------------
 # make install is validated only for Linux, OSX, BSD, Hurd and Solaris targets
 #-----------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


fbrosson:
We have added lz4 to [HaikuPorts](https://github.com/haikuports/haikuports/tree/master/app-arch/lz4) 4 years ago, and I can confirm that release `1.8.2` works like a charm on Haiku :)
So if you could merge this PR it would be cool!
Thanks!

fbrosson:
I did not dare to mention Haiku in the comment (`make install is validated only for Linux, OSX, BSD, Hurd and Solaris targets`) above the changed lines, but I can do so if you wish.
So please do not hesitate to ask. I can change it to e.g. `make install is validated only for Linux, OSX, BSD, Haiku, Hurd and Solaris targets` or any variant of your choice.

Cyan4973:
That's fine @fbrosson , thanks for the patch !
We can always edit the comment later ;)

PR #542, opened 2018-05-28
allow to override uname when cross-compiling

wbx-github:
When cross-compiling for example from Darwin to Linux it might be
useful to override uname output to force Linux and create Linux
libraries instead of Darwin libraries.


Cyan4973:
thanks @wbx-github ! it's a great feature !

PR #543, opened 2018-06-01
Aix 32bit fuzzer

nrgiii:
Fix compile error of fuzzer.c on AIX 32 bit builds.  On AIX 32 bit, sys/mman.h must be included after platform.h else we get errors shown in the attachment, likely caused by bugs in AIX header files.
AIX 64 bit builds also run clean with this change.

[lz4_aix_32bit.txt](https://github.com/lz4/lz4/files/2063868/lz4_aix_32bit.txt)

### DIFF
--- a/tests/fuzzer.c
+++ b/tests/fuzzer.c
@@ -38,7 +38,7 @@
 /*-************************************
 *  Dependencies
 **************************************/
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks @nrgiii !

PR #547, opened 2018-06-26
Add --fast command to cli

jennifermliu:


programs/bench.c
```diff
@@ -407,8 +407,6 @@ static void BMK_benchCLevel(void* srcBuffer, size_t benchedSize,
     if (g_displayLevel == 1 && !g_additionalParam)
         DISPLAY("bench %s %s: input %u bytes, %u seconds, %u KB blocks\n", LZ4_VERSION_STRING, LZ4_GIT_COMMIT_STRING, (U32)benchedSize, g_nbSeconds, (U32)(g_blockSize>>10));
```
Cyan4973: Why is this bound remove ?
Is there some kind of guarantee that `cLevelLast >= level` ? 
In which case, it should be stated with an `assert()`.

Or does it not matter ?
In which case, what happen if a user defines an end level lower than start level ?
such as `zstd -b3e1` for example ?

programs/lz4cli.c
```diff
@@ -363,6 +379,23 @@ int main(int argc, const char** argv)
                 if (!strcmp(argument,  "--help")) { usage_advanced(exeName); goto _cleanup; }
                 if (!strcmp(argument,  "--keep")) { LZ4IO_setRemoveSrcFile(0); continue; }   /* keep source file (default) */
                 if (!strcmp(argument,  "--rm")) { LZ4IO_setRemoveSrcFile(1); continue; }
+                if (longCommandWArg(&argument, "--fast")) {
+                        /* Parse optional window log */
```
terrelln: This isn't a window log, it is the "acceleration factor".

tests/Makefile
```diff
@@ -262,8 +262,13 @@ test-lz4-basic: lz4 datagen unlz4 lz4cat
 	cat tmp-tlb-hw >> tmp-tlb-hw.lz4
 	$(LZ4) -f tmp-tlb-hw.lz4               # uncompress valid frame followed by invalid data
 	$(LZ4) -BX tmp-tlb-hw -c -q | $(LZ4) -tv  # test block checksum
+	test "$(shell ./datagen -g20KB | $(LZ4) -c --fast | wc -c)" -lt "$(shell ./datagen -g20KB | $(LZ4) -c --fast=9 | wc -c)" # compressed size of compression level -1 should be lower than -9
```
terrelln: Lets try this:

```
./datagen -g20KB > tmp-tlb-fast
test "$(shell $(LZ4) -c tmp-tlb-fast --fast | wc -c)" -lt "$(shell $(LZ4) -c tmp-tlb-fast --fast=9 | wc -c)"
```

That way it is clear all the calls are getting the same input, and it is a bit shorter. The `shell` command keeps the current working directory, so it should work (otherwise `$(LZ4)` wouldn't work).

programs/lz4cli.c
```diff
@@ -272,13 +273,28 @@ static unsigned readU32FromChar(const char** stringPtr)
...
+{
+    size_t const comSize = strlen(longCommand);
+    int const result = !strncmp(*stringPtr, longCommand, comSize);
+    if (result) *stringPtr += comSize;
+    return result;
+}
+
 typedef enum { om_auto, om_compress, om_decompress, om_test, om_bench } operationMode_e;
 
+#define CLEAN_RETURN(i) { operationResult = (i); goto _cleanup; }
```
terrelln: You don't need this. Unlike in `zstd`, `badusage()` in LZ4 calls `exit(1)`, so it never returns.

programs/lz4cli.c
```diff
@@ -363,6 +379,23 @@ int main(int argc, const char** argv)
...
+                        if (*argument == '=') {
+                            U32 fastLevel;
+                            ++argument;
+                            fastLevel = readU32FromChar(&argument);
+                            if (fastLevel) {
+                              cLevel = -(int)fastLevel;
+                            }
+                        } else if (*argument != 0) {
+                            /* Invalid character following --fast */
+                            CLEAN_RETURN(badusage(exeName));
```
terrelln: Just `badUsage(exeName)` will suffice.

tests/Makefile
```diff
@@ -262,8 +262,13 @@ test-lz4-basic: lz4 datagen unlz4 lz4cat
...
+	test "$(shell ./datagen -g20KB | $(LZ4) -c --fast=1 | wc -c)" -eq "$(shell ./datagen -g20KB| $(LZ4) -c --fast| wc -c)" # checks default fast compression is -1
+	test "$(shell ./datagen -g20KB | $(LZ4) -c -9 | wc -c)" -lt "$(shell ./datagen -g20KB| $(LZ4) -c --fast=1 | wc -c)" # compressed size of compression level 9 should be lower than -1
```
terrelln: Lets check level `1` vs level `-1` if we can.

programs/lz4cli.c
```diff
@@ -363,6 +379,23 @@ int main(int argc, const char** argv)
...
                 if (!strcmp(argument,  "--rm")) { LZ4IO_setRemoveSrcFile(1); continue; }
+                if (longCommandWArg(&argument, "--fast")) {
+                        /* Parse optional window log */
+                        if (*argument == '=') {
+                            U32 fastLevel;
+                            ++argument;
+                            fastLevel = readU32FromChar(&argument);
+                            if (fastLevel) {
+                              cLevel = -(int)fastLevel;
+                            }
```
terrelln: We should probably fail if `!fastLevel`. Lets also add that patch to `zstd`.

programs/lz4cli.c
jennifermliu: I just updated this comment in lz4. Since I copied this line from zstd, should I change it in zstd as well?

terrelln: Yeah, that'd be great!

programs/lz4cli.c
```diff
@@ -380,17 +378,20 @@ int main(int argc, const char** argv)
...
-                        /* Parse optional window log */
+                        /* Parse optional acceleration factor */
                         if (*argument == '=') {
                             U32 fastLevel;
                             ++argument;
                             fastLevel = readU32FromChar(&argument);
                             if (fastLevel) {
                               cLevel = -(int)fastLevel;
                             }
+                            else {
```
terrelln: nit: We generally put the else on the same line as the previous closing brace `} else {`

tests/Makefile
```diff
@@ -262,9 +262,14 @@ test-lz4-basic: lz4 datagen unlz4 lz4cat
...
+	# ./datagen -g20KB generates the same file every single time
+	# cannot save output of ./datagen -g20KB as input file to lz4 because the following shell commands are run before ./datagen -g20KB
+	test "$(shell ./datagen -g20KB | $(LZ4) -c --fast | wc -c)" -lt "$(shell ./datagen -g20KB | $(LZ4) -c --fast=9 | wc -c)" # -1 vs -9
 	test "$(shell ./datagen -g20KB | $(LZ4) -c --fast=1 | wc -c)" -eq "$(shell ./datagen -g20KB| $(LZ4) -c --fast| wc -c)" # checks default fast compression is -1
-	test "$(shell ./datagen -g20KB | $(LZ4) -c -9 | wc -c)" -lt "$(shell ./datagen -g20KB| $(LZ4) -c --fast=1 | wc -c)" # compressed size of compression level 9 should be lower than -1
+	test "$(shell ./datagen -g20KB | $(LZ4) -c -3 | wc -c)" -lt "$(shell ./datagen -g20KB| $(LZ4) -c --fast=1 | wc -c)" # 3 vs -1
+	test "$(shell ./datagen -g20KB | $(LZ4) -c -1 | wc -c)" -lt "$(shell ./datagen -g20KB| $(LZ4) -c --fast=2 | wc -c)" # 1 vs -2
+	test "$(shell $(LZ4) -c --fast=0 tmp-tlb-dg20k | wc -c)" -eq 0 # lz4 should fail when fast=0
```
terrelln: This should work:

```sh
! $(LZ4) -c --fast=0 tmp-tlb-dg20K1
```

The `!` tells the shell to exit zero if the command exits non-zero, and non-zero when it exits zero.

terrelln:
Looks like level `-1` is the same as level `1` https://github.com/lz4/lz4/blob/dev/lib/lz4frame.c#L752, the test you added caught a bug!

When the level is negative, we should offset the level by one, so -1 is acceleration 2. Then your test for level 1 vs level -1 should pass.

terrelln:
Looks like something isn't working with the benchmarking. It is getting the negative level, but it isn't translating it into an acceleration value, since -2 has the same compression ratio as 1.

```
> ./lz4 --fast=2 -be1 ~/bench/silesia.tar
Benchmarking levels from -2 to 1
-2#silesia.tar       : 211988480 -> 100887443 (2.101), 514.8 MB/s ,2750.1 MB/s
-1#silesia.tar       : 211988480 -> 100887443 (2.101), 514.9 MB/s ,2667.4 MB/s
 0#silesia.tar       : 211988480 -> 100887443 (2.101), 511.9 MB/s ,2749.7 MB/s
 1#silesia.tar       : 211988480 -> 100887443 (2.101), 507.0 MB/s ,2675.2 MB/s
```

PR #548, opened 2018-07-02
Fix bug in frame decompress example

cd127:
The decompression was failing as the srcEnd pointer in
decompress_file_internal was wrongly computed beyond
the end of the memory block.
We need to account for the fact that the header ("info")
was already read in the calling function ("alreadyConsumed").


Cyan4973:
Thanks @cd127 , I completely missed this PR,
and you are right, your correction is spot on.

PR #553, opened 2018-08-01
Add CMake option to not build legacy lz4c program

mrjoel:
Provide a CMake option to not build legacy command line support in lz4c.

CMake FYI: @nemequ


mrjoel:
Yep, defaulting to current behavior is certainly better. Commit updated to default legacy binary build to ON.

PR #556, opened 2018-08-13
Add missing $(EXT)

kou:


Cyan4973:
The patch adds the `.exe` file extension, which is specific to Windows for executable binaries,
onto several targets, many of which are symbolic links, or artifact of the `make install` command.

 But : 
- symbolic links to not exist on Windows
- `make install` is reserved to posix platforms, which excludes Windows platform.

Is there a scenario where this change is impactful ?

Cyan4973:
Sure @kou ! Clear enough !

kou:
Thanks!

PR #575, opened 2018-09-13
unpublish static-only functions

Cyan4973:
Functions declared behind protection macro `LZ4_STATIC_LINKING_ONLY`
are now _not_ published in dll by default.
This can be reversed via opt-in build macro `LZ4_PUBLISH_STATIC_FUNCTIONS`,
mirroring what's currently happening in `lz4frame.h`.

Used this opportunity to update a bunch of api comments within `lz4.h`.


Cyan4973:
cc @felixhandte , as Managed Compression is likely the most impacted by the proposed change.

PR #577, opened 2018-09-18
Add lastError field into LZ4_stream_t_internal

khabinov:
This change adds a new field called `lastError` into `LZ4_stream_t_internal` structure indicating last error that happened during stream compression.

The field can be used by API users (indirectly via auxiliary functions) to get more information about cause of compression failure.

The field can also be used by `_fastReset_` family of functions to check if full reset is necessary.


tests/fuzzer.c
```diff
@@ -676,6 +676,7 @@ static int FUZ_test(U32 seed, U32 nbCycles, const U32 startCycle, const double c
         LZ4_loadDict(&LZ4dict, dict, dictSize);
         ret = LZ4_compress_fast_continue(&LZ4dict, block, compressedBuffer, blockSize, blockContinueCompressedSize-1, 1);
         FUZ_CHECKTEST(ret>0, "LZ4_compress_fast_continue using ExtDict should fail : one missing byte for output buffer : %i written, %i buffer", ret, blockContinueCompressedSize);
+        FUZ_CHECKTEST(LZ4dict.internal_donotuse.lastError!=stream_buffer_overflow, "Should be buffer overflow");
```
khabinov: Here we are getting better granularity for error conditions.

PR #580, opened 2018-09-18
Rename initCheck to dirtyContext and use it in LZ4_resetStream_fast()

khabinov:
Rename `initCheck` to `dirtyContext` and use it in `LZ4_resetStream_fast()` to check if full reset is needed.


lib/lz4.c
```diff
@@ -1003,7 +1019,9 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
         *inputConsumed = (int) (((const char*)ip)-source);
     }
     DEBUGLOG(5, "LZ4_compress_generic: compressed %i bytes into %i bytes", inputSize, (int)(((char*)op) - dest));
-    return (int)(((char*)op) - dest);
+    result = (int)(((char*)op) - dest);
+    if (result <= 0) cctx->needsFullReset = 1;
```
felixhandte: Is this actually needed? I believe at this point we are guaranteed success.

Cyan4973: Right, better to `assert(result>0)` instead.

Cyan4973:
I agree that the actions setting `needsFullReset` are too dispersed, 
it becomes too easy to miss that condition in a future evolution (think that it can be performed by another programmer).

2 main solutions : 
1) nest that function, the upper one being solely in charge to set `needsFullReset` when return value is `0`
2) As @felixhandte suggested, use a final `goto` and perform set action there. This is a close relative, almost equivalent. The only downside is that you have to force the flow through controlled exit points : any future developper will have to use the same `goto` and avoid creating its own early exits, so this should be well documented.

I'm generally in favor of function nesting, but there are exceptions too. Sometimes, the nb of parameters to pass around is too large and becomes inconvenient. And one should ensure that `inlining` properties are well preserved, since performance highly depends on that (it's a `C` limited equivalent for "template programming"). Hence `goto` might end up being simpler to implement.

lib/lz4.c
khabinov: Done, thanks!

khabinov:
Thank you @felixhandte and @Cyan4973. I really appreciate your comments.

Cyan4973:
Does that open the opportunity to merge `LZ4_resetStream_fast()` into `LZ4_resetStream()`, (and presumably `LZ4_compress_fast_extState_fastReset ()`) or should it be a separate PR ?

felixhandte:
@khabinov, actually, the check and fallback to full reset should happen in `LZ4_prepareTable()`, since we also call that directly internally.

khabinov:
@felixhandte @Cyan4973 I addressed https://github.com/lz4/lz4/pull/580#issuecomment-422562308 and https://github.com/lz4/lz4/pull/580#issuecomment-422561244. Not sure if anything has to be done for https://github.com/lz4/lz4/pull/580#issuecomment-422560441.

Also, my English is not perfect, so, I would appreciate if you look carefully into doc changes.

P.S.: there seems to be an html file with documentation. Should it be updated as well?

Thanks!

Cyan4973:
@khabinov : don't worry about the `html` documentation, it's automatically generated from the source file.

lib/lz4.h
```diff
@@ -404,35 +404,37 @@ LZ4LIB_API int LZ4_decompress_fast_usingDict (const char* src, char* dst, int or
...
+ *  Otherwise, it will fall back to the full, expensive reset.
  *
  *  LZ4_streams are guaranteed to be in a valid state when:
  *  - returned from LZ4_createStream()
  *  - reset by LZ4_resetStream()
  *  - memset(stream, 0, sizeof(LZ4_stream_t)), though this is discouraged
  *  - the stream was in a valid state and was reset by LZ4_resetStream_fast()
  *  - the stream was in a valid state and was then used in any compression call
- *    that returned success
+ *    and was then reset by LZ4_resetStream_fast()
```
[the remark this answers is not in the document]
khabinov: I updated the description to make it clear that if `LZ4_compress_fast_continue()` has failed, then the stream can still be passed to `LZ4_resetStream_fast()`, though one has to keep in mind that the history is not preserved.

khabinov:
> So let's call LZ4_resetStream_fast() in LZ4_attach_dictionary() to mitigate that issue.

Done, also added a clarifying comment on that.

Cyan4973:
Looks good to me

nh2:
Is this change supposed to be changing performance?

When bisecting some recent performance improvements (~15% for the use of the advanced frame API) between `v1.8.3` and `dev` (at commit c910db3a), I found that this commit causes them.

Obviously performance improvements are nice, but I was surprised because I didn't expect this change to introduce them based on the commit message.

For some details: A difference seems to be that this condition (from the commit before this one; I added a `printf` there in both commits for my investigation)

```diff
diff --cc lib/lz4.c
index d0fa3e8,d0fa3e8..0ec612e
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@@ -985,7 -985,7 +987,8 @@@ _last_literals
                  lastRun -= (lastRun+240)/255;
              }
              if (outputLimited == limitedOutput)
--                return 0;
++                {{fprintf(stderr, "failure 6 return 0 -- outputLimited == limitedOutput\n"); return 0;}
++return 0;}
          }
          if (lastRun >= RUN_MASK) {
              size_t accumulator = lastRun - RUN_MASK;
```

triggers; before the commit, after it triggers the first time, it triggers on almost every input I feed it, while after the commit it triggers only once overall.

Again, this might be expected, but lz4 working as intended is important to me so I feel I better ask this.

felixhandte:
@nh2, hello!

Can you share more about the sequence of calls you're making on the `LZ4_stream_t`?

This commit in particular makes two substantive changes:

- A context that has not been correctly initialized, or that may be in an inconsistent state, as indicated by `initCheck` / `dirtyContext` != 0, will be fully reset (i.e., `LZ4_resetStream(stream)` aka `memset(stream, 0, sizeof(LZ4_stream_t))`).
- In this commit specifically, calls to `LZ4_compress*` that fail certain preconditions will cause LZ4 to unnecessarily mark the context dirty (which will cause it to be cleared on the next compression call). This change has since been reverted (in #588).

Both of these changes are triggered when LZ4 encounters an error, which should hopefully be an infrequent event (not common enough to cause a 15% change in performance). And neither of them directly interact with the check you call out, which checks whether the very last part of the provided input will fit in the provided output buffer.

So more information would be helpful in understanding what's going on. In particular:

1. How are you preparing the context?
2. What compression call(s) are you making?
3. How are you handling errors?

nh2:
@felixhandte Hmm, then it seems my worries might be justified.

I'm writing a binding to lz4frame in a high-level language (Haskell in my case) and I'm checking the return value of every lz4 call I made and on error throw an exception that immediately terminates the process. So it seems my `LZ4_stream_t` should never get into a situation where it has to reset after an error.

> Can you share more about the sequence of calls you're making on the LZ4_stream_t?

* I am creating the context with `LZ4F_createCompressionContext()` [here](https://github.com/nh2/lz4-frame-conduit/blob/9735e6a7409aaacc2e6486fad6ecb442891688a2/src/Codec/Compression/LZ4/Conduit.hsc#L567)
* Computing the size of my output buffers using `LZ4F_compressBound()` [here](https://github.com/nh2/lz4-frame-conduit/blob/9735e6a7409aaacc2e6486fad6ecb442891688a2/src/Codec/Compression/LZ4/Conduit.hsc#L578)
* Call `LZ4F_compressBegin()` [here](https://github.com/nh2/lz4-frame-conduit/blob/9735e6a7409aaacc2e6486fad6ecb442891688a2/src/Codec/Compression/LZ4/Conduit.hsc#L590)
* in a loop
  * call `LZ4F_compressUpdate()` [here](https://github.com/nh2/lz4-frame-conduit/blob/9735e6a7409aaacc2e6486fad6ecb442891688a2/src/Codec/Compression/LZ4/Conduit.hsc#L624)
* when all input is consumed, call `LZ4F_compressEnd()` [here](https://github.com/nh2/lz4-frame-conduit/blob/9735e6a7409aaacc2e6486fad6ecb442891688a2/src/Codec/Compression/LZ4/Conduit.hsc#L594)

The `printf` I inserted above triggered in one of the `LZ4F_compressUpdate()` calls (like the 6th or 7th of many hundreds). At no time did I receive any error from any of the LZ4F API functions (unless I made a mistake).

If this is unexpected I may try to make a reproduction of it by logging all calls and return values from the API in addition to the inserted printf, if this would be helpful.

PR #581, opened 2018-09-18
Added cppcheck static analyzer test

Cyan4973:
to travisCI,

and fixed a bunch of associated warnings.


Cyan4973:
The fact that different versions of `cppcheck` do not produce the same list of warnings is a worrying sign.
Let's try it for a while, but if it gets too far into noise territory, we'll disable it.

PR #583, opened 2018-09-21
Adding information about dirty context for _HC_ family of functions

khabinov:
With this change an HC compression function in case of failure will update the stream's state to have "dirty context". It is safe now to pass such a stream into `LZ4_resetStreamHC_fast()` function, because it will detect "dirty context" internally and fall back to full reset then.

What was changed:
* renamed `favorDecSpeed` field to `flags`, first 11 bits are not used currently, 12th bit is used to indicate if context is dirty, last 4 bits are reserved for favor kinds;
* updated `LZ4HC_compress_generic_internal()` function to set "dirty context" flag (`lz4hc.c` has complex hierarchy of function calls, therefore refer to the picture);
* updated `LZ4_resetStreamHC_fast()` to detect "dirty context" case;
* updated the tests.

### DIFF
--- a/lib/lz4hc.c
+++ b/lib/lz4hc.c
@@ -744,16 +744,22 @@ LZ4_FORCE_INLINE int LZ4HC_compress_generic_internal (
     cLevel = MIN(LZ4HC_CLEVEL_MAX, cLevel);
     {   cParams_t const cParam = clTable[cLevel];
         HCfavor_e const favor = ctx->favorDecSpeed ? favorDecompressionSpeed : favorCompressionRatio;
-        if (cParam.strat == lz4hc)
-            return LZ4HC_compress_hashChain(ctx,
+        int result;
+
+        if (cParam.strat == lz4hc) {
+            result = LZ4HC_compress_hashChain(ctx,
                                 src, dst, srcSizePtr, dstCapacity,
                                 cParam.nbSearches, limit, dict);
-        assert(cParam.strat == lz4opt);
-        return LZ4HC_compress_optimal(ctx,
-                            src, dst, srcSizePtr, dstCapacity,
-                            cParam.nbSearches, cParam.targetLength, limit,
-                            cLevel == LZ4HC_CLEVEL_MAX,   /* ultra mode */
-                            dict, favor);
+        } else {
+            assert(cParam.strat == lz4opt);
+            result = LZ4HC_compress_optimal(ctx,
+                                src, dst, srcSizePtr, dstCapacity,
+                                cParam.nbSearches, cParam.targetLength, limit,
+                                cLevel == LZ4HC_CLEVEL_MAX,   /* ultra mode */
+                                dict, favor);
+        }
+        if (result <= 0) ctx->dirty = 1;
+        return result;
     }
 }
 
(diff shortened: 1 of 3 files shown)


lib/lz4hc.h
```diff
@@ -161,7 +161,13 @@ struct LZ4HC_CCtx_internal
     uint32_t   lowLimit;        /* below that point, no more dict */
     uint32_t   nextToUpdate;    /* index from which to continue dictionary update */
     short      compressionLevel;
-    short      favorDecSpeed;
+    short      flags;           /* - first 11 bits are not used currently;
```
felixhandte: I think I'd rather put off the cost of doing bitflag stuff as long as possible. Can we just break
```
short favorDecSpeed;
```
into
```
int8_t favorDecSpeed;
int8_t dirty;
```
until we need to pack more in?

lib/lz4hc.h
```diff
@@ -314,6 +326,12 @@ void LZ4_favorDecompressionSpeed(LZ4_streamHC_t* LZ4_streamHCPtr, int favor);
  *  - the stream was in an indeterminate state and was used in a compression
  *    call that fully reset the state (LZ4_compress_HC_extStateHC()) and that
  *    returned success
+ *
+ *  Note:
```
felixhandte: This is kind of misleading. Passing a context into any `_resetStream` function should always clear the history.

lib/lz4hc.h
```diff
@@ -357,6 +375,8 @@ int LZ4_compress_HC_extStateHC_fastReset (void* state, const char* src, char* ds
  */
 LZ4LIB_API void LZ4_attach_HC_dictionary(LZ4_streamHC_t *working_stream, const LZ4_streamHC_t *dictionary_stream);
 
+LZ4LIB_API int LZ4HC_isDirtyContext(const LZ4_streamHC_t *stream);
```
felixhandte: What's the value in exposing this to users? As I see it, a primary goal of this project is to hide complexity from the user, so they don't need to care what state the context is in.

lib/lz4hc.c
```diff
@@ -736,24 +744,37 @@ LZ4_FORCE_INLINE int LZ4HC_compress_generic_internal (
 
     DEBUGLOG(4, "LZ4HC_compress_generic(%p, %p, %d)", ctx, src, *srcSizePtr);
 
-    if (limit == limitedDestSize && dstCapacity < 1) return 0;         /* Impossible to store anything */
-    if ((U32)*srcSizePtr > (U32)LZ4_MAX_INPUT_SIZE) return 0;          /* Unsupported input size (too large or negative) */
+    if (limit == limitedDestSize && dstCapacity < 1) {
```
felixhandte: I know these two conditions produce errors, since we can't proceed. However, I believe the context is left in a clean state, since we haven't in fact performed any writes into the context yet. So I believe it's safe to not mark the context dirty here?

lib/lz4hc.h
khabinov: Yeah, I really like this suggestion. Thanks!

lib/lz4hc.h
khabinov: Updated.

lib/lz4hc.h
khabinov: Agree, it was used in unit tests only. Now it should be removed.

lib/lz4hc.c
khabinov: This is a very good point which leads to a question whether here https://github.com/lz4/lz4/pull/580/files#diff-a4dc0fa3ccaea8a4219c81928cba271aR708 I have to revert my changes as well since no action was taken?

khabinov:
@felixhandte Thanks for taking a look! I've replied to your comments.

lib/lz4hc.c
felixhandte: Ah yes, I believe it should be safe to skip marking the context dirty there as well. Maybe with a note as to why they're not `goto _failure;`ing?

lib/lz4hc.h
felixhandte: I think this comment overall should be cleaned up a little. Like with the LZ4 fast changes, we no longer need to distinguish between successful returns and errors: the context is in a valid state either way. So the wording above can be slightly simplified.

I also wouldn't call out `LZ4_compress_HC_continue()` specifically below, since the statement applies to all APIs. I might word the note as:

>  A stream that was last used in a compression call that returned an error may be passed to this function. However, it will be fully reset, which will clear any existing history and settings from the context.

khabinov:
Thank you @felixhandte! I've updated the comment.

Cyan4973:
Many thanks for the great set of patches @khabinov !

PR #586, opened 2018-09-28
support custom block sizes

blezsan:
Make the -B option support custom block sizes for generic lz4 use rather than just benchmark mode.
Add a new "checkFrame" test tool to verify that all the blocks in a given .lz4 file are of a certain size, and a script to run a number of test cases on this functionality.

A couple of caveats regarding "checkFrame":
- for lower custom block sizes (<4096), the actual block sizes used are occasionally off by +/- 4 bytes. This doesn't appear to be a bug in checkFrame. For now I relaxed the test condition in checkFrame, let me know if I should investigate further.
- when reaching the end of a frame, the LZ4F_decompress() function stops before processing the 4-byte frame checksum, and returns 4 so that it can be done in a subsequent call. This is a special case for checkFrame as the block that was just processed could be smaller than the expected blocksize, since it's the last block in the frame. I couldn't find an easy way to check for this condition, so I'm just skipping the blocksize check whenever LZ4F_decompress() returns 4. Better approaches welcome...

Finally I assumed that the test script would be from the "tests" directory - if not I'll need to tweak the shell variables at the top of the file.


programs/lz4cli.c
```diff
@@ -134,7 +134,8 @@ static int usage_advanced(const char* exeName)
     DISPLAY( " -r     : operate recursively on directories (sets also -m) \n");
 #endif
     DISPLAY( " -l     : compress using Legacy format (Linux kernel compression)\n");
-    DISPLAY( " -B#    : Block size [4-7] (default : 7) \n");
+    DISPLAY( " -B#    : cut file into independent blocks of size # bytes [32+] \n");
```
Cyan4973: It's not necessity for these blocks to be "independent".
Block dependency is controlled by another command, `-BD`.
It should be possible to mix it with custom block sizes.

tests/checkFrame.c
```diff
@@ -0,0 +1,269 @@
...
+*********************************************************/
+#define MIN(a,b)  ( (a) < (b) ? (a) : (b) )
+#define MAX(a,b)  ( (a) > (b) ? (a) : (b) )
+
+void frameCheck(const char *srcFileName, unsigned bsid, size_t blockSize)
+{
+    FILE *srcFile;
+    LZ4F_decompressionContext_t dctx = NULL;
+    size_t srcBufferSize = 4 MB;
+    void *srcBuffer = malloc(srcBufferSize);
```
Cyan4973: coding style :
We like to `const` anytime possible.
This is a useful signal for reviewer / next programmer, who can then concentrate its attention to mutable (nont-`const`) variables.

In this case, buffers could be `void* const`, indicating that they are not going to be re-assigned later in the function.
Same for `srcBufferSize`, which never changes later in the function.

Unfortunately, we haven't yet found a tool to automate such check. So it's entirely reviewer - reliant.

programs/lz4cli.c
blezsan: I removed the "independent"; I also did a manual test with "lz4 -BD" to make sure nothing breaks.

tests/checkFrame.c
blezsan: I added const for the buffer sizes, but it doesn't work for the pointers: they're passed to fread() and LZ4F_decompress() where they're not declared as const.

Cyan4973: > it doesn't work for the pointers:

A difficult point : don't confuse `void* const`, which is a fixed pointer to mutable data,
with `const void*`, which is a mutable pointer to read-only data.
The third variable `void const *` is actually the same as `const void*`.

`const void*` is incorrect, indeed : `fread()` will need to write into the buffer, so it's content must be mutable.

But `void* const` should be correct. I don't see any part of the code trying to modify the pointer itself.

Cyan4973:
> for lower custom block sizes (<4096), the actual block sizes used are occasionally off by +/- 4 bytes. 

This is surprising.
I don't see any good reason for this to happen.
It could be a bug in `lz4frame` though.

4 bytes is suspiciously the same as block header size, frame checksum size, and block checksum size.
Could there be a confusion there ?
Also, is it : 
- always by exactly 4 bytes ? could it be 1,2,3 bytes ?
- always positive, always negative ? regularly switched (once positive, once negative) ?
- only when frame checksum is enabled, or irrespective of this condition ?

blezsan:
It's always exactly 4 bytes, and it follows this pattern for successive blocks (here with -B512):
512 / .... / 512 / 516 / 508 / 512 / .... / 512 / 516 / 508 / 512 / ....
i.e. it's always one +4 block immediately followed by one -4 block.
It does not happen with --no-frame-crc, which I find pretty amazing since that should only affect the last 4 bytes of the stream.

Cyan4973:
Thanks for these elements @blezsan .
It looks like a bug on the generation side then,
either `lz4frame` or `lz4io`.
I'll look into it.

tests/checkFrame.c
```diff
@@ -0,0 +1,269 @@
...
+                }
+                curblocksize = 0;
+            }
+        }
+    }
+    /* can be out because readSize == 0, which could be an fread() error */
+    if (ferror(srcFile)) EXM_THROW(67, "Read error");
+
+    if (nextToLoad!=0) EXM_THROW(68, "Unfinished stream");
+
```
Cyan4973: minor issue, since it's just a test program :
`srcFile` should rather be `fclose()` before the end of the function.

tests/checkFrame.c
```diff
@@ -0,0 +1,269 @@
...
+}
+
+int FUZ_usage(const char* programName)
+{
+    DISPLAY( "Usage :\n");
+    DISPLAY( "      %s [args] filename\n", programName);
+    DISPLAY( "\n");
+    DISPLAY( "Arguments :\n");
+    DISPLAY( " -b#    : expected blocksizeID [4-7] (required)\n");
+    DISPLAY( " -B#    : expected blocksize [4-7] (required)\n");
```
Cyan4973: Do you expect a value between [4-7] with this parameter ?

Cyan4973:
OK, I could generate the same issue, witnessing blocks of size 516/508 bytes in a stream of blocks of 512 bytes.
Although, in my case, they do not follow each other immediately.
There can be a distance between them, so the length vary, from 0 to 8 intermediate 512 bytes blocks.

To be investigated...

Cyan4973:
I added traces inside `lz4frame.c` to track block sizes.
The traces always provide me with the expected block size (except for last block, which can be partial).
I've run the test with a number of options and a number of files, it's always flawless.

I'm now wondering if this could be an artifact of the test program.

Cyan4973:
I've added traces in the `lz4frame` decoder,
used by `checkFrame`,
tracking the size of decoded blocks, directly at block level.

It always reports 512 bytes, in all circumstances, save once again the last block.

It restricts the issue to either the test program itself,
or the buffering mechanism inside `LZ4F_decompress()`.

blezsan:
Thanks Yann for the investigation. I wonder if it has something with the hint returned from LZ4F_decompress(), maybe that is off by a few bytes. It's starting to look like I may have to change tack (use lower level routine)... Before I do that I'll spend a bit more time trying to get the LZ4F_decompress() approach to work. I'll also integrate your other comments, thx.

Cyan4973:
I think I got it narrowed to an issue in the buffering mechanism within `LZ4F_decompress()`.
A difficult mixture involving uncompressed blocks and other conditions,
a situation which is more frequently produced when blocks are small and independent.
On to the fix ...

Cyan4973:
Bug found and fixed in https://github.com/lz4/lz4/pull/587 , now merged into `dev`.

It should fix observed `checkFrame` test result.
It was definitely a bug in `LZ4F_decompress()`.

tests/checkFrame.c
blezsan: Indeed, in fact this was flagged by the CI builds. I moved fopen() / fclose() to the main() routine and the frameCheck() function now returns the error value rather than calling exit().

tests/checkFrame.c
blezsan: Oops, copy-paste error, good spot. Fixed.

blezsan:
Many thanks, the fix works great, all tests now pass without the 4-byte fudging.

PR #589, opened 2018-10-03
compressEnd: add bounds checks for frame ends

raggi:
This patch adds bounds checks for two cases caught by usage under ASAN.

### DIFF
--- a/lib/lz4frame.c
+++ b/lib/lz4frame.c
@@ -1,6 +1,7 @@
 /*
 LZ4 auto-framing library
 Copyright (C) 2011-2016, Yann Collet.
+Copyright (C) 2018, Google.
 
 BSD 2-Clause License (http://www.opensource.org/licenses/bsd-license.php)
 
@@ -951,10 +952,16 @@ size_t LZ4F_compressEnd(LZ4F_cctx* cctxPtr, void* dstBuffer, size_t dstMaxSize, 
     if (LZ4F_isError(flushSize)) return flushSize;
     dstPtr += flushSize;
 
+    if ((size_t)((dstPtr+4) - dstStart) > dstMaxSize) {
+        return err0r(LZ4F_ERROR_dstMaxSize_tooSmall);
+    }
     LZ4F_writeLE32(dstPtr, 0);
     dstPtr+=4;   /* endMark */
 
     if (cctxPtr->prefs.frameInfo.contentChecksumFlag == LZ4F_contentChecksumEnabled) {
+        if ((size_t)((dstPtr+4) - dstStart) > dstMaxSize) {
+            return err0r(LZ4F_ERROR_dstMaxSize_tooSmall);
+        }
         U32 const xxh = XXH32_digest(&(cctxPtr->xxh));
         LZ4F_writeLE32(dstPtr, xxh);
         dstPtr+=4;   /* content Checksum */


lib/lz4frame.c
```diff
@@ -951,10 +952,16 @@ size_t LZ4F_compressEnd(LZ4F_cctx* cctxPtr, void* dstBuffer, size_t dstMaxSize,
...
     dstPtr += flushSize;
 
+    if ((size_t)((dstPtr+4) - dstStart) > dstMaxSize) {
+        return err0r(LZ4F_ERROR_dstMaxSize_tooSmall);
+    }
     LZ4F_writeLE32(dstPtr, 0);
     dstPtr+=4;   /* endMark */
 
     if (cctxPtr->prefs.frameInfo.contentChecksumFlag == LZ4F_contentChecksumEnabled) {
+        if ((size_t)((dstPtr+4) - dstStart) > dstMaxSize) {
```
Cyan4973: Just push this branch after the digest, it should pass the checks.

Cyan4973:
I'm currently trying to confirm this bug,
in order to understand how come it was not detected before.

I suspect a first condition for this situation to happen is that ,
when invoking `LZ4F_compressEnd()`, 
the destination buffer does not respect the minimum size condition : 
`dstCapacity >= LZ4F_compressBound(0, prefsPtr)` ?

This is allowed, but not recommended,
because without this minimum size, compression may fail,
and if it does, the compression context cannot "resume" from where it fails
(it can only be reset), so it can be a bad situation to recover from.

More importantly, this scenario is less stressed during tests, since it is (much) less common.
So that could explain why this issue wasn't detected.

raggi:
Thanks for fixing this :)

We had to read through the code a few times to understand how our usage was wrong as well.

In our use, we'd constructed and used a buffer as follows:

* malloc(buf, compressbound(st_size))
* compressBegin(buf)
* while(read) compressUpdate(buf)
* compressEnd(buf)

The problem with the above approach is that compressbound does not account for the header.

As I was reading through this, I also noticed that on the implementation side there is mention of an undefined constant for the header size, and then on the header side it says 15. I'm guessing users are copying this magic constant value into their code (we just have).

It's probably worth doing a pass over the related docs, as it took a while to review where the issue was in the calling code and compare that against expected usage.

Cyan4973:
I guess one issue is that `LZ4F_compressBound()` is only valid for the _next_ call to `LZ4F_compressUpdate()`. 

It doesn't work if input is then provided in multiple segments, because each call to `LZ4F_compressUpdate()` generates at least one block, therefore introducing a small amount of additional data at each block boundary (extreme cases: each byte is one block => 5x expansion).

A valid construction : 

- buf = malloc(compressbound(max_inputSize))
- compressBegin(buf)
- while(read) 
    + compressUpdate(buf)
    + __flush(buf)__
- compressEnd(buf)

Or, if you _must_ produce the output as a single buffer : 

- buf = malloc( compressFrameBound (total_inputSize) )   // just guidance, valid only when using `LZ4F_compressFrame()`
- compressBegin(buf)
- while (read)
    + worstDstSize = compressBound(inputSize)
    + if (currentDstSize + worstDstSize > bufSize)
        * realloc (buf)
    + compressUpdate(buf)
- compressEnd(buf)

The API was rather designed with the "flush" scenario in mind, it's a bit awkward in the second case.

PR #594, opened 2018-10-22
fix 560 bug and update message

### LINKED ISSUE #560: Rare decompression issue in 1.8.2
Hi,

I've run into an issue when I'm getting broken archives with the recent version of lz4. The previous version works well with the same files.

In short, it seems like under certain circumstances it turns consequent zero characters "0" into zero bytes \0 after decompression. That is, the diff shows

< 0.15000000596046448
\> 0.150596046448

lz4 -t tells off "Error 66 : Decompression error : ERROR_contentChecksum_invalid" on these files.

The example file is rather sensitive to make it public, so I'm ready to provide it to the developers upon request. You should be able to reproduce the bug using lz4 -9.

bettermanzzy:
fix #560 bug and update message


Cyan4973:
?

PR #604, opened 2018-11-17
Add DLLTOOL variable so that one can override dlltool binary

vtorri:
This fixes cross compilation on linux for Windows

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -54,6 +54,8 @@ DEBUGFLAGS:= -Wall -Wextra -Wcast-qual -Wcast-align -Wshadow \
 CFLAGS  += $(DEBUGFLAGS) $(MOREFLAGS)
 FLAGS    = $(CPPFLAGS) $(CFLAGS) $(LDFLAGS)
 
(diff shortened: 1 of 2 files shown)


Cyan4973:
Do you want to document this new build macro somewhere in `lib/README.md` ?

vtorri:
that's something to do, indeed. I'll do another pull request for README.md.

Cyan4973:
It seems both the creation of `DLLTOOL` build macro and its documentation would do well in an encompassing PR, since it's the same topic.

Is there a reason you tend to prefer small PR ?

vtorri:
simpler to review and see if there are problems, that's all

Cyan4973:
I agree, small PR are easier to review.
Obviously, it's a balancing act. In this case, I believe the presence of a documentation feels related to the feature itself, and makes more sense bundled together.

vtorri:
that's exactly why i didn't want to write documentation... I will change that and you will still something find to change again and again. I would have prefer you to write this simple doc...

vtorri:
so ? not good enough ?

PR #605, opened 2018-11-20
Allow installation of lz4 for Windows (MSYS2 or when cross-compiling)

vtorri:
- When cross-compiling (64 bits cross compiler) : 

```make install PREFIX=$HOME/lz4_test BUILD_STATIC=no CC=x86_64-w64-mingw32-gcc DLLTOOL=x86_64-w64-mingw32-dlltool OS=Windows_NT```

```find /home/vtorri/lz4_test``` gives : 

/home/vtorri/lz4_test/
/home/vtorri/lz4_test/lib
/home/vtorri/lz4_test/lib/liblz4.lib
/home/vtorri/lz4_test/lib/pkgconfig
/home/vtorri/lz4_test/lib/pkgconfig/liblz4.pc
/home/vtorri/lz4_test/share
/home/vtorri/lz4_test/share/man
/home/vtorri/lz4_test/share/man/man1
/home/vtorri/lz4_test/share/man/man1/lz4c.1
/home/vtorri/lz4_test/share/man/man1/lz4cat.1
/home/vtorri/lz4_test/share/man/man1/lz4.1
/home/vtorri/lz4_test/share/man/man1/unlz4.1
/home/vtorri/lz4_test/bin
/home/vtorri/lz4_test/bin/unlz4.exe
/home/vtorri/lz4_test/bin/liblz4-1.dll
/home/vtorri/lz4_test/bin/lz4.exe
/home/vtorri/lz4_test/bin/lz4c.exe
/home/vtorri/lz4_test/bin/lz4cat.exe
/home/vtorri/lz4_test/include
/home/vtorri/lz4_test/include/lz4.h
/home/vtorri/lz4_test/include/lz4hc.h
/home/vtorri/lz4_test/include/lz4frame.h

- With MSYS2 (here 32 bits cross compiler) : 

```make install PREFIX=$HOME/lz4_test BUILD_STATIC=no CC=x86_64-w64-mingw32-gcc ```

same files are installed

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -204,9 +204,14 @@ endif
 
 uninstall:
 	$(Q)$(RM) $(DESTDIR)$(pkgconfigdir)/liblz4.pc
+ifneq (,$(filter Windows%,$(OS)))
+	$(Q)$(RM) $(DESTDIR)$(bindir)/$(LIBLZ4).dll
+	$(Q)$(RM) $(DESTDIR)$(libdir)/liblz4.lib
+else
 	$(Q)$(RM) $(DESTDIR)$(libdir)/liblz4.$(SHARED_EXT)
 	$(Q)$(RM) $(DESTDIR)$(libdir)/liblz4.$(SHARED_EXT_MAJOR)
 	$(Q)$(RM) $(DESTDIR)$(libdir)/liblz4.$(SHARED_EXT_VER)
+endif
 	$(Q)$(RM) $(DESTDIR)$(libdir)/liblz4.a
 	$(Q)$(RM) $(DESTDIR)$(includedir)/lz4.h
 	$(Q)$(RM) $(DESTDIR)$(includedir)/lz4hc.h
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -184,6 +184,9 @@ ifeq ($(BUILD_STATIC),yes)
 	$(Q)$(INSTALL_DATA) lz4frame_static.h $(DESTDIR)$(includedir)/lz4frame_static.h
 endif
 ifeq ($(BUILD_SHARED),yes)
+# Traditionnally, one installs the DLLs in the bin directory as programs
+# search them first in their directory. This allows to not pollute system
+# directories (like c:/windows/system32), nor modify the PATH variable.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: lib/Makefile [hunk 1]
```diff
@@ -140,6 +144,8 @@ PREFIX      ?= /usr/local
 prefix      ?= $(PREFIX)
 EXEC_PREFIX ?= $(prefix)
 exec_prefix ?= $(EXEC_PREFIX)
+BINDIR      ?= $(exec_prefix)/bin
```
Cyan4973: It seems you need to install libraries into `/bin` rather than `/lib`, presumably due to `PATH` constraints.
OK, fine.
Just add a code comment here to explain why `BINDIR` exists, so that it doesn't get misused or wrongly modified by some future developer.

vtorri: The DLL are indeed usually installed in the bin/ directory. That is where build systems (autotools with libtool, cmake, meson) are installing DLL, contrary to UNIX where shared lib are in lib/
The import lib (libz4.lib) is usually installed in the lib/ directory

Also note the name of the DLL : libz4-1.dll This number is the major version of the library. I also follow what other build systems do.

I will add the comment in lib/Makefile

should I also update README.md, and if yes which one ? top level or lib/README.md ?

Cyan4973: There is very little in term of doc in this case,
because it's not about providing a control that user must be aware of for the functionality to work.
Here, everything is transparent, and the objective is: `MSYS2` is now a valid environment for `make install` target.

In theory, it would be useful to document that `MSYS2` is a new valid environment,
but since we don't have such list within any `README`, there is no place for this statement.

Except maybe the `Makefile` itself (see line 131).

vtorri:
the NEWS file should maybe have an entry about that, though

vtorri:
comment added. I hope the next release of lz4 will be soon out.

Cyan4973:
I generally update the `NEWS` file just before a release.

Note that there is no `lz4` release planned in the near future.
`lz4` has entered "maintenance mode", since no big feature is expected in the near future.
That means releases can be expected on a 6-month basis, though there is considerable variance possible to adapt to specific situations.

PR #606, opened 2018-11-24
lz4_Frame_format.md: Fix typo HexaBytes -> Exabytes

nh2:


Cyan4973:
Good catch ! Thanks for the correction !

PR #611, opened 2018-12-01
Support for Meson build and its CI tests

lzutao:
If you want to move this meson build changesets to under directory `contrib`. Just tell me.
But having meson in source tree help maintaining and using it more easier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -156,8 +156,26 @@ matrix:
       sudo: false
       env: Cmd="make -C tests checkTag && tests/checkTag $TRAVIS_BRANCH " COMPILER=cc
 
+    - dist: xenial
+      sudo: required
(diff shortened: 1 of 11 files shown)


Cyan4973:
Thanks @lzutao ,
the patch looks good. While I'm no expert at meson, it looks clean and seems to build upon your experience on `zstd` repository, it also includes an automated test, which is great.

However, as the solution deploy way more than a single hidden file,
I would indeed prefer the meson build scripts to occupy their own directory, 
with `contrib/meson` being a good choice.

As a rule, I'm trying to keep the root directory as lean as possible.

lzutao:
Thanks for your review. I moved meson bulild to under `contrib/meson`.

PR #613, opened 2018-12-02
meson test: Update timeout and tests that use COPYING

lzutao:
In meson, default timeout when running tests is 30 seconds.


lzutao:
Ok, ready to merge now.

Cyan4973:
Sorry, some tests can have spurious errors sometimes.
In which case, it's necessary to restart them.
A new push is all it takes to trigger new tests.
Hopefully it doesn't happen frequently.

Cyan4973:
Thanks @lzutao ,
looks good to me !
My only comment is that I suspect a `README.md` doc file in `contrib/meson` might be useful for newcomers.

PR #615, opened 2018-12-02
 Windows: create import library with gcc directly, remove now useless def file

vtorri:


Cyan4973:
The first commit listed is `Merge remote-tracking branch 'upstream/master' into dev `.
It could be a source of this conflict.
Our "top level" branch is `dev`, not `master`.
`master` is merely our staging area before reaching "release" status, it's very late compared to `dev`.

PR #616, opened 2018-12-04
Don't Attach Very Small Dictionaries

felixhandte:
Fixes a mismatch in behavior between loading into the context (via `LZ4_loadDict()`) a very small (<= 4 bytes) non-contiguous dictionary, versus attaching it with `LZ4_attach_dictionary()`.

Before this patch, this divergence could be reproduced by running

```
make -C tests fuzzer MOREFLAGS="-m32"
tests/fuzzer -v -s1239 -t3146
```

Making sure these two paths behave exactly identically is an easy way to test the correctness of the attach path, so it's desirable that this remain an unpolluted, high signal test.

This condition mirrors the condition in `LZ4_compress_fast_continue()`. It's implemented here instead to avoid performing the check on every `_continue()` call.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1271,7 +1271,9 @@ void LZ4_attach_dictionary(LZ4_stream_t *working_stream, const LZ4_stream_t *dic
      */
     LZ4_resetStream_fast(working_stream);
 
-    if (dictionary_stream != NULL) {
+    if (dictionary_stream != NULL
+     && dictionary_stream->internal_donotuse.dictSize - 1 >= 4
+        /* intentional underflow */) {
         /* If the current offset is zero, we will never look in the
          * external dictionary context, since there is no value a table
          * entry can take that indicate a miss. In that case, we need


lib/lz4.c
```diff
@@ -1271,7 +1271,9 @@ void LZ4_attach_dictionary(LZ4_stream_t *working_stream, const LZ4_stream_t *dic
      */
     LZ4_resetStream_fast(working_stream);
 
-    if (dictionary_stream != NULL) {
+    if (dictionary_stream != NULL
+     && dictionary_stream->internal_donotuse.dictSize - 1 >= 4
```
Cyan4973: In `LZ4_loadDict()`, 
the condition is : `if (dictSize < (int)HASH_UNIT) return 0;`.

Why is it different here ?

The commit stated goal is "Making sure these two paths behave exactly identically".

felixhandte: Because `LZ4_compress_fast_continue()` tests `if (dictSize-1 < 4) { /* unset dict */ }` [here](https://github.com/lz4/lz4/blob/dev/lib/lz4.c#L1322). When `HASH_UNIT == 5` those tests are equivalent-ish, but when `HASH_UNIT == 4`, they're not.

So maybe we should just change the `_continue()` test to be against `HASH_UNIT`?

Cyan4973: `LZ4_loadDict()` is a better reference for this case (than `LZ4_compress_fast_continue()`).

Both `LZ4_loadDict()` and `LZ4_attach_dictionary()` are about starting a new compression from a dictionary. So they directly relate.

`LZ4_compress_fast_continue()`, in contrast, is about _continuing a stream_, using previously seen data as a starting point. It's used within `LZ4F_compress()` for example, without any dictionary.

Another important difference is that `LZ4_loadDict()` must load some tables, while `LZ4_compress_fast_continue()` just takes whatever is already present in the state and continues from there. So it doesn't have to hash new positions. Hence `HASH_UNIT` is irrelevant. Only the 4-bytes initial comparison must not fail (ensure no out-of-bound read).

minor: note that `HASH_UNIT` can be equal to `4` or `8`, but never `5`.

PR #618, opened 2018-12-11
Add snap packaging support

dosaboy:
This commit adds a snapcraft.yaml file to allow lz4
to be distributed as a snap (see https://snapcraft.io/
for more info on snaps).

Building the Snap
-----------------

To build the snap simply install snapcraft and build the
snap e.g. on Ubuntu by doing:

$ sudo apt install snapcraft
$ snapcraft

This will build a snap that can be published [1] to
https://snapcraft.io/lz4

Installing the Snap
-------------------

Once published and released in the snapstore your snap
can be installed on any platform that supports snaps
by doing:

snap install lz4

[1] https://docs.snapcraft.io/releasing-to-the-snap-store


Cyan4973:
Thanks @dosaboy , it's a great addition.

This patch adds an element at root directory, which is something which is best avoided whenever possible.

Is it possible to move that file into `contrib/snap` instead ?

dosaboy:
Thanks @Cyan4973. Yeah I hear your point about having that in the top-level dir. There is a reason for  this though which is that if you want to be able to build the snap as follows:

git clone <repo>; cd <repo>; snapcraft

then the snap/snapcraft.yaml has to be in the top-level dir otherwise the build tool will not find it. To achieve what you ask I can modify the snapcraft.yaml to go back a directory for the code i.e. set source: ../ which now means that to build you need to do:

git clone <repo>; cd <repo>/contrib; snapcraft

I'll make that change now and leave it up to you which you prefer but it might make things easier in the long run to have it at the top-level dir.

Cyan4973:
Thanks @dosaboy , I believe your proposed solution is good,
I'll be fine with running `snapcraft` from `contrib` directory, at least for the time being.

I also believe a short `README.md`, explaining the basics (nothing more than what you already explained in this thread) would be useful for anyone to understand what's the purpose of `contrib/snap`.

dosaboy:
@Cyan4973 yep good idea i'll add that in.

Cyan4973:
Thanks for this great patch @dosaboy !

PR #620, opened 2018-12-13
Update meson symlink and man1 extension

lzutao:


lzutao:
cc  @Cyan4973

Cyan4973:
Looks good to me !

PR #631, opened 2019-01-06
 lz4hc.c: change (length >> 8) to (length / 255)

qiuyangs:
Every 0xff byte in the compressed block corresponds to a length of 255 (not 256) in the input data. For long repeating sequences, using (length >> 8) may generate bad compressed blocks.

### DIFF
--- a/lib/lz4hc.c
+++ b/lib/lz4hc.c
@@ -437,7 +437,7 @@ LZ4_FORCE_INLINE int LZ4HC_encodeSequence (
 
     /* Encode Literal length */
     length = (size_t)(*ip - *anchor);
-    if ((limit) && ((*op + (length >> 8) + length + (2 + 1 + LASTLITERALS)) > oend)) return 1;   /* Check output limit */
+    if ((limit) && ((*op + (length / 255) + length + (2 + 1 + LASTLITERALS)) > oend)) return 1;   /* Check output limit */
     if (length >= RUN_MASK) {
         size_t len = length - RUN_MASK;
         *token = (RUN_MASK << ML_BITS);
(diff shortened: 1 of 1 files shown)


terrelln:
Nice catch! @qiuyangs would you please add a test case that exposes this bug to [fuzzer.c](https://github.com/lz4/lz4/blob/dev/tests/fuzzer.c#L592) in the limited output section?

qiuyangs:
@terrelln Thanks. To trigger this bug will need a blockSize larger than the current implementation in fuzzer.c allows. Below is a simple C program to show the issue:

`void main(int argc, char *argv[]) {
        int blockSize, srcSize, targetSize;
        char *block, *compressedBuffer;
        void *ctx;
        int cLevel = 1;

        blockSize = srcSize = 1024 * 1024;
        targetSize = 4116;

        block = malloc(blockSize);
        memset(block, 0, blockSize);
        compressedBuffer = malloc(targetSize);

        ctx = LZ4_createStreamHC();
        LZ4_compress_HC_destSize(ctx, block, compressedBuffer, &srcSize, targetSize, cLevel);

        LZ4_freeStreamHC(ctx);
        free(block);
        free(compressedBuffer);
}
`
Running this program will cause Segmentation Fault.

Cyan4973:
It's possible to add a test following exactly this recipe within [`FUZ_unitTests()`](https://github.com/lz4/lz4/blob/dev/tests/fuzzer.c#L972).
It will be useful to observe the issue, validate its fix, and ensure the property remains valid in the future.

I can take care of it if that's too much on @qiuyangs side.

Cyan4973:
Just added the long sequence test in a dedicated feature branch.
The test fails, as expected.
It's expected to be fixed by this patch.

PR #632, opened 2019-01-09
Fix lz4 extension in input filename not causing decompression

### LINKED ISSUE #596: lz4 extension not causing decompression
Hi all,

Per the usage of `lz4` I see: 

```
Usage :
      lz4 [arg] [input] [output]

input   : a filename
          with no FILE, or when FILE is - or stdin, read standard input
Arguments :
 -1     : Fast compression (default) 
 -9     : High compression 
 -d     : decompression (default for .lz4 extension)
 -z     : force compression
 -f     : overwrite output without prompting 
 -h/-H  : display help/long help and exit
```

Now, when I run `lz4 file.lz4 file` the file is compressed again.

However, when I add `-d` (`lz4 -d file.lz4 file`) the file decompresses properly. I interpret the usage as `lz4 options+ input_file output_file` and the information under `:decompression (default for .lz4 extension)` to mean that if the input file ends with `.lz4` then it will add `-d` automatically as part of the default behavior. This is not the case for me.

I'm using the latest lz4 available in `ubuntu/bionic liblz4-tool` (`*** LZ4 command line interface 64-bits r128, by Yann Collet (Feb 17 2016) ***`).

Example:

```
root@x:~# ls
root@x:~# echo "TEST" > test
root@x:~# lz4 test 
Compressed filename will be : test.lz4 
Compressed 5 bytes into 24 bytes ==> 480.00%                                   
root@x:~# lz4 test.lz4 test2
Compressed 24 bytes into 43 bytes ==> 179.17%   
root@x:~# ls
test  test2  test.lz4                               
root@x:~# cat test
TEST
root@x:~# cat test2
"M?d@�?�"M?d@��TEST
��      ��groot@x:~# 
```

The interesting thing is that it compresses again even though it has the `.lz4` extension:

```
root@x:~# lz4 -d test2 test2.out
Successfully decoded 24 bytes                                                  
root@x:~# md5sum * 
2debfdcf79f03e4a65a667d21ef9de14  test
0c7629404e201c0965af2642a1e357f8  test2
99af8b5896cf5fc71ba3c0cad5a370d1  test2.out
99af8b5896cf5fc71ba3c0cad5a370d1  test.lz4
```

(Note that `test.lz4` is now the same file as `test2.out`, so `test2.out` appears to be compressed twice).

I'd argue that the help text should be updated to indicate that the `lz4` extension does not mean that `-d` is default.

Thanks for your time,

Aaron

rubenochiavone:
The issue happens because the default operation mode is compression. Even though input file has .lz4 extension, operation mode is only changed by input file extension when output file is NOT defined - this behavior was also reported in #596.

In order to fix it, I re-use that logic to determine operation mode when no mode is explicitly set.

Closes #596

### DIFF
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -287,6 +287,19 @@ static unsigned longCommandWArg(const char** stringPtr, const char* longCommand)
 
 typedef enum { om_auto, om_compress, om_decompress, om_test, om_bench } operationMode_e;
 
+/** determineOpMode() :
+ *  auto-determine operation mode, based on input filename extension
+ *  @return `om_decompress` if input filename has .lz4 extension and `om_compress` otherwise.
+ */
+static operationMode_e determineOpMode(const char* inputFilename)
+{
+    size_t const inSize  = strlen(inputFilename);
+    size_t const extSize = strlen(LZ4_EXTENSION);
+    size_t const extStart= (inSize > extSize) ? inSize-extSize : 0;
+    if (!strcmp(inputFilename+extStart, LZ4_EXTENSION)) return om_decompress;
+    else return om_compress;
+}
+
 int main(int argc, const char** argv)
 {
     int i,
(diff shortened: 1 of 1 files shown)


rubenochiavone:
Sure.

rubenochiavone:
Done. Please let me know if it is enough

Cyan4973:
Looks great ! Thanks @rubenochiavone !

PR #633, opened 2019-01-09
Make LZ4F_getBlockSize public and public in experimental section

### LINKED ISSUE #608: Export LZ4F_getBlockSize
Right now it seems I cannot obtain the info how many bytes are taken by a given `LZ4F_blockSizeID_t`.

In particular, it's impossible to ask the API how many bytes `LZ4F_default` takes.

It would thus be great to have `LZ4F_getBlockSize` in the public API -- or even better, a compile-time macro for the same information.

tzakian:
Publish the `LZ4_getBlockSize` function to the experimental section of the API. 

This resolves #608

### DIFF
--- a/tests/frametest.c
+++ b/tests/frametest.c
@@ -658,6 +658,28 @@ int basicTests(U32 seed, double compressibility)
         CHECK( LZ4F_freeCompressionContext(cctx) ); cctx = NULL;
     }
 
+    DISPLAYLEVEL(3, "getBlockSize test: \n");
+    { size_t result;
+      for (unsigned blockSizeID = 4; blockSizeID < 8; ++blockSizeID) {
+        result = LZ4F_getBlockSize(blockSizeID);
+        CHECK(result);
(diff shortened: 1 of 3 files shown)


tests/frametest.c
```diff
@@ -658,6 +658,28 @@ int basicTests(U32 seed, double compressibility)
         CHECK( LZ4F_freeCompressionContext(cctx) ); cctx = NULL;
     }
 
+    DISPLAYLEVEL(3, "getBlockSize test: \n");
+    { size_t result;
+      for (unsigned blockSizeID = 4; blockSizeID < 8; ++blockSizeID) {
```
Cyan4973: unfortunately, this is a `C99` construct.
We try to preserve `C90` compatibility in most `lz4` codebase, 
including its tests tools.

You'll have to declare `unsigned blockSizeID;` just before the `for(`.

Cyan4973:
Thanks @tzakian , that's mostly good.
Just a minor `C90` compatibility detail to fix.

PR #637, opened 2019-01-10
Fix pass-through mode

### LINKED ISSUE #601: lz4 -f with bad magic number invokes pass-through in addition to overwriting output file
According to the command line usage:

*** LZ4 command line interface 64-bits v1.8.3, by Yann Collet ***
...
-f     : overwrite output without prompting
...

There is no mention of pass-through mode. Without -f, a bad header is reported:
$ lz4 README_first_256_bad.lz4
Error 44 : Unrecognized header : file cannot be decoded

With -f, no error occurs, and the output file is identical to the input file:
$ lz4 -f README_first_256_bad.lz4
README_first_256_bad : decoded 233 bytes

In the attachment, one bit of the magic number is flipped, first 8 bytes are:
00000000: 04 22 4c 18 64 40 a7 d6
[README_first_256_bad.zip](https://github.com/lz4/lz4/files/2590752/README_first_256_bad.zip)

tzakian:
This PR resolves issues surrounding pass-through mode with lz4. It does so by adding an additional flag to to signify pass-through mode in `LZ4IO`, we then use this when selecting the decoder.

Resolves #601

### DIFF
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -340,6 +340,7 @@ int main(int argc, const char** argv)
     if (exeNameMatch(exeName, LZ4CAT)) {
         mode = om_decompress;
         LZ4IO_setOverwrite(1);
+        LZ4IO_setPassThrough(1);
         LZ4IO_setRemoveSrcFile(0);
         forceStdout=1;
         output_filename=stdoutmark;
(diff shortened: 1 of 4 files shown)


Cyan4973:
Thanks @tzakian !

PR #638, opened 2019-01-10
Travis: Clean up .travis.yml

lzutao:
- Prefer `script` field than `Cmd` 
- Prefer `apt-get` in `install` field than `addons-apt-sources`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,90 +1,116 @@
 language: c
+
 matrix:
   fast_finish: true
   include:
     # OS X Mavericks
-    - os: osx
+    - name: (macOS) General Test
+      os: osx
       install:
         - export CC=clang
-      env: Ubu=OS_X_Mavericks Cmd='make -C tests test-lz4 MOREFLAGS="-Werror -Wconversion -Wno-sign-conversion" && CFLAGS=-m32 make -C tests clean test-lz4-contentSize' COMPILER=clang
+      script:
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks @lzutao , looks good to me.

For my culture,
any specific reason you prefer the `apt-get` method instead of `addons: apt: packages:` ?

lzutao:
The [`installing dependencies`](https://docs.travis-ci.com/user/installing-dependencies/) guide use `apt-get` method to install Ubuntu packages that are not included in the standard precise, trusty, or xenial distribution.,

Cyan4973:
The linked help page mentions both methods. I don't read it as stating a preference of one method over the other one (or I missed something ?).

In a separate repository, I actually tried to measure if there was any difference between the 2 methods.
See this travis test : https://travis-ci.org/facebook/zstd/builds/475163038

Tasks 3 and 4 are identical, they just change in the way they initialize.
Task 4 runs noticeably faster.

This test seems to suggest that `addons: apt: packages:` results in faster run, presumably thanks to faster install.

But it could also be a measurement fluke.

I would recommend a similar methodology here, 
testing both initialization method side by side,
and see if it translates into any runtime difference.

Cyan4973:
Thanks for the excellent reporting @lzutao .
So it seems that speed differences, while measurable, are essentially random,
related for example to differences in underlying host specs, 
or depending on other colocalized VM activity.

Hence no real winner.
In which case, select whichever you prefer.

PR #640, opened 2019-01-10
Remove a bunch of global variables that tracked settings for the IO module

tzakian:
Previously, the various preferences for the IO module were tracked via a number of global variables. This PR, gets rid of the global variables related to the IO and compression, and packages them up in a struct that is passed around throughout the code.

### DIFF
--- a/programs/lz4io.c
+++ b/programs/lz4io.c
@@ -107,20 +107,23 @@ static clock_t g_time = 0;
 /**************************************
 *  Local Parameters
 **************************************/
-static int g_passThrough = 0;
-static int g_overwrite = 1;
-static int g_testMode = 0;
-static int g_blockSizeId = LZ4IO_BLOCKSIZEID_DEFAULT;
-static size_t g_blockSize = 0;
-static int g_blockChecksum = 0;
-static int g_streamChecksum = 1;
-static int g_blockIndependence = 1;
-static int g_sparseFileSupport = 1;
-static int g_contentSizeFlag = 0;
-static int g_useDictionary = 0;
-static unsigned g_favorDecSpeed = 0;
-static const char* g_dictionaryFilename = NULL;
 
+struct LZ4IO_prefs_s {
+  int passThrough;
+  int overwrite;
+  int testMode;
+  int blockSizeId;
+  size_t blockSize;
+  int blockChecksum;
+  int streamChecksum;
+  int blockIndependence;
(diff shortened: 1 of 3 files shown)


programs/lz4io.c
```diff
@@ -152,81 +155,108 @@ static const char* g_dictionaryFilename = NULL;
 /* ****************** Parameters ******************** */
 /* ************************************************** */
 
-int LZ4IO_setDictionaryFilename(const char* dictionaryFilename) {
-    g_dictionaryFilename = dictionaryFilename;
-    g_useDictionary = dictionaryFilename != NULL;
-    return g_useDictionary;
+LZ4IO_prefs_t* LZ4IO_defaultPreferences(void)
+{
+  LZ4IO_prefs_t* ret = malloc(sizeof(LZ4IO_prefs_t));
```
Cyan4973: `malloc()` can fail

programs/lz4io.h
```diff
@@ -48,69 +48,77 @@ static const char nulmark[] = "nul";
...
+LZ4IO_prefs_t* LZ4IO_defaultPreferences(void);
+void LZ4IO_freePreferences(LZ4IO_prefs_t* prefs);
 
 /* ************************************************** */
 /* ****************** Functions ********************* */
 /* ************************************************** */
 
-int LZ4IO_compressFilename  (const char* input_filename, const char* output_filename, int compressionlevel);
-int LZ4IO_decompressFilename(const char* input_filename, const char* output_filename);
+int LZ4IO_compressFilename  (LZ4IO_prefs_t* prefs, const char* input_filename, const char* output_filename, int compressionlevel);
```
Cyan4973: Do you need the `LZ4IO_prefs_t` structure to be writable in these functions ?

tzakian: For all of them, it cannot be a `const LZ4IO_prefs_t*` due to the eventual calls to`LZ4IO_openDstFile` possibly modifying the `sparseFileSupport` field of the struct. Thus this does need to be writable.

That being said though, these can all be made `LZ4IO_prefs_t* const`s since the pointer should never changed (which will be done shortly).

programs/lz4io.c
tzakian: Good catch! Sorry about not catching that earlier.

Only question I have is which error number to throw, but we can sync up on that tomorrow morning :)

Issue #411, opened 2017-10-26
Store forwarding in copying match


odaira:
This section of code in LZ4_decompress_generic suffers from many blocked loads due to failed store forwarding.

        if (unlikely(offset<8)) {
            const int dec64 = dec64table[offset];
            op[0] = match[0];
            op[1] = match[1];
            op[2] = match[2];
            op[3] = match[3];
            match += dec32table[offset];
            memcpy(op+4, match, 4);
            match -= dec64;
        } else { LZ4_copy8(op, match); match+=8; }
        op += 8;

        if (unlikely(cpy>oend-12)) {
            BYTE* const oCopyLimit = oend-(WILDCOPYLENGTH-1);
            if (cpy > oend-LASTLITERALS) goto _output_error;    /* Error : last LASTLITERALS bytes must be literals (uncompressed) */
            if (op < oCopyLimit) {
                LZ4_wildCopy(op, match, oCopyLimit);
                match += oCopyLimit - op;
                op = oCopyLimit;
            }
            while (op<cpy) *op++ = *match++;
        } else {
            LZ4_copy8(op, match);
            if (length>16) LZ4_wildCopy(op+8, match+8, cpy);
        }

I confirmed it using OProfile with`ld_blocks 0x2` on x86 Haswell and `PM_CMPLU_STALL_ST_FWD` on POWER8.

Is the current implementation a result of balancing the code simplicity against the speed, or is it acceptable to further optimize the code?

It is hard to estimate a speed-up by removing the store forwarding, but my guesstimate by accumulating the CPU cycles hit around the load instructions is up to a 20% speed-up in LZ4_decompress_fast on POWER8. Recent x86 CPUs have relatively better store forwarding than POWER, so the benefit might be smaller on x86.

FYI, LZ4_decompress_fast on POWER8 spends 50% of the CPU cycles stalling in the CPU backend.  I guess 20% for the store forwarding and 30% for cache misses.

terrelln:
I'd be very interested if this bit of code can be made faster (especially for x86), so long as it doesn't regress on other architectures. This piece of code is replicated in [zstd](https://github.com/facebook/zstd/blob/dev/lib/decompress/zstd_decompress.c#L1044), so any win here should translate to zstd as well.

As long as there is a measurable improvement, it is worthwhile to add more code. However, we would like to avoid architecture-specific code or assembly. We would accept it for a large enough win, but it would have to be well encapsulated, and have a fallback implementation in C.

[...]

Issue #126, opened 2015-06-26
Idea to improve decompression performance for repeated sequences with short period


mtopolnik:
I am testing the performance of the LZ4 library through the JNI binding provided by jpountz. Currently the Java binding uses r123. I have determined a quite bad degradation of performance when the decompressed data consists of a short byte sequence repeated many times (with a period of 1-32 bytes). In reviewing the memory operations used for decompression I noticed that unaligned memory access is applied, which at least my Intel Core i7 does quite slowly when source and target are close together. As a specific example, for a period length 101 decompression reaches 16 GB/s, but for periods of 3, 5, or 7 the speed drops to 1.8 GB/s. A simple stretch of the same byte value works with 5 GB/s.

To alleviate this issue I have written a wild copying routine which does no reading at all in the copy loop for periods up to 8 and for wider periods uses word-aligned writing. For periods of 3, 5, and 7 I get at least 6 GB/s; for periods of 1, 2, 4, 8 I get ~30 GB/s (vs. 5 for the JNI version). For wider periods, say 13, I get 3.6 GB/s vs. 1.8 GB/s native.

I have opened this issue in case there might be some interest in involving this kind of approach on the native LZ4 project, My code was presented at https://github.com/jpountz/lz4-java/issues/72. It is specific to a little-endian architecture, but is trivial to adapt to big-endian (swapping all `>>>` and `<<` ops).

Cyan4973:
Thanks for the suggestion !
I'll have a look into it.
Rgds

Cyan4973:
I've been looking at your suggested improvement at https://github.com/jpountz/lz4-java/issues/72.
They look promising. But it's not trivial to grasp the entire mechanism in a single read.
Consequently, I've got a few questions.

The first is about measurements. I can't guarantee that Adrien's Java version uses the same tricks as the "C" (native) version, nor that they are translated into the same assembler code by the jvm. Nonetheless, it's the version you compare to. So it becomes a bit more difficult to evaluate the potential improvements from your method.
I see that you have also compared to the "JNI" version, which binds the C code. As you know, there is an additional penalty associated with JNI which becomes more and more prevalent as the sample size of data to compress becomes smaller.

So I would suggest to have some measurement on the "C" native version, if possible, to establish a baseline for future improvements comparisons.

[...]

PR #645, opened 2019-01-29
Optimize decompress generic

djwatson:
The current decompress_generic code suffers from many blocked loads, originally mentioned in issue #411. This patchset attemps to fix that issue.  Also similar issues mentioned in pull #470 , issue #126, and probably others.

The heart of the issue is that LZ4_wildCopy does a copy by 8 bytes at a time, leading to lots of work happening side-by-side.   The farther away we can move these copies by extending the copy width, the faster things seem to go.  Increasing the copy width to as wide as 32 bytes results in overall decompression speed improvement of around ~10% a xeon, ~20% on a modern AMD.

The primary code challenge then is that we have only 8 bytes of WILDCOPYLENGTH, when we'd really like far larger.  So we end up duplicating most of the decode loop, to add a fastpath for a larger copy length, and then the last handful of bytes can use the original loop.  

I've split up the code in what I hope are readable small diffs.  Testing has mostly consisted of 'make test', and an attempt at fuzzing with afl.  Has *not* been tested on other arches, but it would be fairly easy to #ifdef the first loop away and be the same as previous if it shows a regression.

Benchmarks on Intel(R) Xeon(R) CPU E5-2680 v4 @ 2.40GHz, from lzbench.  The final percent is how much faster this patchset is vs. the current dev branch
```
./lzbench -elz4 silesia/*

lz4 1.8.3                 297 MB/s  2598 MB/s     6428742  63.07 silesia/dickens
lz4 1.8.3                 328 MB/s  2452 MB/s     6428742  63.07 silesia/dickens    -5% 

lz4 1.8.3                 472 MB/s  2374 MB/s    26435667  51.61 silesia/mozilla
lz4 1.8.3                 493 MB/s  2658 MB/s    26435667  51.61 silesia/mozilla        12%

lz4 1.8.3                 454 MB/s  2778 MB/s     5440937  54.57 silesia/mr
lz4 1.8.3                 493 MB/s  3037 MB/s     5440937  54.57 silesia/mr             9%

lz4 1.8.3                 796 MB/s  3240 MB/s     5533040  16.49 silesia/nci
lz4 1.8.3                 877 MB/s  4061 MB/s     5533040  16.49 silesia/nci            25%

lz4 1.8.3                 385 MB/s  2383 MB/s     4338918  70.53 silesia/ooffice
lz4 1.8.3                 426 MB/s  2709 MB/s     4338918  70.53 silesia/ooffice        14%

lz4 1.8.3                 413 MB/s  2463 MB/s     5256666  52.12 silesia/osdb
lz4 1.8.3                 439 MB/s  2950 MB/s     5256666  52.12 silesia/osdb           20%

lz4 1.8.3                 303 MB/s  2181 MB/s     3181387  48.00 silesia/reymont
lz4 1.8.3                 331 MB/s  2182 MB/s     3181387  48.00 silesia/reymont        0%

lz4 1.8.3                 511 MB/s  2806 MB/s     7716839  35.72 silesia/samba
lz4 1.8.3                 552 MB/s  3330 MB/s     7716839  35.72 silesia/samba          19%

lz4 1.8.3                 381 MB/s  2899 MB/s     6790273  93.63 silesia/sao
lz4 1.8.3                 412 MB/s  3586 MB/s     6790273  93.63 silesia/sao            24%

lz4 1.8.3                 448 MB/s  2475 MB/s   100881331  47.60 silesia/silesia.tar
lz4 1.8.3                 474 MB/s  2784 MB/s   100881331  47.60 silesia/silesia.tar    12%

lz4 1.8.3                 344 MB/s  2395 MB/s    20139988  48.58 silesia/webster
lz4 1.8.3                 372 MB/s  2536 MB/s    20139988  48.58 silesia/webster        6%

lz4 1.8.3                1104 MB/s  7769 MB/s     8390195  99.01 silesia/x-ray
lz4 1.8.3                1032 MB/s  9056 MB/s     8390195  99.01 silesia/x-ray          17%

lz4 1.8.3                 715 MB/s  2932 MB/s     1227495  22.96 silesia/xml
lz4 1.8.3                 729 MB/s  3514 MB/s     1227495  22.96 silesia/xml            20%
```

An all-zeroes test is particularly ugly for the current code:

```

dd if=/dev/zero of=zerotest bs=1M count=2
(Currently any match extending in to the last 64 bytes uses the slowpath, TODO)
dd if=/dev/random bs=65 count=1 >> zerotest

lz4 1.8.3               15996 MB/s  4920 MB/s        8296   0.40 silesia/zerotest
lz4 1.8.3               16041 MB/s 10151 MB/s        8296   0.40 silesia/zerotest       106%
```

I also tested on a AMD 2400G:

```
dickens          7%
mozilla          12%
mr               18%
nci              29%
ooffice          23%
osdb             28%
reymont          7%
samba            25%
sao              39%
silesia.tar      20%
webster          17%
xml              32%
x-ray            6%
```

The same issue likely exists in zstd (where it is likely easier to fix, since parsing isn't mixed up with sequence execution).

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -297,6 +297,30 @@ void LZ4_wildCopy(void* dstPtr, const void* srcPtr, void* dstEnd)
     do { memcpy(d,s,8); d+=8; s+=8; } while (d<e);
 }
 
+static const unsigned inc32table[8] = {0, 1, 2,  1,  0,  4, 4, 4};
+static const int      dec64table[8] = {0, 0, 0, -1, -4,  1, 2, 3};
+
+
+LZ4_FORCE_O2_INLINE_GCC_PPC64LE
+void LZ4_memcpy_using_offset_base(BYTE* dstPtr, const BYTE* srcPtr, BYTE* dstEnd, const size_t offset) {
+    if (offset < 8) {
+        dstPtr[0] = srcPtr[0];
+        dstPtr[1] = srcPtr[1];
+        dstPtr[2] = srcPtr[2];
+        dstPtr[3] = srcPtr[3];
+        srcPtr += inc32table[offset];
+        memcpy(dstPtr+4, srcPtr, 4);
+        srcPtr -= dec64table[offset];
+        dstPtr += 8;
+    } else {
+        memcpy(dstPtr, srcPtr, 8);
+        dstPtr += 8;
+        srcPtr += 8;
+    }
+
+    LZ4_wildCopy(dstPtr, srcPtr, dstEnd);
+}
+
 /* customized variant of memcpy, which can overwrite up to 32 bytes beyond dstEnd */
 LZ4_FORCE_O2_INLINE_GCC_PPC64LE
 void LZ4_wildCopy32(void* dstPtr, const void* srcPtr, void* dstEnd)
@@ -306,6 +330,40 @@ void LZ4_wildCopy32(void* dstPtr, const void* srcPtr, void* dstEnd)
     BYTE* const e = (BYTE*)dstEnd;
 
     do { memcpy(d,s,16); memcpy(d+16,s+16,16); d+=32; s+=32; } while (d<e);
+}
+LZ4_FORCE_O2_INLINE_GCC_PPC64LE
+void LZ4_memcpy_using_offset(BYTE* dstPtr, const BYTE* srcPtr, BYTE* dstEnd, const size_t offset) {
+    BYTE v[8];
+    switch(offset) {
+    case 1: 
+        memset(v, *srcPtr, 8);
+        goto copy_loop;
+    case 2: 
+        memcpy(v, srcPtr, 2);
+        memcpy(&v[2], srcPtr, 2);
+        memcpy(&v[4], &v[0], 4);
+        goto copy_loop;
+    case 4: 
+        memcpy(v, srcPtr, 4);
+        memcpy(&v[4], srcPtr, 4);
+        goto copy_loop;
+    case 3:
+    case 5:
+    case 6:
+    case 7:
+    case 8:
+    default: 
+        LZ4_memcpy_using_offset_base(dstPtr, srcPtr, dstEnd, offset);
+        return;
+    }
+
+ copy_loop:
+    memcpy(dstPtr, v, 8);
+    dstPtr += 8;
+    while (dstPtr < dstEnd) {
+        memcpy(dstPtr, v, 8);
+        dstPtr += 8;
+    }
 }
 
 /*-************************************
@@ -1505,8 +1563,6 @@ LZ4_decompress_generic(
         BYTE* cpy;
 
         const BYTE* const dictEnd = (dictStart == NULL) ? NULL : dictStart + dictSize;
-        const unsigned inc32table[8] = {0, 1, 2,  1,  0,  4, 4, 4};
-        const int      dec64table[8] = {0, 0, 0, -1, -4,  1, 2, 3};
 
         const int safeDecode = (endOnInput==endOnInputSize);
         const int checkOffset = ((safeDecode) && (dictSize < (int)(64 KB)));
@@ -1643,23 +1699,7 @@ LZ4_decompress_generic(
             /* partialDecoding : may not respect endBlock parsing restrictions */
             assert(op<=oend);
             if (unlikely(offset<16)) {
-                if (offset < 8) {
-                    op[0] = match[0];
-                    op[1] = match[1];
-                    op[2] = match[2];
-                    op[3] = match[3];
-                    match += inc32table[offset];
-                    memcpy(op+4, match, 4);
-                    match -= dec64table[offset];
-                    op += 8;
-                } else {
-                    memcpy(op, match, 8);
-                    op += 8;
-                    match += 8;
-                }
-
-                memcpy(op, match, 8);
-                if (length > 16) LZ4_wildCopy(op+8, match+8, cpy);
+                LZ4_memcpy_using_offset(op, match, cpy, offset);
             } else {
                 LZ4_wildCopy32(op, match, cpy);
             }
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -297,6 +297,16 @@ void LZ4_wildCopy(void* dstPtr, const void* srcPtr, void* dstEnd)
     do { memcpy(d,s,8); d+=8; s+=8; } while (d<e);
 }
 
+/* customized variant of memcpy, which can overwrite up to 16 bytes beyond dstEnd */
+LZ4_FORCE_O2_INLINE_GCC_PPC64LE
+void LZ4_wildCopy16(void* dstPtr, const void* srcPtr, void* dstEnd)
+{
+    BYTE* d = (BYTE*)dstPtr;
+    const BYTE* s = (const BYTE*)srcPtr;
+    BYTE* const e = (BYTE*)dstEnd;
+
+    do { memcpy(d,s,16); d+=16; s+=16; } while (d<e);
+}
 
 /*-************************************
 *  Common Constants
@@ -1627,33 +1637,28 @@ LZ4_decompress_generic(
                 continue;
             }
 
-            if (unlikely(offset<8)) {
-                op[0] = match[0];
-                op[1] = match[1];
-                op[2] = match[2];
-                op[3] = match[3];
-                match += inc32table[offset];
-                memcpy(op+4, match, 4);
-                match -= dec64table[offset];
-            } else {
-                memcpy(op, match, 8);
-                match += 8;
-            }
-            op += 8;
-
-            if (unlikely(cpy > oend-MATCH_SAFEGUARD_DISTANCE)) {
-                BYTE* const oCopyLimit = oend - (WILDCOPYLENGTH-1);
-                if (cpy > oend-LASTLITERALS) goto _output_error;    /* Error : last LASTLITERALS bytes must be literals (uncompressed) */
-                if (op < oCopyLimit) {
-                    LZ4_wildCopy(op, match, oCopyLimit);
-                    match += oCopyLimit - op;
-                    op = oCopyLimit;
+            if (unlikely(offset<16)) {
+                if (offset < 8) {
+                    op[0] = match[0];
+                    op[1] = match[1];
+                    op[2] = match[2];
+                    op[3] = match[3];
+                    match += inc32table[offset];
+                    memcpy(op+4, match, 4);
+                    match -= dec64table[offset];
+                    op += 8;
+                } else {
+                    memcpy(op, match, 8);
+                    op += 8;
+                    match += 8;
                 }
-                while (op < cpy) *op++ = *match++;
-            } else {
+
                 memcpy(op, match, 8);
                 if (length > 16) LZ4_wildCopy(op+8, match+8, cpy);
+            } else {
+                LZ4_wildCopy16(op, match, cpy);
             }
+            
             op = cpy;   /* wildcopy correction */
         }
 
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1632,20 +1632,6 @@ LZ4_decompress_generic(
 
             /* partialDecoding : may not respect endBlock parsing restrictions */
             assert(op<=oend);
-            if (partialDecoding && (cpy > oend-MATCH_SAFEGUARD_DISTANCE)) {
-                size_t const mlen = MIN(length, (size_t)(oend-op));
-                const BYTE* const matchEnd = match + mlen;
-                BYTE* const copyEnd = op + mlen;
-                if (matchEnd > op) {   /* overlap copy */
-                    while (op < copyEnd) *op++ = *match++;
-                } else {
-                    memcpy(op, match, mlen);
-                }
-                op = copyEnd;
-                if (op==oend) goto decode_done;
-                continue;
-            }
-
             if (unlikely(offset<16)) {
                 if (offset < 8) {
                     op[0] = match[0];
@@ -1858,7 +1844,6 @@ LZ4_decompress_generic(
             }
             op = cpy;   /* wildcopy correction */
         }
-    decode_done:
 
         /* end of decoding */
         if (endOnInput)
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -297,15 +297,15 @@ void LZ4_wildCopy(void* dstPtr, const void* srcPtr, void* dstEnd)
     do { memcpy(d,s,8); d+=8; s+=8; } while (d<e);
 }
 
-/* customized variant of memcpy, which can overwrite up to 16 bytes beyond dstEnd */
+/* customized variant of memcpy, which can overwrite up to 32 bytes beyond dstEnd */
 LZ4_FORCE_O2_INLINE_GCC_PPC64LE
-void LZ4_wildCopy16(void* dstPtr, const void* srcPtr, void* dstEnd)
+void LZ4_wildCopy32(void* dstPtr, const void* srcPtr, void* dstEnd)
 {
     BYTE* d = (BYTE*)dstPtr;
     const BYTE* s = (const BYTE*)srcPtr;
     BYTE* const e = (BYTE*)dstEnd;
 
-    do { memcpy(d,s,16); d+=16; s+=16; } while (d<e);
+    do { memcpy(d,s,16); memcpy(d+16,s+16,16); d+=32; s+=32; } while (d<e);
 }
 
 /*-************************************
@@ -317,7 +317,7 @@ void LZ4_wildCopy16(void* dstPtr, const void* srcPtr, void* dstEnd)
 #define LASTLITERALS   5   /* see ../doc/lz4_Block_format.md#parsing-restrictions */
 #define MFLIMIT       12   /* see ../doc/lz4_Block_format.md#parsing-restrictions */
 #define MATCH_SAFEGUARD_DISTANCE  ((2*WILDCOPYLENGTH) - MINMATCH)   /* ensure it's possible to write 2 x wildcopyLength without overflowing output buffer */
-#define FASTLOOP_SAFE_DISTANCE 32
+#define FASTLOOP_SAFE_DISTANCE 64
 static const int LZ4_minLength = (MFLIMIT+1);
 
 #define KB *(1 <<10)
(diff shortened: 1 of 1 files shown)


Cyan4973:
Looks great @djwatson !

I noticed some non-trivial differences in compression speed, according to your measurements.
Are they related to your patch too, or is it measurement noise ?

Cyan4973:
I confirm big decompression speed gains in benchmark measurements on `silesia`.
On my laptop, with `clang`, decompression speed went up, from 2800 MB/s to 3000 MB/s.
On my desktop, with `gcc`, gains were more dramatic, from 3150 MB/s to 3700 MB/s.
Desktop + gcc8 + `calgary` : from 3200 MB/s to 3550 MB/s.

However, in all tests, compression speed was completely stable.
I don't understand yet why your measurements seem to show some differences on compression side.

djwatson:
Any compression speed changes are noise.

lib/lz4.c
```diff
@@ -1434,6 +1504,26 @@ typedef enum { decode_full_block = 0, partial_decode = 1 } earlyEnd_directive;
 #undef MIN
 #define MIN(a,b)    ( (a) < (b) ? (a) : (b) )
 
+LZ4_FORCE_INLINE unsigned read_variable_length(const BYTE**ip, const BYTE* lencheck, int loop_check, int initial_check, int* error) {
```
Cyan4973: Could you document this function ?
Especially its arguments and their roles ?

djwatson: Documented, and changed error code to a typedef enum.

lib/lz4.c
```diff
@@ -297,6 +297,75 @@ void LZ4_wildCopy(void* dstPtr, const void* srcPtr, void* dstEnd)
...
+
+/* customized variant of memcpy, which can overwrite up to 32 bytes beyond dstEnd */
+LZ4_FORCE_O2_INLINE_GCC_PPC64LE
+void LZ4_wildCopy32(void* dstPtr, const void* srcPtr, void* dstEnd)
+{
+    BYTE* d = (BYTE*)dstPtr;
+    const BYTE* s = (const BYTE*)srcPtr;
+    BYTE* const e = (BYTE*)dstEnd;
+
+    memcpy(d,s,16); d+=16; s+=16;
```
Cyan4973: question : 
this function selects to copy 16 bytes first,
and then, provided that length is large enough, will proceed with further stripes of 32 bytes.

Any reason it starts with 16 bytes, instead of 32 bytes ?
I guess you made tests to reach this conclusion, and maybe could share some results ?

djwatson: Starting with 16, then doing a copy loop of 32:

```
lz4 1.8.3                 319 MB/s  2914 MB/s     6428742  63.07 silesia/dickens
lz4 1.8.3                 493 MB/s  2614 MB/s    26435667  51.61 silesia/mozilla
lz4 1.8.3                 494 MB/s  3376 MB/s     5440937  54.57 silesia/mr
lz4 1.8.3                 845 MB/s  3884 MB/s     5533040  16.49 silesia/nci
lz4 1.8.3                 412 MB/s  2651 MB/s     4338918  70.53 silesia/ooffice
lz4 1.8.3                 456 MB/s  2977 MB/s     5256666  52.12 silesia/osdb
lz4 1.8.3                 329 MB/s  2483 MB/s     3181387  48.00 silesia/reymont
lz4 1.8.3                 557 MB/s  3516 MB/s     7716839  35.72 silesia/samba
lz4 1.8.3                 441 MB/s  3985 MB/s     6790273  93.63 silesia/sao
lz4 1.8.3                 361 MB/s  2714 MB/s    20139988  48.58 silesia/webster
lz4 1.8.3                1218 MB/s  8687 MB/s     8390195  99.01 silesia/x-ray
lz4 1.8.3                 718 MB/s  3611 MB/s     1227495  22.96 silesia/xml
```

Going straight to a 32byte copy loop:
```
lz4 1.8.3                 312 MB/s  3091 MB/s     6428742  63.07 silesia/dickens
lz4 1.8.3                 499 MB/s  2744 MB/s    26435667  51.61 silesia/mozilla
lz4 1.8.3                 496 MB/s  3486 MB/s     5440937  54.57 silesia/mr
lz4 1.8.3                 877 MB/s  4328 MB/s     5533040  16.49 silesia/nci
lz4 1.8.3                 424 MB/s  2839 MB/s     4338918  70.53 silesia/ooffice
lz4 1.8.3                 459 MB/s  3202 MB/s     5256666  52.12 silesia/osdb
lz4 1.8.3                 326 MB/s  2619 MB/s     3181387  48.00 silesia/reymont
lz4 1.8.3                 551 MB/s  3555 MB/s     7716839  35.72 silesia/samba
lz4 1.8.3                 431 MB/s  4116 MB/s     6790273  93.63 silesia/sao
lz4 1.8.3                 355 MB/s  2872 MB/s    20139988  48.58 silesia/webster
lz4 1.8.3                1206 MB/s  9369 MB/s     8390195  99.01 silesia/x-ray
lz4 1.8.3                 682 MB/s  3623 MB/s     1227495  22.96 silesia/xml
```

So yea we should change it to 32-byte, which makes sense: We only use this wildcopy when we are already copying >= 15 bytes.

Cyan4973:
Thanks for all these interesting data points @djwatson .
Indeed, the improvements related to overlapping copies are much larger than expectation. It's great !

One last question : 
the new decoder seems to completely remove the previous "2-stages shortcut", which is designed to special-case "common" small_run+small_matches.

Is that intentional ?

From a high level, it seems your optimizations, which target long copies, and overlapping copies, feel complementary with the shortcut, which specifically avoids these cases.

Cyan4973:
> For match copy lengths that don't overflow, it did a memcopy by 8+8+2. We still do this, but 16+2.

There is a bit more to this story.
By doing `8+8+2`, this makes the shortcut compatible
with offsets in the [8-15] range.

In contrast, doing `16+2` requires the offset to be at least 16+.

The second version saves one load instruction per sequence,
but the first version makes more sequences compatible with the shortcut.
It might not seem much, but this triggers an impact on branch predictability.

The shorcut is expected to be useful "as often as possible".
Restraining conditions of compatible sequences slightly reduces its predictability,
so now there are a bit more branch misses, hence more pipeline flushes, etc.

With modern cpus, the cost of branch unpredictability is much higher that the cost of an instruction.
So it favors the `8+8+2` trade off.

djwatson:
> I just made a test of the proposed PR on an ARM platform.
> Unfortunately, results are not so good on this platform :

Unfortunately I don't have any qualcomm arm chips, I'll have to borrow one to dig in to it. 

Alternatively, we could #ifdef this to only x86, but it looked like there were some nice wins on larger arm chips, it would be nice to fix it. 

> By doing 8+8+2, this makes the shortcut compatible with offsets in the [8-15] range.
> The second version saves one load instruction per sequence, but the first version makes more sequences compatible with the shortcut. It might not seem much, but this triggers an impact on branch predictability.

It's even slightly more complicated:  The first version (8+8+2) potentially triggers the load blocked by store condition, while doing it by-16 doesn't seem to.  However we don't do anything special to avoid this in memcpy_using_offset for 8-15 (currently only 1,2,4), although we could, but for the shortcut it's probably too small to worry about anyway. 

Benchmarks:

```
this branch:
lz4 1.8.3                 325 MB/s  3074 MB/s     6428742  63.07 silesia/dickens
lz4 1.8.3                 527 MB/s  2808 MB/s    26435667  51.61 silesia/mozilla
lz4 1.8.3                 515 MB/s  3522 MB/s     5440937  54.57 silesia/mr
lz4 1.8.3                 886 MB/s  4299 MB/s     5533040  16.49 silesia/nci
lz4 1.8.3                 436 MB/s  2885 MB/s     4338918  70.53 silesia/ooffice
lz4 1.8.3                 468 MB/s  3198 MB/s     5256666  52.12 silesia/osdb
lz4 1.8.3                 334 MB/s  2576 MB/s     3181387  48.00 silesia/reymont
lz4 1.8.3                 569 MB/s  3581 MB/s     7716839  35.72 silesia/samba
lz4 1.8.3                 447 MB/s  4133 MB/s     6790273  93.63 silesia/sao
lz4 1.8.3                 475 MB/s  3013 MB/s   100881331  47.60 silesia/silesia.tar
lz4 1.8.3                 379 MB/s  2857 MB/s    20139988  48.58 silesia/webster
lz4 1.8.3                1236 MB/s  9404 MB/s     8390195  99.01 silesia/x-ray
lz4 1.8.3                 729 MB/s  3616 MB/s     1227495  22.96 silesia/xml

      41681.143709      task-clock (msec)         #    0.991 CPUs utilized          
               505      context-switches          #    0.012 K/sec                  
                11      cpu-migrations            #    0.000 K/sec                  
            291159      page-faults               #    0.007 M/sec                  
      123153166169      cycles                    #    2.955 GHz                      (75.03%)
      275849355190      instructions              #    2.24  insn per cycle           (75.04%)
       49297484532      branches                  # 1182.729 M/sec                    (75.02%)
        1668515452      branch-misses             #    3.38% of all branches          (75.02%)

      42.073131221 seconds time elapsed


Adding this patch: 

diff --git a/lib/lz4.c b/lib/lz4.c
index ddaab77..ccc697e 100644
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1649,8 +1649,9 @@ LZ4_decompress_generic(
 
-                    if (offset >= 16) {
-                        memcpy(op, match, 16);
+                    if (offset >= 8) {
+                        memcpy(op, match, 8);
+                        memcpy(op+8, match+8, 8);
                         memcpy(op+16, match+16, 2);
                         op += length;
                         continue;

lz4 1.8.3                 329 MB/s  2991 MB/s     6428742  63.07 silesia/dickens
lz4 1.8.3                 536 MB/s  2867 MB/s    26435667  51.61 silesia/mozilla
lz4 1.8.3                 514 MB/s  3495 MB/s     5440937  54.57 silesia/mr
lz4 1.8.3                 901 MB/s  4321 MB/s     5533040  16.49 silesia/nci
lz4 1.8.3                 438 MB/s  2958 MB/s     4338918  70.53 silesia/ooffice
lz4 1.8.3                 469 MB/s  3150 MB/s     5256666  52.12 silesia/osdb
lz4 1.8.3                 333 MB/s  2574 MB/s     3181387  48.00 silesia/reymont
lz4 1.8.3                 568 MB/s  3571 MB/s     7716839  35.72 silesia/samba
lz4 1.8.3                 446 MB/s  3977 MB/s     6790273  93.63 silesia/sao
lz4 1.8.3                 499 MB/s  3067 MB/s   100881331  47.60 silesia/silesia.tar
lz4 1.8.3                 378 MB/s  2819 MB/s    20139988  48.58 silesia/webster
lz4 1.8.3                1234 MB/s  9311 MB/s     8390195  99.01 silesia/x-ray
lz4 1.8.3                 729 MB/s  3645 MB/s     1227495  22.96 silesia/xml


      41687.349560      task-clock (msec)         #    0.991 CPUs utilized          
               505      context-switches          #    0.012 K/sec                  
                 9      cpu-migrations            #    0.000 K/sec                  
            291158      page-faults               #    0.007 M/sec                  
      120422000483      cycles                    #    2.889 GHz                      (75.01%)
      274586415771      instructions              #    2.28  insn per cycle           (75.03%)
       47621821770      branches                  # 1142.357 M/sec                    (75.03%)
        1566369291      branch-misses             #    3.29% of all branches          (75.04%)

      42.076941335 seconds time elapsed
```

The overall benchmark numbers look neutral-ish to me, but indeed the total branches and branch misses drops a bit.  I can add the above patch.  Thanks!

Cyan4973:
I noticed a few changes in the PR, so decided to give it another benchmark go.

I confirm the great speed up on x64. Tested on `clang-7` and `gcc-8`, they both feature equivalent benefits.

However, the situation on Qualcom's `aarch64` is still the same,
with many files showing large performance losses.

I guess it's not possible to correctly observe/debug the situation from x64.
You will likely need a similar platform to measure performance and optimize.

I can lend you a mobile `aarch64` terminal with `termux` setup for your convenience if you wish.
Another possibility is to request a temporary access to one `aarch64` server platform.
Both actions seem complementary by the way, I would expect both chips to behave differently, in spite of the same underlying `aarch64` ipo.

Worst case would be to only keep the new decoder for `x86`/`x64`,
but I worry about having 2 versions of the decoder to maintain.
The main reason we have a single `LZ4_decompress_generic()` with all modes converging into it is to avoid having multiple places to maintain and keep in sync. Better avoid it again if possible.

Cyan4973:
Worth noting : 
on another ARM terminal, the difference in perf is no longer as detrimental.
It actually seems a _little_ bit positive, though this is arguably within noise level.

cpu is different : 
```
> lscpu
Architecture:         aarch64
Byte Order:           Little Endian
CPU(s):               8
On-line CPU(s) list:  0
Off-line CPU(s) list: 1-7
Thread(s) per core:   1
Core(s) per socket:   1
Socket(s):            1
Vendor ID:            ARM
Model:                2
Model name:           Cortex-A53
Stepping:             r0p2
CPU max MHz:          1500.0000
CPU min MHz:          400.0000
Flags:                fp asimd aes pmull sha1 sha2 crc32
```

makes things a bit more complex to interpret ...

djwatson:
Restricted the new fastpath to x86 until we can track down the regression.

Cyan4973:
> Restricted the new fastpath to x86 until we can track down the regression.

Was this last statement supposed to be followed by additional patches ?

djwatson:
Ah sorry, pushed the wrong branch :/

Last patch skips over first loop for things not x86.  I also kept the previous shortcut in the original loop.  Thanks

djwatson:
fix msvc unreachable code warning

Cyan4973:
Thanks @djwatson !
It's a great speed optimization for LZ4 decoder ! Definitely worth being part of next release !

I'll try to save up some time later this month to investigate the issue on Qualcom's chips.

Cyan4973:
OK, I'm back onto `lz4` code base, with the objective to aim at `v1.9.0` release.

One major point of this last sprint is checking this optimization path on ARM, and more specifically for `aarch64`. The path has been gated to be enabled on x86/x64 only, so it needs to be force-enabled for other targets, such as`aarch64`. To this end, I created a new build macros, `LZ4_FAST_DEC_LOOP`, which makes it possible to control the decoder loop behavior from command line at compile time.

First results, on `Qualcomm Kryo 2` using `clang version 7.0.1`:
normal (gated off):
```
 1#calgary.tar       :   3265536 ->   1686059 (1.937), 139.1 MB/s , 841.8 MB/s
```
force-enabled : 
```
 1#calgary.tar       :   3265536 ->   1686059 (1.937), 139.8 MB/s , 644.2 MB/s
```

Tested run consistently multiple times. That's clearly a large regression.

Interestingly, `gcc v8.3.0` has recently been fixed for `termux` on Android environment, so it's another viable option. And with this compiler, I cannot measure any difference.

normal (gated off):
```
 1#calgary.tar       :   3265536 ->   1686059 (1.937), 113.3 MB/s , 815.3 MB/s
```
force-enabled : 
```
 1#calgary.tar       :   3265536 ->   1686059 (1.937), 109.7 MB/s , 823.6 MB/s
```

`clang` outcome != `gcc` outcome. Since there is no gain either, gating this optimization to remain disabled on `aarch64` seems a good choice for the time being. 

Now, onto investigating more finely....

lib/lz4.c
```diff
@@ -297,7 +297,76 @@ void LZ4_wildCopy(void* dstPtr, const void* srcPtr, void* dstEnd)
...
+
+/* customized variant of memcpy, which can overwrite up to 32 bytes beyond dstEnd */
+LZ4_FORCE_O2_INLINE_GCC_PPC64LE
+void LZ4_wildCopy32(void* dstPtr, const void* srcPtr, void* dstEnd)
+{
+    BYTE* d = (BYTE*)dstPtr;
+    const BYTE* s = (const BYTE*)srcPtr;
+    BYTE* const e = (BYTE*)dstEnd;
+
+    do { memcpy(d,s,16); memcpy(d+16,s+16,16); d+=32; s+=32; } while (d<e);
```
Cyan4973: I'll document that this function makes 2 copies of 16 bytes instead of a single copy of 32 bytes,
because it must be compatible with `offset >= 16`.

It probably doesn't make a difference in most architectures nowadays, 
but as `AVX` 256-bit registers become the world in the x64 world,
copying 32-bytes might actually become an advantage at some point, and the question will come.

lib/lz4.c
```diff
@@ -1483,13 +1587,135 @@ LZ4_decompress_generic(
...
+                        op += restSize;
+                }   }
+                continue;
+            }
+
+            /* copy match within block */
+            cpy = op + length;
+
+            /* partialDecoding : may not respect endBlock parsing restrictions */
+            assert(op<=oend);
```
Cyan4973: Considering that `LZ4_wildCopy32` starts by blindly copying 32 bytes at `op` position, I guess this condition `op<=oend` is not enough to avoid a buffer overflow.

Note that I'm pretty sure that the condition is met. It's just the `assert()` itself which feels wrong.

lib/lz4.c
```diff
@@ -1483,13 +1587,135 @@ LZ4_decompress_generic(
...
+                        memcpy(op, lowPrefix, restSize);
+                        op += restSize;
+                }   }
+                continue;
+            }
+
+            /* copy match within block */
+            cpy = op + length;
+
+            /* partialDecoding : may not respect endBlock parsing restrictions */
```
Cyan4973: I guess this comment is a left-over

lib/lz4.c
```diff
@@ -297,7 +297,76 @@ void LZ4_wildCopy(void* dstPtr, const void* srcPtr, void* dstEnd)
...
+    case 2: 
+        memcpy(v, srcPtr, 2);
+        memcpy(&v[2], srcPtr, 2);
+        memcpy(&v[4], &v[0], 4);
+        goto copy_loop;
+    case 4: 
+        memcpy(v, srcPtr, 4);
+        memcpy(&v[4], srcPtr, 4);
+        goto copy_loop;
+    case 3:
```
Cyan4973: I guess this can be simplified to `default:`

terrelln:
Seeing as the performance is fine on gcc, and terrible on clang, I would bet its a compiler problem. Something like it makes a bad inlining decision now, or it is failing to constant fold one of the "template parameters".

If you can narrow down what part of the patch is causing the regression, I bet it can be improved by rewriting the same or similar logic in a slightly different way.

Cyan4973:
I agree @terrelln .

Narrowing down the issue may help us understand what works better for the compiler, eventually uncovering an even more favorable trade-off.

PR #659, opened 2019-04-05
LZ4_resetStream*_fast()

Cyan4973:
Following discussion on LZ4 v1.9.0 API evolution,
here is a proposal pushing `LZ4_resetStream*fast()` into "stable" category.

As a consequence, it triggers a number of additional decisions related to API consistency.
- `LZ4_resetStream*_fast()` is now the recommended method to start a new stream of blocks, since it's faster.
- But it cannot completely replace `LZ4_resetStream*()` for uninitialized objects
- Since it's hard to explain the difference between `reset` and `reset_fast` in term of "one is an initializer, the other is the recommended reset", I created `LZ4_initStream*()` for the sole purpose of initializing `LZ4_stream*_t` objects allocated by the user (typically on stack) and which may contain arbitrary garbage values.

The main issue in this patch is : what is the exact scope of `LZ4_initStream*()` ?
See the quip document on LZ4 API evolution for details.

I have 2 different logics implemented in this patch, depending on fast of HC variant.

choice 1 (implemented in HC) : `LZ4_initStreamHC()` is the exact copy of `LZ4_resetStreamHC`, just a name change. As it does essentially the same thing, transition is easier. But it mixes initialization and reset role, by requesting a compression level too. Also, it accepts any `void*` pointer as argument, as an attempt to be a more general initializer, but does not control buffer size.

choice 2 (implemented in fast) : `LZ4_initStream()` is a pure universal initializer, able to accept any kind of buffer, allocated in any way by the user. As a consequence, prototype is slightly different : it controls the size, it returns a pointer of target type (`LZ4_stream_t`), and it may fail (return `NULL`), specifically if size condition is not respected (note : it should also control alignment - tbd).

This situation will have to be updated before merge, by selecting a single logic for API consistency.

The purpose of this PR is to collect feedback on the target API design.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1126,6 +1126,7 @@ _failure:
 int LZ4_compress_fast_extState(void* state, const char* source, char* dest, int inputSize, int maxOutputSize, int acceleration)
 {
     LZ4_stream_t_internal* const ctx = & LZ4_initStream(state, sizeof(LZ4_stream_t)) -> internal_donotuse;
+    assert(ctx != NULL);
     if (acceleration < 1) acceleration = ACCELERATION_DEFAULT;
     if (maxOutputSize >= LZ4_compressBound(inputSize)) {
         if (inputSize < LZ4_64Klimit) {
@@ -1236,7 +1237,8 @@ int LZ4_compress_fast_force(const char* src, char* dst, int srcSize, int dstCapa
  * _continue() call without resetting it. */
 static int LZ4_compress_destSize_extState (LZ4_stream_t* state, const char* src, char* dst, int* srcSizePtr, int targetDstSize)
 {
-    LZ4_initStream(state, sizeof (*state));
+    void* const s = LZ4_initStream(state, sizeof (*state));
+    assert(s != NULL); (void)s;
 
     if (targetDstSize >= LZ4_compressBound(*srcSizePtr)) {  /* compression success is guaranteed */
         return LZ4_compress_fast_extState(state, src, dst, *srcSizePtr, targetDstSize, 1);
@@ -1307,6 +1309,8 @@ LZ4_stream_t* LZ4_initStream (void* buffer, size_t size)
     return (LZ4_stream_t*)buffer;
 }
 
+/* resetStream is now deprecated,
+ * prefer initStream() which is more general */
 void LZ4_resetStream (LZ4_stream_t* LZ4_stream)
 {
     DEBUGLOG(5, "LZ4_resetStream (ctx:%p)", LZ4_stream);
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1125,9 +1125,8 @@ _failure:
 
 int LZ4_compress_fast_extState(void* state, const char* source, char* dest, int inputSize, int maxOutputSize, int acceleration)
 {
-    LZ4_stream_t_internal* ctx = &((LZ4_stream_t*)state)->internal_donotuse;
+    LZ4_stream_t_internal* const ctx = & LZ4_initStream(state, sizeof(LZ4_stream_t)) -> internal_donotuse;
     if (acceleration < 1) acceleration = ACCELERATION_DEFAULT;
-    LZ4_resetStream((LZ4_stream_t*)state);
     if (maxOutputSize >= LZ4_compressBound(inputSize)) {
         if (inputSize < LZ4_64Klimit) {
             return LZ4_compress_generic(ctx, source, dest, inputSize, NULL, 0, notLimited, byU16, noDict, noDictIssue, acceleration);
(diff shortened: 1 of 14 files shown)


lib/lz4.h
```diff
@@ -238,11 +238,25 @@ typedef union LZ4_stream_u LZ4_stream_t;  /* incomplete type (defined later) */
...
+ *  it's necessary to initialize first using LZ4_initStream().
+ *
+ *  After that, start any new stream with LZ4_resetStream_fast().
+ *  A same LZ4_stream_t can be re-used multiple times consecutively
+ *  and compress multiple streams,
+ *  provided that it starts each new stream with LZ4_resetStream_fast().
+ *
+ *  LZ4_resetStream_fast() is much faster than LZ4_initStream(),
+ *  but is not compatible with memory regions containing garbage data.
+ *  For this reason, LZ4_stream_t must be initialized at least once,
```
felixhandte: The one topic I think this comment doesn't cover is how this interacts with the `_extState` functions. One would like to think that the interaction is trivial, and that a sequence of calls:

1. `LZ4_initStream()`
2. `LZ4_compress_fast_extState_fastReset()`
3. `LZ4_resetStream_fast()`
4. `LZ4_compress_fast_extState_fastReset()`

would be just as fast as:

1. `LZ4_initStream()`
2. `LZ4_compress_fast_extState_fastReset()`
3. `LZ4_compress_fast_extState_fastReset()`

However, they actually have different behaviors. `LZ4_resetStream_fast()` forces the stream into `byU32` mode. Whereas `LZ4_compress_fast_extState_fastReset()` on small values will use `byU16` mode. So in the first case, two (or actually is it three?) full resets happen, while in the second only one does.

In short: you shouldn't call `LZ4_resetStream_fast()` with the `*extState*` functions. Do we want to point that out? I would propose adding:

> Note: it's only useful to call this in the context of streaming compression. The `*extState*` functions perform their own resets and using this before calling them is redundant or even counterproductive.

Speaking of which, do we want to promote `LZ4_compress_fast_extState_fastReset()` to stable as well?

Cyan4973: Indeed, I haven't yet covered `_extState` functions.
It's good you bring that up.

The additional comment you propose looks good to me.
I'll add it.

> LZ4_compress_fast_extState_fastReset()

I'm not sure.
I was afraid of the increased complexity.

If it's just promoted, with no other change, now there will be 2 functions with a minor performance behavior. Hence a risk of confusion. And for what gain ?

I understand the gain in the context of Managed Compression,
but for external users, they may as well be lost in the subtleties.

Another solution is to go "all in", with `LZ4_compress_fast_extState_fastReset()` taking the place of `LZ4_compress_fast_extState()`.
Since `LZ4_compress_fast_extState_fastReset()` cannot completely replace `LZ4_compress_fast_extState()`, due to issue with uninitialized states, `LZ4_compress_fast_extState()` must remain available. But it can be de-prioritized.

Then it's a matter of documenting that the state must be initialized first, using `LZ4_initStream()`, before invoking `LZ4_compress_fast_extState_fastReset()`.
Also, this function name is really long and feels complex.
Not too big of an issue in the experimental section, but something to think about on reaching "stable" API status.

Cyan4973: _edit_ : suggested comment added

lib/lz4.c
```diff
@@ -1281,10 +1280,18 @@ LZ4_stream_t* LZ4_createStream(void)
     LZ4_STATIC_ASSERT(LZ4_STREAMSIZE >= sizeof(LZ4_stream_t_internal));    /* A compilation error here means LZ4_STREAMSIZE is not large enough */
     DEBUGLOG(4, "LZ4_createStream %p", lz4s);
     if (lz4s == NULL) return NULL;
-    LZ4_resetStream(lz4s);
+    LZ4_initStream(lz4s, sizeof(*lz4s));
     return lz4s;
 }
 
+LZ4_stream_t* LZ4_initStream (void* buffer, size_t size)
```
terrelln: This buffer needs to be aligned, right? We should either return `NULL`, or align into the buffer, and state that the returned pointer might not be at the same address.

Cyan4973: Yeah, I was hesitating between the 2 options.

This is equivalent to `ZSTD_*_initStatic()`,
and I'm vaguely remembering someone asking me if the return pointer had effectively the same value as the source pointer, because he wanted to use this property to only store one pointer for the later `free()`.

So if it's a natural inclination, then it's better to spell the alignment restriction explicitly, and refuse any input that does not respect it, by returning `NULL`.

terrelln: I think thats the better option

lib/lz4.h
felixhandte: Yeah. Happy to leave `LZ4_compress_fast_extState_fastReset()` in experimental.

Cyan4973:
`initStream()` and `initStreamHC` have been refactored
to become pure initializers, able to accept any memory area,
presumed statically allocated.
They both control size and alignment, and may fail if relevant conditions are not respected.

Changed all test cases to use the new pair of functions.

PR #678, opened 2019-04-17
Fix out-of-bound read in LZ4_decompress_fast()

### LINKED ISSUE #676: ASAN issues in the new decompression loop in 1.9.0
Hi, guys!
I have run large set of tests using your brand new decompression, and AddressSanitizer had found troubles with https://github.com/lz4/lz4/blob/dev/lib/lz4.c#L1684
Probably `16` should be replaced with `length` variable.
Here is ASAN output, worth to note that code in `lz4.c` is shifted someway because I patched it.

```
==544083==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x606000005040 at pc 0x000000721886 bp 0x7fffaaddbb50 sp 0x7fffaaddb300
READ of size 16 at 0x606000005040 thread T0
    #0 0x721885 in __asan_memcpy .../asan_interceptors_memintrinsics.cc:23:3
    #1 0x1577e51 in LZ4_decompress_generic .../lz4/lz4.c:1681:17
    #2 0x1577e51 in LZ4_decompress_fast .../lz4/lz4.c:1986

SUMMARY: AddressSanitizer: heap-buffer-overflow .../asan_interceptors_memintrinsics.cc:23:3 in __asan_memcpy
Shadow bytes around the buggy address:
  0x0c0c7fff89b0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c0c7fff89c0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c0c7fff89d0: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x0c0c7fff89e0: fa fa fa fa fa fa fa fa fd fd fd fd fd fd fd fd
  0x0c0c7fff89f0: fa fa fa fa fd fd fd fd fd fd fd fd fa fa fa fa
=>0x0c0c7fff8a00: 00 00 00 00 00 00 00 00[fa]fa fa fa fd fd fd fd
  0x0c0c7fff8a10: fd fd fd fd fa fa fa fa fd fd fd fd fd fd fd fd
  0x0c0c7fff8a20: fa fa fa fa fd fd fd fd fd fd fd fd fa fa fa fa
  0x0c0c7fff8a30: fd fd fd fd fd fd fd fd fa fa fa fa fd fd fd fd
  0x0c0c7fff8a40: fd fd fd fd fa fa fa fa fd fd fd fd fd fd fd fd
  0x0c0c7fff8a50: fa fa fa fa fd fd fd fd fd fd fd fd fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:              cc
```

Cyan4973:
fix #676,

also deprecated `LZ4_decompress_fast()`,
which now generates deprecation warnings by default.

Note also that `LZ4_decompress_fast()` is now slower than `LZ4_decompress_safe()`,
removing any incentive to use it in place of the more secure variant.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -135,9 +135,6 @@
 #  endif  /* _MSC_VER */
 #endif /* LZ4_FORCE_INLINE */
 
-#undef LZ4_FORCE_INLINE
-#define LZ4_FORCE_INLINE static /* disable */
-
 /* LZ4_FORCE_O2_GCC_PPC64LE and LZ4_FORCE_O2_INLINE_GCC_PPC64LE
  * Gcc on ppc64le generates an unrolled SIMDized loop for LZ4_wildCopy,
  * together with a simple 8-byte copy loop as a fall-back path.
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -135,6 +135,9 @@
 #  endif  /* _MSC_VER */
 #endif /* LZ4_FORCE_INLINE */
 
+#undef LZ4_FORCE_INLINE
+#define LZ4_FORCE_INLINE static /* disable */
+
 /* LZ4_FORCE_O2_GCC_PPC64LE and LZ4_FORCE_O2_INLINE_GCC_PPC64LE
  * Gcc on ppc64le generates an unrolled SIMDized loop for LZ4_wildCopy,
  * together with a simple 8-byte copy loop as a fall-back path.
@@ -1671,7 +1674,10 @@ LZ4_decompress_generic(
                     {
                         goto safe_literal_copy;
                     }
-                LZ4_wildCopy32(op, ip, cpy);
+                if (endOnInput)
+                    LZ4_wildCopy32(op, ip, cpy);
+                else
+                    LZ4_wildCopy(op, ip, cpy);  /* LZ4_decompress_fast() cannot copy more than 8 bytes at a time : it doesn't know input length, and only relies on end-of-block properties */
                 ip += length; op = cpy;
             } else {
                 cpy = op+length;
@@ -1681,7 +1687,12 @@ LZ4_decompress_generic(
                         goto safe_literal_copy;
                     }
                 /* Literals can only be 14, but hope compilers optimize if we copy by a register size */
-                memcpy(op, ip, 16);
+                if (endOnInput)
+                    memcpy(op, ip, 16);
+                else {  /* LZ4_decompress_fast() cannot copy more than 8 bytes at a time : it doesn't know input length, and only relies on end-of-block properties */
+                    memcpy(op, ip, 8);
+                    if (length > 8) memcpy(op+8, ip+8, 8);
+                }
                 ip += length; op = cpy;
             }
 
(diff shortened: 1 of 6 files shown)


hsiangkao:
Hi Yann,
Some news to see deprecated LZ4_decompress_fast()...

Considering our Linux EROFS have the usage which will calulate the needed decompressed size, and we could also get if the decompressed size is the full decompressed size or not.

If it is not the decompressed size, upstream LZ4_decompress_safe_partial have already been used for this case.
If it is a full decompressed size, LZ4_decompress_fast can be used before since we know the _exact_ decompressed size but not compressed size (since we are fixed-output 4k compression, it means it can compress as much as possible data to 4k block. We only know the srcCapacity is 4k).

If LZ4_decompress_fast is deprecated, we can only use LZ4_decompress_safe_partial in that case.

My concern is "is it possible to introduce srcCapacity(4k but it is not the exact compressed size, which can be promised >= compressed size) and the _exact_ decompressed size to behave as fast as the current optimized LZ4_decompress_safe?"

Thanks,
Gao Xiang

Cyan4973:
Hi @hsiangkao ,

yes, what you suggest is the proper way to do it.

We could have a function that stops at a pre-defined output size, like `LZ4_decompress_fast()`, without a need to know the source size. 
But it needs to know, at a minimum, the maximum input size, in order to guarantee it will not read beyond input buffer limits. This is something `LZ4_decompress_fast()` lacks. A new prototype needs to be created to request and use this parameter.

The problem is that it requires time to create a new variant. And I'm in a bind, with other priorities to attend.

On short term, a simple solution is : continue to use `LZ4_decompress_fast()`, it's simple to disable the deprecation warnings, and you are in one of the rare cases where using `LZ4_decompress_fast()` makes sense.

Medium term, a new decompression variant will have to be created. Its design might be close to current `LZ4_decompress_safe_partial()`, as you suggest, it's just that I haven't thought this through.

hsiangkao:
Yes, I agree.

At a time, erofs was designed after I read LZ4 format, noticed the original
existed LZ4_decompress_fast and hack some code like the current
LZ4_decompress_safe_partial.

It seemd no need to record the exact compressed size in principle
as the result of my preliminary analysis after we had the exact decompressed size
and srcCapacity. Therefore I decided to not record decompressed size since it is redundant (at least add a srcCapacity is enough to ensure that it can be decompressed safely as fast as another approach.)

and Yes, that is another story, but I am looking forward a new interface could behave
as the current LZ4_decompress_safe... or I will dig into that and have a try after fs is stable...

Thanks,
Gao Xiang

PR #685, opened 2019-04-20
Add --list option to display compressed file information.

### LINKED ISSUE #598: [Feature] list an archive content with --list
Almost all compressors allows to see the archive contents without of full extracting:
```
$ gzip --help | grep list
  -l, --list        list compressed file contents
```
This allows to see a real uncompressed file size, the original file name, the original modify time and extremely useful for archive manages to show this information for a user.

For example I archived with `lzop` the file `Original File Name.txt` and then listed it's contents:
```
$ lzop -lvN archive.lzo 
Method         Length    Packed  Ratio     Date    Time   Name
------         ------    ------  -----     ----    ----   ----
LZO1X-1       3454395   1287345  37.3%  2018-10-31 05:53  Original File Name.txt
```

So I wanted to add `lz4` support to the `engrampa` (from Mate Desktop)  and to the `File Roller` (from Gnome)  archive managers and wanted to use the `-l` option but it turned out that

```
 -l     : compress using Legacy format (Linux kernel compression)
```

And it looks like no any option to just see an archive content.
So my question is:
1. Which meta information about the original file lz4 does preserve? 
 * Can we see uncompressed size (looks like with `--content-size` this information is stored) and which maximum size can be shown (e.g. gzip can't show sizes more that 4Gib).
* modification date and time and does it contains time zone info
* permissions
* file name - and it can be non ASCII and contains some special characters and even new line symbol `\n`
* MIME content type
2. Is any possibility to add the `--list` option?

gabrielstedman:
This should close #598
Supports both single lz4 file and multiple files with -m
```
+ echo '1234!'
+ ./lz4
+ ./lz4 README.md -BX --content-size
+ echo '1234!'
+ ./lz4
+ ./lz4 README.md -BD --content-size
+ ./lz4 --list test1.lz4
BlockChecksumFlag       Compressed              Uncompressed            Ratio           Filename
0                       25                      -                       -               test1
+ ./lz4 --list test2.lz4
BlockChecksumFlag       Compressed              Uncompressed            Ratio           Filename
1                       3373                    5534                    0.6095          test2
+ ./lz4 --list -m test1.lz4 test2.lz4
BlockChecksumFlag       Compressed              Uncompressed            Ratio           Filename
0                       25                      -                       -               test1
1                       3373                    5534                    0.6095          test2
+ ./lz4 --list /tmp/test1.lz4
BlockChecksumFlag       Compressed              Uncompressed            Ratio           Filename
0                       25                      -                       -               test1
+ ./lz4 --list /tmp/test2.lz4
BlockChecksumFlag       Compressed              Uncompressed            Ratio           Filename
0                       3369                    5534                    0.6088          test2
+ ./lz4 --list -m /tmp/test1.lz4 /tmp/test2.lz4
BlockChecksumFlag       Compressed              Uncompressed            Ratio           Filename
0                       25                      -                       -               test1
0                       3369                    5534                    0.6088          test2
+ ./lz4 --list -m test1.lz4 test2.lz4 /tmp/test1.lz4 /tmp/test2.lz4
BlockChecksumFlag       Compressed              Uncompressed            Ratio           Filename
0                       25                      -                       -               test1
1                       3373                    5534                    0.6095          test2
0                       25                      -                       -               test1
0                       3369                    5534                    0.6088          test2
```

### DIFF
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -40,7 +40,6 @@
 #include <stdlib.h>   /* exit, calloc, free */
 #include <string.h>   /* strcmp, strlen */
 #include "bench.h"    /* BMK_benchFile, BMK_SetNbIterations, BMK_SetBlocksize, BMK_SetPause */
-#include "lz4frame.h"
 #include "lz4io.h"    /* LZ4IO_compressFilename, LZ4IO_decompressFilename, LZ4IO_compressMultipleFilenames */
 #include "lz4hc.h"    /* LZ4HC_CLEVEL_MAX */
 #include "lz4.h"      /* LZ4_VERSION_STRING */
@@ -712,7 +711,7 @@ int main(int argc, const char** argv)
         if(!multiple_inputs){
             inFileNames[ifnIdx++] = input_filename;
         }
-        operationResult = LZ4IO_getCompressedFilesInfo(inFileNames, ifnIdx);
+        operationResult = LZ4IO_displayCompressedFilesInfo(inFileNames, ifnIdx);
         inFileNames=NULL;
     } else {
        /* compression is default action */
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -693,7 +693,7 @@ int main(int argc, const char** argv)
         break;
     }
 
-    if (multiple_inputs==0) assert(output_filename);
+    if (multiple_inputs==0 && mode != om_list) assert(output_filename);
     /* when multiple_inputs==1, output_filename may simply be useless,
      * however, output_filename must be !NULL for next strcmp() tests */
     if (!output_filename) output_filename = "*\\dummy^!//";
@@ -721,7 +721,14 @@ int main(int argc, const char** argv)
             operationResult = LZ4IO_decompressMultipleFilenames(prefs, inFileNames, (int)ifnIdx, !strcmp(output_filename,stdoutmark) ? stdoutmark : LZ4_EXTENSION);
         } else {
             operationResult = DEFAULT_DECOMPRESSOR(prefs, input_filename, output_filename);
-
+        }
+    } else if (mode == om_list){
+        if(!multiple_inputs){
+            inFileNames[ifnIdx++] = input_filename;
+        }
+        operationResult = LZ4IO_displayCompressedFilesInfo(inFileNames, ifnIdx);
+        inFileNames=NULL;
+    } else {   /* compression is default action */
         if (legacy_format) {
             DISPLAYLEVEL(3, "! Generating LZ4 Legacy format (deprecated) ! \n");
             LZ4IO_compressFilename_Legacy(prefs, input_filename, output_filename, cLevel);
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -142,7 +142,7 @@ static int usage_advanced(const char* exeName)
     DISPLAY( " -BX    : enable block checksum (default:disabled) \n");
     DISPLAY( "--no-frame-crc : disable stream checksum (default:enabled) \n");
     DISPLAY( "--content-size : compressed frame includes original size (default:not present)\n");
-    DISPLAY( "--list  : list information about .lz4 files. Only useful if compressed with --content-size flag.\n");
+    DISPLAY( "--list  : lists information about .lz4 files. Useful if compressed with --content-size flag.\n");
     DISPLAY( "--[no-]sparse  : sparse mode (default:enabled on file, disabled on stdout)\n");
     DISPLAY( "--favor-decSpeed: compressed files decompress faster, but are less compressed \n");
     DISPLAY( "--fast[=#]: switch to ultra fast compression level (default: %i)\n", 1);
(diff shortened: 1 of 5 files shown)


Cyan4973:
Looks great @gabrielstedman  !

Could you add documentation for it, within `programs/lz4.1.md` ?
I'll take care of the translation into `man` format.

programs/lz4io.c
```diff
@@ -1265,3 +1265,41 @@ int LZ4IO_decompressMultipleFilenames(LZ4IO_prefs_t* const prefs, const char** i
...
+    cfinfo->fileSize = statbuf.st_size;
+    /* Get basename without extension */
+    const char *b = strrchr(input_filename, '/');
+    if (b && b != input_filename){
+      b++;
+    } else{
+      b=input_filename;
+    }
+    const char *e = strrchr(b, '.');
+    cfinfo->fileName = malloc( (e-b+1) * sizeof(char));
```
Cyan4973: note that `malloc()` may fail, and return `NULL`.
It's important to be guarded against such exceptional conditions.
You can pair the test with `EXM_THROW()` to quickly get rid of the bad case.

programs/lz4io.c
```diff
@@ -1265,3 +1265,41 @@ int LZ4IO_decompressMultipleFilenames(LZ4IO_prefs_t* const prefs, const char** i
...
+
+size_t LZ4IO_getCompressedFileInfo(const char* input_filename, LZ4F_compFileInfo_t* cfinfo){
+    /* Get file size */
+    stat_t statbuf;
+    if (!UTIL_getFileStat(input_filename, &statbuf)){
+      EXM_THROW(60, "Can't stat file : %s", input_filename);
+    }
+    cfinfo->fileSize = statbuf.st_size;
+    /* Get basename without extension */
+    const char *b = strrchr(input_filename, '/');
```
Cyan4973: Declaring variables anywhere in the block is valid in C99, but not in C90.

For extended portability, we target a fairly strict C90 compliance (with 2 exceptions : variadic macros, and `long long` support). This makes a big accessibility difference.

programs/lz4cli.c
```diff
@@ -705,7 +708,34 @@ int main(int argc, const char** argv)
             operationResult = LZ4IO_decompressMultipleFilenames(prefs, inFileNames, ifnIdx, !strcmp(output_filename,stdoutmark) ? stdoutmark : LZ4_EXTENSION);
         else
             operationResult = DEFAULT_DECOMPRESSOR(prefs, input_filename, output_filename);
-    } else {   /* compression is default action */
+    } else if (mode == om_list){
+        LZ4F_compFileInfo_t cfinfo;
```
Cyan4973: In term of design :
try to have as little work as possible within `lz4cli.c`.

`lz4cli.c` primary role is to be a command line parser, and a job dispatcher.

I believe most of this section could probably become a function within `lz4io.c`.

programs/lz4io.h
```diff
@@ -54,6 +54,14 @@ static const char nulmark[] = "/dev/null";
 
 typedef struct LZ4IO_prefs_s LZ4IO_prefs_t;
 
+typedef struct {
+  LZ4F_frameInfo_t frameInfo;
+  char* fileName;
```
Cyan4973: unless you intend to manage `fileName` buffer lifetime from this pointer, it's likely better to make this pointer `const char*`.

programs/lz4io.c
```diff
@@ -1265,3 +1265,41 @@ int LZ4IO_decompressMultipleFilenames(LZ4IO_prefs_t* const prefs, const char** i
...
+    strncpy(cfinfo->fileName, b, (e-b));
+    cfinfo->fileName[e-b] = '\0';
+    /* init */
+    dRess_t ress;
+    LZ4F_errorCode_t const errorCode = LZ4F_createDecompressionContext(&ress.dCtx, LZ4F_VERSION);
+    if (LZ4F_isError(errorCode)) EXM_THROW(60, "Can't create LZ4F context : %s", LZ4F_getErrorName(errorCode));
+    FILE* const finput = LZ4IO_openSrcFile(input_filename);
+    if (finput==NULL) return 1;
+    /* Allocate Memory */
+    ress.srcBuffer = malloc(LZ4IO_dBufferSize);
```
Cyan4973: Same issue with `malloc()` potential `NULL` return value

gabrielstedman:
Hi, thanks for the feadback I'll update the code with this comments.

programs/lz4io.h
```diff
@@ -116,6 +125,10 @@ int LZ4IO_setSparseFile(LZ4IO_prefs_t* const prefs, int enable);
 /* Default setting : 0 == no content size present in frame header */
 int LZ4IO_setContentSize(LZ4IO_prefs_t* const prefs, int enable);
 
+int LZ4IO_getCompressedFilesInfo(const char** inFileNames,const size_t ifnIdx);
```
Cyan4973: This is the only function that is being called. 
So maybe it's the only one that deserves to be exposed.

Keep the rest internal.
There is a universal rule of API design that the less capabilities are exposed, the more implementation flexibility is preserved.

Also : naming: prefer `display` to differentiate from `get`, since in this case, information is not actually extracted, it's purely pushed onto the screen.

Cyan4973:
I intend to release a new version `v1.9.1` very soon.
Do you think this patch could make it ?

gabrielstedman:
Yes I think so just need to resolve conflicts on my end.

Cyan4973:
Thanks for the great effort in trying to bring this patch in a mergeable state.
I see there are a few issues left, on Visual compilation (which can be hard to test if you don't have a Visual compiler around), and on `make c_standards`, which is our strictest compilation test with regards to C90 compilation support.

If you wish, I could try to import your patch in a feature branch, and finish the last adaptations on my side, in the interest of time. I know it can be time-consuming to figure out all these things, and the release is now very close.

gabrielstedman:
I notice from both CI build failures compiler doesn't like compound literals in C90.  
I was following instructions found in lz4frame_manual.html using LZ4F_INIT_FRAMEINFO when creating
LZ4F_INIT_FILEINFO.
I can replace the type init with a function to make the compiler happy.  

Would you like to repalce the other compound literals such as LZ4F_INIT_FRAMEINFO and LZ4F_INIT_PREFERENCES and update lz4frame_manual.html?

gabrielstedman:
I just saw your message after posting mine.  
I'm very happy to carry on or hand it over whatever is better for you and the release.

Cyan4973:
I created feature branch `list`, where I merged your patch.
I'll proceed with fixing this branch.

gabrielstedman:
OK thank you, I've created init functions and c_standards build is now happy:
https://github.com/gabrielstedman/lz4/commit/b9f25296cdabba299b2e33ab9dfe055ee6266b4e

Cyan4973:
This patch has been integrated into #691.

Its most important limitation is that it only works for files containing a single frame.
This implementation is blind beyond the header of the first frame. 
Ideally, a good implementation should scrub the file, to check for additional frames beyond the first one, understand legacy / new / skippable frames, provide an upper bound of decompressed size based on nb of blocks, etc.

This is obviously more work.

I'm afraid I can't afford it in the limited amount of time before next release. So a limited implementation it is.

Issue #688, opened 2019-04-22
Makefile build system does not with MSYS and CYGWIN as well as having issues with MINGW-W64


JPeterMugaas:
When attempting to build lz4 for MINGW using the MSYS2 bash shell, encountered some issues.  The install does not work, the export library has a .lib suffix instead of the standard .dll.a (using gcc for mingw), and the dll has a version number in it (not usual with the MSYS2 MINGW-W64 distribution).  If building for MSYS2 itself, the Makefile generates the liblz4-$version name instead of msys-lz4-$version.dll.   It's not unusual for this stuff to be patched by the MSYS2 and Cygwin mainters.  I deally, I would like see a fix here since cmake depends indirectly on lz4.

There's several issues that I see here:
1) THe proper library suffixes and prefixes are NOT specified in the Makefiles.
2) The Makefile does not take into account Poxis-like environments such as MSYS2 and Cygwin which have version of the uname command.  The OS venvironment variable would probably always be Windows_NT even in those environments.  I prefer using "uname" rather than $(OS) in a Makefile.  Interestingly enough, the MSYS2 version of uname will return a different value based the MSYSTEM environment variable.  Examples are:
# MSYSTEM=MINGW32 uname
MINGW32_NT-10.0

# MSYSTEM=MINGW64 uname
MINGW64_NT-10.0

# uname
MSYS_NT-10.0

3) The Makefiles probably would not be able to compile binaries for Windows on a Unix-based operating system where the target is NOT the same as the host.
4) Cygwin and MSYS2 binaries are compiled similarly to MINGW in that a  seperate .dll export is usually generated and the .dll is placed in the bin directory instead of the lib directory.
5)  The MSYS, MINGW, and CYGWIN uname values might vary based upon the version of Windows where they are being built on.

I have attached a diff I made showing what I did to try address these things.  Incidentally, I would also think it's possible to embed version info in the .EXE's and .DLL's as well as an icon for the .EXE's.  It's a good practice in Windows.

[001-mingw-msys-cygwin-build-install.diff.txt](https://github.com/lz4/lz4/files/3103444/001-mingw-msys-cygwin-build-install.diff.txt)

Cyan4973:
Thanks for the insightful suggestions @JPeterMugaas .
I'm completely opened to a PR on this topic.

I see you have already attached a `diff` file in this issue.
[...]

PR #689, opened 2019-04-22
Jpm makefile - as described in https://github.com/lz4/lz4/issues/688

JPeterMugaas:
As per https://github.com/lz4/lz4/issues/688 discussion.


Makefile
```diff
@@ -36,15 +36,50 @@ TESTDIR = tests
...
+ifneq (,$(filter CYGWIN%,$(TARGET_OS)))
 EXT  = .exe
-VOID = nul
 else
 EXT  =
-VOID = /dev/null
+endif
+endif
+endif
 endif
```
Cyan4973: I'm a bit lost in the nesting levels here.
Could it be indented ?

Makefile
```diff
@@ -36,15 +36,50 @@ TESTDIR = tests
...
+TARGET_OS ?= $(shell uname)
+ifeq ($(TARGET_OS),)
+  TARGET_OS ?= $(OS)
+endif
+
+ifneq (,$(filter Windows%,$(TARGET_OS)))
+LIBLZ4 = liblz4-$(LIBVER_MAJOR)
+EXT  = .exe
+else
+EXT  = .exe
```
Cyan4973: It seems this line will define `EXT` as `=.exe`, so why redefining the same afterwards ?

Makefile
```diff
@@ -36,15 +36,50 @@ TESTDIR = tests
...
+ifneq (,$(filter MINGW%,$(TARGET_OS)))
```
Cyan4973: Several of these tests have the same outcome, could they be coalesced ?

Makefile
```diff
@@ -36,15 +36,50 @@ TESTDIR = tests
...
-VOID = nul
 else
 EXT  =
-VOID = /dev/null
+endif
+endif
+endif
 endif
 
+#determine if dev/nul based on host environment
```
Cyan4973: We have about the same questions regarding readability and nest levels here.

Cyan4973:
Generally speaking, I believe this patch goes in the right direction.
Outcome is more correct for Windows.
I would just try to simplify the top blocks, which define `.exe` or `nul`, to make them more readable.

Makefile
JPeterMugaas: Yes, they are now indented.  And I also tried to isolate some of tests that I had made or worked on.  They are now in a new Makefile.inc.  In fact, there are some things that probably belong in Makefile.inc related to some other platform stuff as well as some run-of-mill macros that's duplicate cruft.

programs/Makefile
```diff
@@ -56,17 +56,48 @@ LZ4_VERSION=$(LIBVER)
 MD2ROFF   = ronn
 MD2ROFF_FLAGS = --roff --warnings --manual="User Commands" --organization="lz4 $(LZ4_VERSION)"
```
Cyan4973: Can't `programs/Makefile` use `include Makefile.inc` too ?

JPeterMugaas: It should.

Cyan4973:
This patch looks good to me.
I can merge it now if you wish.

Or do you plan to work a bit more on it ?

JPeterMugaas:
Go ahead.  What I would do with version info could be a separate checkin.

JPeterMugaas:
OOPS. I didn't mean to close this.  Go ahead and merge.

PR #694, opened 2019-04-23
lib/Makefile: Fix detection of `Darwin`.

lembacon:
This fixes the detection of `Darwin` introduced in https://github.com/lz4/lz4/commit/773b66547f84792fd4143e605ca27f31c2989945#diff-94e18c7e8b5bab8a1215265cdb5ca1f4.

cc @JPeterMugaas

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -55,10 +55,11 @@ FLAGS    = $(CPPFLAGS) $(CFLAGS) $(LDFLAGS)
 
 SRCFILES := $(sort $(wildcard *.c))
 
+include ../Makefile.inc
 
 # OS X linker doesn't support -soname, and use different extension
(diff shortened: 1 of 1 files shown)


Cyan4973:
Nice catch @lembacon !

We have OS-X automated tests on Travis, but they were apparently unable to catch that one.
Maybe there is a need to update them...

JPeterMugaas:
I'm wondering if this DARWIN SHARED_EXT stuff should be consolidated with the $(EXT) variable or possibly the .dll suffix at the name of redistributable shared library on Windows (not the separate windows export file).

Cyan4973:
OK, I've updated tests in branch `mactest`,
and it now fails, as intended.

I'll now redirect this patch into this branch,
in order to ensure it fixes it as expected.

Cyan4973:
OK, seems to work fine.
I'll merge that patch in #695, and then merge both test + patch into `dev`.

PR #697, opened 2019-04-23
Fix test-amalgamation

ldv-alt:
* test-amalgamation: fix the list of prerequisites
* test-amalgamation: use a single cat command
* test-amalgamation: fix compilation options
* test-amalgamation: split the rule


ldv-alt:
I've force-pushed this branch to correct the commit ids fixed by this PR.

Cyan4973:
Thanks @ldv-alt , it's definitely clearer !

PR #699, opened 2019-04-26
Include block checksum in worst case scenario calculation of dstCapacity

brendene:
This results in a slightly increased dstCapacity when preferencesPtr is null.


Cyan4973:
Thanks @brendene !
You are correct, worst case must include block checksum, even if it's a rare option disabled by default.

PR #700, opened 2019-04-26
contrib: Make Meson build non optional

lzutao:
* Update ninja version to 1.9 in CI.
* Update default project version in meson script.

### DIFF
--- a/contrib/meson/meson.build
+++ b/contrib/meson/meson.build
@@ -11,9 +11,7 @@ project('lz4', ['c'],
   license: ['BSD', 'GPLv2'],
   default_options : ['c_std=c99',
     'buildtype=release'],
-  # Fall-back version in case of extracting version number from
-  # `lz4.h` failed.
-  version: '1.9.1',
+  version: 'DUMMY',
   meson_version: '>=0.47.0')
 
 cc = meson.get_compiler('c')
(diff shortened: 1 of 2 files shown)


contrib/meson/meson.build
```diff
@@ -11,7 +11,7 @@ project('lz4', ['c'],
   license: ['BSD', 'GPLv2'],
   default_options : ['c_std=c99',
     'buildtype=release'],
-  version: '1.8.3',
+  version: '1.9.1',
```
Cyan4973: Do you believe there could be a way to extract the version number from `lz4.h` directly ?
That's the way the `Makefile` works currently.

lzutao: It already extract version number from `lz4.h`. This is just a fallback
project version in case extracting version number failed.

Cyan4973: Is there a way to add a comment which would explain that ?

lzutao: Done.

devnoname120: I would suggest failing if the version cannot be extracted instead of having a fallback value.

lzutao: Currently, meson build script compares extracted version and
default fallback version. If extracted is smaller than the default, meson will
use default version. Otherwise it'll use extracted version.
Do you prefer to fail in case extracted version is smaller than fallback version
as @devnoname120 suggest, @Cyan4973 ?

Cyan4973: I believe @devnoname120 suggests to _only rely on extracted version_, and have no manually set fallback version.

While I don't think strongly about it, I would rather lean towards @devnoname120 's suggestion.

This way, if version extraction fails, we immediately have a problem, hence an alert,
which we can act upon.

lzutao: Thanks for the feedback, I will change this PR soon.

lzutao: Done.

PR #702, opened 2019-04-30
Added -fPIC compile option for static library

treefrogframework:
Added -fPIC compile option for static library.


Cyan4973:
I'm afraid this is not a suitable default flag.
There are known development environments where the static library _must not_ be compiled with `-fPIC`.

PR #705, opened 2019-05-02
Vs2019 sutions files

JPeterMugaas:


Cyan4973:
Is there a strong need for a VS2019 project solution ?
Can't VS2019 users just pick up the VS2017 one and let Visual upgrade it automatically ?

The problem with having multiple projects is that, every time there is a change in the source directory, it needs to be reflected into every single build system and script, which drags down maintenance.

I would very much prefer having a single source of truth if possible.

JPeterMugaas:
It should but I feel that those files should be provided for convenience.  I have to be honest in saying that I would prefer using cmake but unfortunately, I can't get the contrib/unnoficial-cmake to work with Visual Studio 2019.

Cyan4973:
>  I feel that those files should be provided for convenience

I used to think the same way.
But then, long term maintenance learned me the hard way the downside of providing too many build scripts.
Every little change (add a file, move a file) will have to be reflected in _all_ these scripts, otherwise CI tests bite immediately.
This is a burden, which increases silently with each new script. Each script, by itself is not a lot, so it seems it can be added. But actually, after a few of them, downsides become pretty real.

It has an invisible effect on how the code is structured, and evolves, afterwards : contributors will avoid creating any new file, since such modification quickly becomes too complex, thus restraining modularity, or generating large header files (bags of everything). It's a kind of silent auto-censorship.

Regarding Visual Studio, I would highly prefer if Visual projects could be all generated from a single `cmake` script.
Given my lack of skill in this category, I went for the second best solution : provide a single VS2010 project, supposed to be upgradable to any later version.
It almost worked, but somehow, along the way, a contributor managed to convince me that there are limitations in the VS2010 script, because of default options, which can't be edited, and are not the ones he wanted for his VS2017 project. I wasn't able to correctly evaluate the need (are all VS2017 projects impacted ? or just his own one ? are there other way to get the same result ?), so leniently accepted to add the VS2017 scripts. But that's already one step too far. If I could, I would rather suppress it. But I would first need to make sure that no project would suffer from this decision.

Because frankly, if it works as intended, just loading the VS2010 project, and letting Visual Studio automatically update it, that doesn't seem too hard nor limitative, does it ?

> I can't get the contrib/unnoficial-cmake to work with Visual Studio 2019

Yeah, so it shows it's not a trivial topic.
And unfortunately, I won't be able to help.
It's a pity we don't have some `Makefile` to `VS project` translator, without a need for `cmake` complex machinery. But I've got no better proposal so far.

For VS2019, the questions are :
- can someone use the provided VS2017 project, and just load it in VS2019 ?
- optionally, does that work as well when loading the VS2010 project instead ?

Silarn:
CMake would likely eliminate these problems, but we do have an issue trying to use appveyor running the automated upgrade with devenv because it requires a logged in account to work. To get CI working we're probably going to have to make a fork with a pre upgraded solution.

A working CMake configuration would work fine, of course. The latest CMake versions do have Visual Studio 2019 generator support, though they dropped the legacy architecture parameters and force you to use the -A flag for it to work.

PR #713, opened 2019-05-24
LZ4_decompress_generic: perofmrance improvement on AARCH64

parheliamm:
LZ4_FAST_DEC_LOOP #645 #707 was introduced to improve decompress
performance not only X86 but also ARM64.

On my ARM64 target(Cortex-a53) with LZ4_FAST_DEC_LOOP feature has
performance downgrade on lzbench test case.

lzbench 1.7.3 (64-bit Linux)   Assembled by P.Skibinski
memcpy                   1919 MB/s  1924 MB/s    10192446 100.00 /sdcard/silesia/dickens
lz4 1.8.3                  80 MB/s   448 MB/s     6428742  63.07 /sdcard/silesia/dickens
lz4 1.8.3                 120 MB/s   562 MB/s    26435667  51.61 /sdcard/silesia/mozilla
lz4 1.8.3                 127 MB/s   556 MB/s     5440937  54.57 /sdcard/silesia/mr
lz4 1.8.3                 321 MB/s   601 MB/s     5533040  16.49 /sdcard/silesia/nci
lz4 1.8.3                  83 MB/s   546 MB/s     4338918  70.53 /sdcard/silesia/ooffice
lz4 1.8.3                  87 MB/s   535 MB/s     5256666  52.12 /sdcard/silesia/osdb
lz4 1.8.3                 115 MB/s   417 MB/s     3181387  48.00 /sdcard/silesia/reymont
lz4 1.8.3                 178 MB/s   572 MB/s     7716839  35.72 /sdcard/silesia/samba
lz4 1.8.3                  68 MB/s   600 MB/s     6790273  93.63 /sdcard/silesia/sao
lz4 1.8.3                 103 MB/s   483 MB/s    20139988  48.58 /sdcard/silesia/webster
lz4 1.8.3                 152 MB/s  1064 MB/s     8390195  99.01 /sdcard/silesia/x-ray
lz4 1.8.3                 244 MB/s   583 MB/s     1227495  22.96 /sdcard/silesia/xml

Current Lz4-1.9.x dev branch has obvious downgrade on dickens case.
lzbench 1.7.3 (64-bit Linux)   Assembled by P.Skibinski
memcpy                   1913 MB/s  1934 MB/s    10192446 100.00 /sdcard/silesia/dickens
lz4 1.9.x                  80 MB/s   436 MB/s     6428742  63.07 /sdcard/silesia/dickens
lz4 1.9.x                 120 MB/s   565 MB/s    26435667  51.61 /sdcard/silesia/mozilla
lz4 1.9.x                 127 MB/s   558 MB/s     5440937  54.57 /sdcard/silesia/mr
lz4 1.9.x                 319 MB/s   623 MB/s     5533040  16.49 /sdcard/silesia/nci
lz4 1.9.x                  82 MB/s   544 MB/s     4338918  70.53 /sdcard/silesia/ooffice
lz4 1.9.x                  87 MB/s   548 MB/s     5256666  52.12 /sdcard/silesia/osdb
lz4 1.9.x                 113 MB/s   404 MB/s     3181387  48.00 /sdcard/silesia/reymont
lz4 1.9.x                 176 MB/s   581 MB/s     7716839  35.72 /sdcard/silesia/samba
lz4 1.9.x                  68 MB/s   597 MB/s     6790273  93.63 /sdcard/silesia/sao
lz4 1.9.x                 103 MB/s   481 MB/s    20139988  48.58 /sdcard/silesia/webster
lz4 1.9.x                 152 MB/s  1073 MB/s     8390195  99.01 /sdcard/silesia/x-ray
lz4 1.9.x                 242 MB/s   597 MB/s     1227495  22.96 /sdcard/silesia/xml
done... (cIters=1 dIters=1 cTime=1.0 dTime=2.0 chunkSize=1706MB cSpeed=0MB)

After apply the patch, the overall test result improvemed on my ARM64 device.
lzbench 1.7.3 (64-bit Linux)   Assembled by P.Skibinski
memcpy                   1909 MB/s  1916 MB/s    10192446 100.00 /sdcard/silesia/dickens
lz4 1.9.x      dickens            82 MB/s   446 MB/s     6428742  63.07 /sdcard/silesia/dickens
lz4 1.9.x                 125 MB/s   571 MB/s    26435667  51.61 /sdcard/silesia/mozilla
lz4 1.9.x                 131 MB/s   565 MB/s     5440937  54.57 /sdcard/silesia/mr
lz4 1.9.x                 327 MB/s   626 MB/s     5533040  16.49 /sdcard/silesia/nci
lz4 1.9.x                  87 MB/s   551 MB/s     4338918  70.53 /sdcard/silesia/ooffice
lz4 1.9.x                  90 MB/s   556 MB/s     5256666  52.12 /sdcard/silesia/osdb
lz4 1.9.x                 121 MB/s   412 MB/s     3181387  48.00 /sdcard/silesia/reymont
lz4 1.9.x                 185 MB/s   584 MB/s     7716839  35.72 /sdcard/silesia/samba
lz4 1.9.x                  73 MB/s   601 MB/s     6790273  93.63 /sdcard/silesia/sao
lz4 1.9.x                 109 MB/s   489 MB/s    20139988  48.58 /sdcard/silesia/webster
lz4 1.9.x                 164 MB/s  1067 MB/s     8390195  99.01 /sdcard/silesia/x-ray
lz4 1.9.x                 249 MB/s   603 MB/s     1227495  22.96 /sdcard/silesia/xml

To make sure this change don't have side effect, this change only enabled on ARM64
with GCC build.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -315,12 +315,7 @@ static const int      dec64table[8] = {0, 0, 0, -1, -4,  1, 2, 3};
 
 
 #ifndef LZ4_FAST_DEC_LOOP
-#  if defined(__i386__) || defined(__x86_64__)
-#    define LZ4_FAST_DEC_LOOP 1
-#  elif defined(__aarch64__) && !defined(__clang__)
-     /* On aarch64, we disable this optimization for clang because on certain
-      * mobile chipsets and clang, it reduces performance. For more information
-      * refer to https://github.com/lz4/lz4/pull/707. */
+#  if defined(__i386__) || defined(__x86_64__) || defined(__aarch64__)
 #    define LZ4_FAST_DEC_LOOP 1
 #  else
 #    define LZ4_FAST_DEC_LOOP 0
@@ -359,7 +354,6 @@ LZ4_wildCopy32(void* dstPtr, const void* srcPtr, void* dstEnd)
     BYTE* d = (BYTE*)dstPtr;
     const BYTE* s = (const BYTE*)srcPtr;
     BYTE* const e = (BYTE*)dstEnd;
-
     do { memcpy(d,s,16); memcpy(d+16,s+16,16); d+=32; s+=32; } while (d<e);
 }
 
@@ -369,7 +363,15 @@ LZ4_memcpy_using_offset(BYTE* dstPtr, const BYTE* srcPtr, BYTE* dstEnd, const si
     BYTE v[8];
     switch(offset) {
     case 1:
+#if defined(__aarch64__)
+        v[0] = srcPtr[0];
+        v[1] = srcPtr[0];
+        v[2] = srcPtr[0];
+        v[3] = srcPtr[0];
+        memcpy(v+4, v, 4);
+#else
         memset(v, *srcPtr, 8);
+#endif
         goto copy_loop;
     case 2:
         memcpy(v, srcPtr, 2);
@@ -1708,10 +1710,9 @@ LZ4_decompress_generic(
             /* get matchlength */
             length = token & ML_MASK;
 
-            if ((checkOffset) && (unlikely(match + dictSize < lowPrefix))) { goto _output_error; } /* Error : offset outside buffers */
-
             if (length == ML_MASK) {
               variable_length_error error = ok;
+              if ((checkOffset) && (unlikely(match + dictSize < lowPrefix))) { goto _output_error; } /* Error : offset outside buffers */
               length += read_variable_length(&ip, iend - LASTLITERALS + 1, endOnInput, 0, &error);
               if (error != ok) { goto _output_error; }
                 if ((safeDecode) && unlikely((uptrval)(op)+length<(uptrval)op)) { goto _output_error; } /* overflow detection */
@@ -1727,7 +1728,16 @@ LZ4_decompress_generic(
 
                 /* Fastpath check: Avoids a branch in LZ4_wildCopy32 if true */
                 if (!(dict == usingExtDict) || (match >= lowPrefix)) {
+#if defined(__aarch64__)
+                    if (offset >= 16) {
+                        memcpy(op, match, 16);
+                        memcpy(op+16, match+16, 2);
+                        op += length;
+                        continue;
+                     } else if (offset >= 8) {
+#else
                     if (offset >= 8) {
+#endif
                         memcpy(op, match, 8);
                         memcpy(op+8, match+8, 8);
                         memcpy(op+16, match+16, 2);
@@ -1735,6 +1745,7 @@ LZ4_decompress_generic(
                         continue;
             }   }   }
 
+            if ((checkOffset) && (unlikely(match + dictSize < lowPrefix))) { goto _output_error; } /* Error : offset outside buffers */
             /* match starting within external dictionary */
             if ((dict==usingExtDict) && (match < lowPrefix)) {
                 if (unlikely(op+length > oend-LASTLITERALS)) {


lib/lz4.c
```diff
@@ -1727,7 +1727,16 @@ LZ4_decompress_generic(
 
                 /* Fastpath check: Avoids a branch in LZ4_wildCopy32 if true */
                 if (!(dict == usingExtDict) || (match >= lowPrefix)) {
+#if defined(__aarch64__) && !defined(__clang__)
```
Cyan4973: If you want your patch to only be triggered for `gcc`, you might want to test `__GNUC__` too.
This would filter out other non-clang compilers, such as Visual for example.

lib/lz4.c
```diff
@@ -1727,7 +1727,16 @@ LZ4_decompress_generic(
...
+                    if (offset >= 16) {
+                        memcpy(op, match, 16);
+                        memcpy(op+16, match+16, 2);
+                        op += length;
+                        continue;
+                     } else if (offset >= 8) {
```
Cyan4973: I'm surprised that adding a branch ends up being so beneficial for `aarch64`.
This is in contrast with expectation, since each added branch adds a risk of pipeline flush.
That being said, your patch is backed by solid benchmark results, so I'll investigate.

lib/lz4.c
parheliamm: OK, I will check this part.

lib/lz4.c
parheliamm: I am not sure the root cause till now, I am not sure it is related with memcpy performance.
It seems like the more data copied in this condition check, the more performance binefit we could got.

parheliamm:
BTW, I port my change to kernel lz4 module, the test result improved on ZRAM but still downgrade compare with 1.8.3 in kernel.
The overall test result is patched version still a little downgrade 1%~5% compare with 1.8.3.
I plan to disassemble obj code to check it.

parheliamm:
Hi @Cyan4973:
I though I found the root cause for the ARM64 device via profiling tools.
    1. invoke kernel memcpy api if size is 32 on ARM64.
    2. using variable assignment isstead of memset on ARM64.
    3. Add offset >= 16 check and memory copy.
On the othre hand, I found the "buffer outsiite check" is much more invoked than 1.8.3.
I thought it should be a bug. I try to fix it and invoke number is back to normal after modification.
Pleae review it.
The ARM64 LZ4 decompress performance obviously improved on benchmark.

parheliamm:
Hi @Cyan4973 :
BTW, clang downgrade seems like resolved as well.
I don't have environment for clang build.
I did a test on  my clang build kernel. The zram performance seems like back to normal.
Would you please double check on your side?
If clang downgrade issue fixed, we can remove clang definition check.

Chenxi

parheliamm:
I can build lzbench for aarch64 via clang.
According to result, clang build still has small downgrade compare with 1.8.3,
v.1.8.3
lzbench-1.8.3 -elz4 /sdcard/silesia/*                                                                                                                  
lzbench 1.7.3 (64-bit Linux)   Assembled by P.Skibinski
memcpy                   1177 MB/s  1038 MB/s    10192446 100.00 /sdcard/silesia/dickens
lz4 1.8.3                  64 MB/s   411 MB/s     6428742  63.07 /sdcard/silesia/dickens
lz4 1.8.3                  93 MB/s   534 MB/s    26435667  51.61 /sdcard/silesia/mozilla
lz4 1.8.3                  95 MB/s   498 MB/s     5440937  54.57 /sdcard/silesia/mr
lz4 1.8.3                 233 MB/s   572 MB/s     5533040  16.49 /sdcard/silesia/nci
lz4 1.8.3                  65 MB/s   522 MB/s     4338918  70.53 /sdcard/silesia/ooffice
lz4 1.8.3                  71 MB/s   507 MB/s     5256666  52.12 /sdcard/silesia/osdb
lz4 1.8.3                  91 MB/s   358 MB/s     3181387  48.00 /sdcard/silesia/reymont
lz4 1.8.3                 134 MB/s   526 MB/s     7716839  35.72 /sdcard/silesia/samba
lz4 1.8.3                  50 MB/s   539 MB/s     6790273  93.63 /sdcard/silesia/sao
lz4 1.8.3                  76 MB/s   418 MB/s    20139988  48.58 /sdcard/silesia/webster
lz4 1.8.3                 112 MB/s   857 MB/s     8390195  99.01 /sdcard/silesia/x-ray
lz4 1.8.3                 168 MB/s   508 MB/s     1227495  22.96 /sdcard/silesia/xml

v1.9.x-with patch
lzbench 1.7.3 (64-bit Linux)   Assembled by P.Skibinski
memcpy                   1188 MB/s  1037 MB/s    10192446 100.00 /sdcard/silesia/dickens
lz4 1.9.x                  56 MB/s   410 MB/s     6428742  63.07 /sdcard/silesia/dickens
lz4 1.9.x                  82 MB/s   532 MB/s    26435667  51.61 /sdcard/silesia/mozilla
lz4 1.9.x                  80 MB/s   493 MB/s     5440937  54.57 /sdcard/silesia/mr
lz4 1.9.x                 207 MB/s   562 MB/s     5533040  16.49 /sdcard/silesia/nci
lz4 1.9.x                  56 MB/s   517 MB/s     4338918  70.53 /sdcard/silesia/ooffice
lz4 1.9.x                  62 MB/s   508 MB/s     5256666  52.12 /sdcard/silesia/osdb
lz4 1.9.x                  79 MB/s   358 MB/s     3181387  48.00 /sdcard/silesia/reymont
lz4 1.9.x                 108 MB/s   495 MB/s     7716839  35.72 /sdcard/silesia/samba
lz4 1.9.x                  43 MB/s   528 MB/s     6790273  93.63 /sdcard/silesia/sao
lz4 1.9.x                  67 MB/s   418 MB/s    20139988  48.58 /sdcard/silesia/webster
lz4 1.9.x                 100 MB/s   843 MB/s     8390195  99.01 /sdcard/silesia/x-ray
lz4 1.9.x                 145 MB/s   507 MB/s     1227495  22.96 /sdcard/silesia/xml
done... (cIters=1 dIters=1 cTime=1.0 dTime=2.0 chunkSize=1706MB cSpeed=0MB)

It is close to 1.8.3 but still has downgrade. so FAST_DEC_LOOP still enable only on GCC.
I will update patch soon.

Cyan4973:
@parheliamm , I will have access to 2 ARM64 systems soon.
I should be able to test both `gcc` and `clang` if all goes according to plan.

lib/lz4.c
```diff
@@ -359,8 +359,11 @@ LZ4_wildCopy32(void* dstPtr, const void* srcPtr, void* dstEnd)
     BYTE* d = (BYTE*)dstPtr;
     const BYTE* s = (const BYTE*)srcPtr;
     BYTE* const e = (BYTE*)dstEnd;
-
+#if defined(__aarch64__) && defined(__GNUC__)
+    do { memcpy(d,s,32); d+=32; s+=32; } while (d<e);
```
Cyan4973: This change seems incorrect, and can corrupt data.

There's a difference between 2x 16-byte writes, and 1x 32-byte write:
it matters whenever dealing with overlapping matches.

For offset in the [16-32[ range, 2x16-byte will work, and regenerate correct data,
while 1x32-byte will copy some garbage into the last byte(s).

This usage happens at line 195.

To witness the problem, one need a file which happen to generate matches at offset [16-32[,
and an ability to check that decompressed content is the same as original one
(`lz4` internal checksum can do it).

lib/lz4.c
```diff
@@ -369,7 +372,15 @@ LZ4_memcpy_using_offset(BYTE* dstPtr, const BYTE* srcPtr, BYTE* dstEnd, const si
     BYTE v[8];
     switch(offset) {
     case 1:
+#if defined(__aarch64__) && defined(__GNUC__)
+        v[0] = srcPtr[0];
+        v[1] = srcPtr[0];
+        v[2] = srcPtr[0];
+        v[3] = srcPtr[0];
+        memcpy(v+4, v, 4);
```
Cyan4973: This modification is interesting.
It's correct I think, just it's surprising it ends up being faster than `memset()`.
I guess it might depend on the compiler, and its ability to properly optimize `memset()` for a constant length.
Anyway, it would be good to try to isolate how much gains are achieved through this change alone.

lib/lz4.c
```diff
@@ -1708,10 +1719,9 @@ LZ4_decompress_generic(
             /* get matchlength */
             length = token & ML_MASK;
 
-            if ((checkOffset) && (unlikely(match + dictSize < lowPrefix))) { goto _output_error; } /* Error : offset outside buffers */
```
Cyan4973: So, if I understand correctly,
this check has been moved forward into the complex, and replaced by 2 instances of the same test.

While I guess it works correctly, I wonder what's the need for it ?

lib/lz4.c
parheliamm: Hi @Cyan4973 :
Please refer to below dump from ARM64:
The GCC invoke memset API directlt.
    case 1:
        memset(v, *srcPtr, 8);
     390:	f94013e0 	ldr	x0, [sp,#32]
     394:	39400000 	ldrb	w0, [x0]
     398:	2a0003e1 	mov	w1, w0
     39c:	9100e3e0 	add	x0, sp, #0x38
     3a0:	d2800102 	mov	x2, #0x8                   	// #8
     3a4:	94000000 	bl	0 <memset>
It is need more time from gprof:
  2.22      2.14     0.05                             memset
On the other hand, X86 GCC invoke the memset with avx instruction which provide much better performance than generic implementation.
  1.05      0.94     0.01                             __memset_avx2

lib/lz4.c
parheliamm: Hi @Cyan4973 :
If you profiling this condition check invoked number comparison between 1.9.x and 1.8.3.
You will see the condition check difference via lzbench with dickens test case.
v1.8.3 34959
v.1.9.x 1055885

After investigate the code, we could see the difference.
v.1.8.3 SKIP the condition check if 
if condition is true in:
https://github.com/lz4/lz4/blob/v1.8.3/lib/lz4.c#L1463
AND below condition is true
https://github.com/lz4/lz4/blob/v1.8.3/lib/lz4.c#L1478
The **"if ((checkOffset) && (unlikely(match + dictSize < lowPrefix))) goto _output_error;   /* Error : offset outside buffers */"** should not be invoked.

v1.9.3
The "if ((checkOffset) && (unlikely(match + dictSize < lowPrefix))) goto _output_error;   /* Error : offset outside buffers */" code will be invoked in every loop which lead to downgrade.
So the fix would be move this check to specific condition to avoid useless condition check.

After this change, the call number is same as v1.8.3
Chenxi

lib/lz4.c
parheliamm: Thanks for explationation, I will roll it back.

Cyan4973:
I've been testing these proposed changes on a Qualcom Kryo 2 cpu (`aarch64`), compiling using both `clang v8.0` and `gcc v8.3`.

In comparing performance with `dev` branch, 
`clang` results are virtually unchanged, which I suspect is normal since this patch targets `gcc` only,
but `gcc` results are unfortunately down, and by a very large margin. 
We are talking 20-30% slower, depending on file.

parheliamm:
> I've been testing these proposed changes on a Qualcom Kryo 2 cpu (`aarch64`), compiling using both `clang v8.0` and `gcc v8.3`.
> 
> In comparing performance with `dev` branch,
> `clang` results are virtually unchanged, which I suspect is normal since this patch targets `gcc` only,
> but `gcc` results are unfortunately down, and by a very large margin.
> We are talking 20-30% slower decoding speed, depending on file.
> 
> Note that, on this cpu at least, `dev` branch produces substantially better decompression speed than `v1.8.3`.

You meant to say GCC build is also has downgrade via your local test?

parheliamm:
Clang with FAST_DEC
On my test build, clang with fast_dec_loop improved as well on patchset  bfdb5db 

lzbench 1.7.3 (64-bit Linux)   Assembled by P.Skibinski
memcpy                   1851 MB/s  1868 MB/s    10192446 100.00 /sdcard/silesia/dickens
lz4 1.9.x                  72 MB/s   452 MB/s     6428742  63.07 /sdcard/silesia/dickens
lz4 1.9.x                 107 MB/s   568 MB/s    26435667  51.61 /sdcard/silesia/mozilla
lz4 1.9.x                 110 MB/s   566 MB/s     5440937  54.57 /sdcard/silesia/mr
lz4 1.9.x                 288 MB/s   610 MB/s     5533040  16.49 /sdcard/silesia/nci
lz4 1.9.x                  74 MB/s   548 MB/s     4338918  70.53 /sdcard/silesia/ooffice
lz4 1.9.x                  79 MB/s   554 MB/s     5256666  52.12 /sdcard/silesia/osdb
lz4 1.9.x                 103 MB/s   416 MB/s     3181387  48.00 /sdcard/silesia/reymont
lz4 1.9.x                 158 MB/s   578 MB/s     7716839  35.72 /sdcard/silesia/samba
lz4 1.9.x                  60 MB/s   599 MB/s     6790273  93.63 /sdcard/silesia/sao
lz4 1.9.x                  92 MB/s   485 MB/s    20139988  48.58 /sdcard/silesia/webster
lz4 1.9.x                 141 MB/s  1079 MB/s     8390195  99.01 /sdcard/silesia/x-ray
lz4 1.9.x                 218 MB/s   599 MB/s     1227495  22.96 /sdcard/silesia/xml

dev branch:
lzbench 1.7.3 (64-bit Linux)   Assembled by P.Skibinski
memcpy                   1896 MB/s  1904 MB/s    10192446 100.00 /sdcard/silesia/dickens
lz4 1.9.x                  72 MB/s   449 MB/s     6428742  63.07 /sdcard/silesia/dickens
lz4 1.9.x                 108 MB/s   562 MB/s    26435667  51.61 /sdcard/silesia/mozilla
lz4 1.9.x                 112 MB/s   554 MB/s     5440937  54.57 /sdcard/silesia/mr
lz4 1.9.x                 286 MB/s   587 MB/s     5533040  16.49 /sdcard/silesia/nci
lz4 1.9.x                  75 MB/s   547 MB/s     4338918  70.53 /sdcard/silesia/ooffice
lz4 1.9.x                  80 MB/s   542 MB/s     5256666  52.12 /sdcard/silesia/osdb
lz4 1.9.x                 104 MB/s   416 MB/s     3181387  48.00 /sdcard/silesia/reymont
lz4 1.9.x                 158 MB/s   568 MB/s     7716839  35.72 /sdcard/silesia/samba
lz4 1.9.x                  60 MB/s   589 MB/s     6790273  93.63 /sdcard/silesia/sao
lz4 1.9.x                  93 MB/s   477 MB/s    20139988  48.58 /sdcard/silesia/webster
lz4 1.9.x                 143 MB/s  1080 MB/s     8390195  99.01 /sdcard/silesia/x-ray
lz4 1.9.x                 217 MB/s   590 MB/s     1227495  22.96 /sdcard/silesia/xml
done... (cIters=1 dIters=1 cTime=1.0 dTime=2.0 chunkSize=1706MB cSpeed=0MB)

Cyan4973:
> You meant to say GCC build is also has downgrade via your local test?

Yes.

My feeling is that the `aarch64` environment is not uniform,
and that a performance change observed in one environment
does not guarantee an equivalent change in another `aarch64` environment.

Therefore, we can separately observe radically different outcome for the same code modification.

I'll have to be cautious on this front, and only consider changes with clear performance benefits across the board. This can happen sometimes, when removing a useless / redundant operation for example. Another good example is offering code simplification at no performance cost.

I'll have to pass, as this patch does not belong to any of these categories, unfortunately.

parheliamm:
I thought you should review the checkoffset change which should be a bug fix for fast_dec_loop.

Cyan4973:
I thought the `checkoffset` change was a performance enhancement one ?
You mean it affects correctness too ?

parheliamm:
> I thought the `checkoffset` change was a performance enhancement one ?
> You mean it affects correctness too ?

Yes, as I mentioned previously, the checkoffset invoke number is much more than v1.8.3
My pull request has 2 parts.
1. aarch64 changes
2. checkoffset calls reductions which could improve performance obviously.

Chenxi

Cyan4973:
OK, so this is a performance enhancement suggestion, not a correctness issue.

However, it seems to belong to the category "reduce nb of operations", so could end up being a general benefit on any platform.

I'll have to test this change in isolation though, and come back with numbers.

parheliamm:
> OK, so this is a performance enhancement suggestion, not a correctness issue.
> 
> However, it seems to belong to the category "reduce nb of operations", so could end up being a general benefit on any platform.
> 
> I'll have to test this change in isolation though, and come back with numbers.

One more comments, I am not sure whether X86 could see the benefit directly.
aarch64 with gcc build would be better choice.

Cyan4973:
Sure !

I'll try all architectures and compilers with the new PR.

parheliamm:
new request #719

PR #714, opened 2019-05-24
Fix src temporary buffer use in lz4frame

dreambottle:
I noticed a bug, that forces lz4frame decompression to use a temporary source buffer in the case when size hint is respected, but dst buffer is not fully exhausted.

After next block header is read, the `dstage_getCBlock` state expects input based on  dctx->tmpInTarget, but remaining buffer size will always be 0, after reading a header, if src buf had been of a hint size, so it transitions to `dstage_storeCBlock` state. So, when one optimizes their code to always provide the src buffer of the hinted size, all compressed blocks will always hit the temporary buffer, unless they're lucky having destination buffer exhausted.

This has a very notable speed impact on some embedded hardware, where only a limited amount of fast memory is available and you never want to hit tmp buffers.

### DIFF
--- a/lib/lz4frame.c
+++ b/lib/lz4frame.c
@@ -1494,7 +1494,7 @@ size_t LZ4F_decompress(LZ4F_dctx* dctx,
                 /* next block is a compressed block */
                 dctx->tmpInTarget = nextCBlockSize + crcSize;
                 dctx->dStage = dstage_getCBlock;
-                if (dstPtr==dstEnd) {
+                if (dstPtr==dstEnd || srcPtr==srcEnd) {
                     nextSrcSizeHint = BHSize + nextCBlockSize + crcSize;
                     doAnotherStage = 0;
                 }


Cyan4973:
Sounds like a good patch !
I believe the condition you added on `srcEnd` is the correct one,
and I'm even wondering if the previous condition on `dstEnd` is useful at all.
Maybe there was some confusion.

What I would like to see on top of this patch is a test scenario, that show the benefit.
It's not simple, since it is a performance benefit, not a correctness issue.

I can take care of this. I'll spend some time trying to reproduce the issue and observe the performance difference.

Cyan4973:
step 1 achieved : 
I could reproduce the issue, by creating a new scenario in `fullbench`, which specifically decompress using `LZ4F` frame API and following the src size hint.

Compared to the "single pass" policy (provide everything to `LZ4F_decompress()` in a single invocation), decompression speed decreases from 3.5 GB/s to 2.3 GB/s.

So that's definitely sensible.

Cyan4973:
step 2 :
tested your patch on the new scenario.

Speed improves, from 2.3 GB/s to 2.7 GB/s.

That's definitely better, though not yet close to 3.5 GB/s.

I'll investigate a bit, to make sure reasons are understood correctly.
Current speculation is that calling the function `LZ4F_decompress()` once per block introduces some overhead compared to once per frame. But let's make sure it's not another reason.

Cyan4973:
I can confirm that the patch achieves the intended effect : input data is read directly from input buffer, instead of being first copied from input buffer into temporary buffer.

So, from my perspective, it's good to go.
I'll ship the new test case separately.

Issue #704, opened 2019-05-02
Question: lz4 -t nonexistant.lz4 returns 0


imartisko:
lz4 -t ./nonexistant.lz4                                                                                         
./nonexistant.lz4: No such file or directory 
echo $?                                                                                                          
0

Is this expected behaviour?

Cyan4973:
It's clearly a bug.
Thanks for reporting @imartisko !

Cyan4973:
fixed by @gabrielstedman (#715)

PR #715, opened 2019-05-25
Handle file not existing case #704

gabrielstedman:


Cyan4973:
Thanks for this welcomed patch @gabrielstedman !

A few questions :
- Could you add a relevant test case in `tests/Makefile`, showing the patch is effective (ideally it should fail without this patch), and ensuring this property remains preserved in future changes
- What happens in cases where `input_filename` is neither a regular file nor `stdin` ? I'm thinking about a directory name, which is a valid input in association with `-r` command (note : even if it works all fine, I believe a relevant test is also welcomed)

gabrielstedman:
I'll definitively add a test for this.
I'll also correct it so that it doesn't occur when `-r` is used.
While I was testing the `-r` option I found a weird behaviour (not sure if I'm doing something wrong).
`

Cyan4973:
This is strange indeed.
`-r` should require or transparently enable ` -m`.
Even then, since `/tmp/test` is supposed to contain only `test1.txt` and `test2.txt`, it should only provide the 2 associated `.lz4` compressed files.

Is it certain than `/tmp/test` wasn't already filled with `.lz4` files before starting the test ?

gabrielstedman:
Sorry my bad env was dirty with bad lz4 files from previous tests.
After cleaning up it works as expected:
```
 » ~/lz4 (ret) mkdir -p /tmp/test && echo "test1" > /tmp/test/test1.txt && echo "test2" > /tmp/test/test2.txt
 » ~/lz4 (ret) ./lz4 -r /tmp/test
```

How about using `-tm` ? is this meant to be supported?

```
 » ~/lz4 (ret) ./lz4 -tm /tmp/test/test1.txt.lz4 /tmp/test/test2.txt.lz4
/tmp/test/test1.txt already exists; not overwritten  
/tmp/test/test2.txt already exists; not overwritten  
 » ~/lz4 (ret) echo $?
2

```

Cyan4973:
> How about using -tm ? is this meant to be supported?

I suppose it should.
But I don't remember if it ever was.

gabrielstedman:
OK, So I've included a simple test to prove this patch works and made sure that it doesn't consider input when run with `-r`

Cyan4973:
Looks great, thanks !

PR #717, opened 2019-05-30
Added documentation and macro to support in-place compression and decompression

Cyan4973:


Cyan4973:
Tracking the new in-place decompression scenario, sanitizer tools signal an issue related to `memcpy()`.

`memcpy()` is only supposed valid if `src` and `dst` memory area are disjointed.

This is normally correct, since `memcpy()` is performed from the `src` buffer into the `dst` buffer, which are generally different.
However, in an in-place decompression scenario, this is no longer the case.

The most important issue is about the _last_ literals run, which is handled by a single last `memcpy()`. Because of this, this last `memcpy()` can be much larger than 32 bytes, hence in an in-place decompression scenario, it can overlap source and destination. This is undefined behavior.

So this must be fixed. There are a number of issues though : 
- Using `memcpy()` instead of `wildcopy()` for the last run makes a big speed difference for poorly compressible data, that tend to have long literal runs. We don't want to lose this performance.
- The newer `wildcopy32()` might reduce the difference, though it remains to be tested. Also, `memcpy()` respects precisely size directive, while `wildcopy()` overwrites beyond intended end, so this must be managed by extra-code.
- Another simple solution could be to replace this last `memcpy()` by a `memmove()`, which takes care of overlapping memory regions. Here also, the main constraint is to avoid losing performance for the general case. This can be tested.

Cyan4973:
OK, I've been testing replacing `memcpy()` by `memmove()` for the final literal run, and from what I can tell, performance difference is not sensible,
so let's use this route, it's the simplest one.

PR #719, opened 2019-05-31
FAST_DEC_LOOP: move offset check in specific condition.

parheliamm:
When I did FAST_DEC_LOOP performance test, I found the
offset check is much more than v1.8.3

You will see the condition check difference via lzbench with dickens test case.
v1.8.3 34959
v.1.9.x 1055885

After investigate the code, we could see the difference.
v.1.8.3 SKIP the condition check if
if condition is true in:
https://github.com/lz4/lz4/blob/v1.8.3/lib/lz4.c#L1463
AND below condition is true
https://github.com/lz4/lz4/blob/v1.8.3/lib/lz4.c#L1478\
The offset check should be invoked.

v1.9.3
The offset check code will be invoked in every loop which lead to downgrade.
So the fix would be move this check to specific condition
to avoid useless condition check.

After this change, the call number is same as v1.8.3

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1708,10 +1708,9 @@ LZ4_decompress_generic(
             /* get matchlength */
             length = token & ML_MASK;
 
-            if ((checkOffset) && (unlikely(match + dictSize < lowPrefix))) { goto _output_error; } /* Error : offset outside buffers */
-
             if (length == ML_MASK) {
               variable_length_error error = ok;
+              if ((checkOffset) && (unlikely(match + dictSize < lowPrefix))) { goto _output_error; } /* Error : offset outside buffers */
               length += read_variable_length(&ip, iend - LASTLITERALS + 1, endOnInput, 0, &error);
               if (error != ok) { goto _output_error; }
                 if ((safeDecode) && unlikely((uptrval)(op)+length<(uptrval)op)) { goto _output_error; } /* overflow detection */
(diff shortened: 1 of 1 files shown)


Cyan4973:
I've been testing this proposed change, and it indeed provides measurable decompression speed benefits, on all platforms. From what I can tell, outcome vary depending on file, from +0% to +4%, hence being at worst neutral, but generally positive.

This is a welcomed speed improvement !
Thanks @parheliamm, I'll merge it into `dev` branch .

parheliamm:
> I've been testing this proposed change, and it indeed provides measurable decompression speed benefits, on all platforms. From what I can tell, outcome vary depending on file, from +0% to +4%, hence being at worst neutral, but generally positive.
> 
> This is a welcomed speed improvement !
> Thanks @parheliamm, I'll merge it into `dev` branch .

@Cyan4973 :
Thanks for your verification.
I guess your verification is x86, I think aarch64 is more sensitive for this patch no matter on gcc or clang.
I will update lkml for this change later.
https://lkml.org/lkml/2019/5/14/939

Cyan4973:
It depends.
On `aarch64` "Kryo 2", I found `clang` 8.0 performance virtually identical, before and after the patch.

`gcc` 8.3 saw some performance improvements, but it was in the 2-3 % range, so not particularly better than `x64` ones.

PR #721, opened 2019-06-04
FAST_DEC_LOOP: decompress speed improvements on ARM64/X86

parheliamm:
@Cyan4973 :
FAST_DEC_LOOP: improve condition check to avoid branch miss 
This patch is to avoid branch miss if offest smaller than 8.
I found if check offset >=8 first, the decompress speed can be improved on ARM64.
Normal DEC check offset first as well.
https://github.com/lz4/lz4/blob/dev/lib/lz4.c#L1815
So this patch don't have function change but improve the performance.


parheliamm:
Need to double check the result.
This change is not the root cause for arm64

PR #722, opened 2019-06-04
FAST_DEC_LOOP: Avoid useless check and save instructions for ARM64/X86

parheliamm:
Commit 4da3360 decompress_generic: Refactor variable length fields
use a inline function for Read the variable-length literal or match length.

However this change will introduce extra condition check for length == ML_MASK

For ARM64 has to run needless code even though initial_check is false.
if (initial_check && unlikely((*ip) >= lencheck))
   175c4:	b94153e0 	ldr	w0, [sp,#336]
   175c8:	7100001f 	cmp	w0, #0x0
   175cc:	540001e0 	b.eq	17608 <LZ4_decompress_safe+0x29c>

This change roll back 4da3360 and avoid the extra
condition check on length == ML_MASK

After this change, ARM64 with GCC FAST_DEC_LOOP obvious improved.
X86 might be improved slightly.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1562,37 +1562,6 @@ typedef enum { decode_full_block = 0, partial_decode = 1 } earlyEnd_directive;
 #undef MIN
 #define MIN(a,b)    ( (a) < (b) ? (a) : (b) )
 
-/* Read the variable-length literal or match length.
- *
- * ip - pointer to use as input.
- * lencheck - end ip.  Return an error if ip advances >= lencheck.
- * loop_check - check ip >= lencheck in body of loop.  Returns loop_error if so.
- * initial_check - check ip >= lencheck before start of loop.  Returns initial_error if so.
- * error (output) - error code.  Should be set to 0 before call.
- */
-typedef enum { loop_error = -2, initial_error = -1, ok = 0 } variable_length_error;
-LZ4_FORCE_INLINE unsigned
-read_variable_length(const BYTE**ip, const BYTE* lencheck, int loop_check, int initial_check, variable_length_error* error)
-{
-  unsigned length = 0;
-  unsigned s;
-  if (initial_check && unlikely((*ip) >= lencheck)) {    /* overflow detection */
-    *error = initial_error;
-    return length;
-  }
-  do {
-    s = **ip;
-    (*ip)++;
-    length += s;
-    if (loop_check && unlikely((*ip) >= lencheck)) {    /* overflow detection */
-      *error = loop_error;
-      return length;
-    }
-  } while (s==255);
-
-  return length;
-}
-
 /*! LZ4_decompress_generic() :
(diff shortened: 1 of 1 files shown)


parheliamm:
Test result on lzbench:
v1.9.1 w/patch:
lz4 1.9.x                  79 MB/s   **466 MB/s**     6428742  63.07 /sdcard/silesia/dickens
v1.8.3
lz4 1.8.3                  77 MB/s   459 MB/s     6428742  63.07 /sdcard/silesia/dickens

Cyan4973:
It seems your `dev` branch is based upon a development commit of `lz4` with `FORCE_INLINE` disabled for debugging purpose. This makes benchmark results difficult to compare with earlier versions, so I would recommend to fix that (you can rebase on current commit of `lz4` project for example).

After this issue fixed, I see this patch featuring different outcome depending on file, sometimes positive, sometimes negative. So far, it's not a clear win.

parheliamm:
> It seems your `dev` branch is based upon a development commit of `lz4` with `FORCE_INLINE` disabled for debugging purpose. This makes benchmark results difficult to compare with earlier versions, so I would recommend to fix that (you can rebase on current commit of `lz4` project for example).
> 
> After this issue fixed, I see this patch featuring different outcome depending on file, sometimes positive, sometimes negative. So far, it's not a clear win.

Hi @Cyan4973 
I will try latest dev branch today.

From code analysis, this change reduce the useless condition check as below:
if (initial_check && unlikely((*ip) >= lencheck))
So it should be positive in all conditions.
Personally, I suggest deliver this patch.

ARM64 should be more sensitive for this patch especially on cortex-a53. X86 might be very very very sightly.
I will try SDM855 (cortex-a76) today if possible.

I will try GCC 8.3 with lzbench to compare dev branch and v1.8.3.

Chenxi

parheliamm:
Hi @Cyan4973:
You are correct if enable the inline, the performance improved like my patch did.
pull request closed.

parheliamm:
I build lzbench via GCC 8.3 which come from https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-a/downloads 
I found GCC 8.3 has big downgrade compare with Linaro GCC6/7 on lzbench.
GCC-8.3 w/o inline
lz4 1.9.x                  59 MB/s   422 MB/s     6428742  63.07 /sdcard/silesia/dickens
GCC-8.3 w/ inline 
lz4 1.9.x                  77 MB/s   435 MB/s     6428742  63.07 /sdcard/silesia/dickens

GCC 6.3.1 Linaro
w/o inline
lz4 1.9.x                  84 MB/s   439 MB/s     6428742  63.07 /sdcard/silesia/dickens
w/ inline
lz4 1.9.x                  82 MB/s   453 MB/s     6428742  63.07 /sdcard/silesia/dickens

GCC6 has much better performance compare with GCC8.

PR #725, opened 2019-06-05
FAST_DEC_LOOP: better decompress performance on ARM64/X86

parheliamm:
Here are 3 commits to improve decompress performance on ARM64/X86.
1. FAST_DEC_LOOP: prefer to lerverage 16 bytes copy if offset >= 16 
This patch is only affect ARM64 which has better 16 bytes copy performance compare with 8 bytes.
After this change, the 16 bytes copy invovled if offest >= 16.
2. FAST_DEC_LOOP: Add hints for better code generation. 
This could be generate better instructions for hot path.
3. FAST_DEC_LOOP: skip duplicate offset check if length == ML_MASK 
Reduce duplicate offset check number by 40% to improve the decompress speed.

Test via lzbench (GCC7.4 on ARM64(cortex-a53))
dev branch with 3 patches.
lz4 1.9.x                 127 MB/s   **586** MB/s   302640041  47.60 /sdcard/silesia
vanilla:
lz4 1.8.3                 128 MB/s   569 MB/s   302640041  47.60 /sdcard/silesia
v1.9.x-348e10(restored FORCE_INLINE)
lz4 1.9.x                 128 MB/s   577 MB/s   302640041  47.60 /sdcard/silesia

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1664,7 +1664,7 @@ LZ4_decompress_generic(
             assert(!endOnInput || ip <= iend); /* ip < iend before the increment */
 
             /* decode literal length */
-            if (length == RUN_MASK) {
+            if (unlikely(length == RUN_MASK)) {
                 variable_length_error error = ok;
                 length += read_variable_length(&ip, iend-RUN_MASK, endOnInput, endOnInput, &error);
                 if (error == initial_error) { goto _output_error; }
@@ -1688,7 +1688,7 @@ LZ4_decompress_generic(
                 if (endOnInput) {  /* LZ4_decompress_safe() */
                     DEBUGLOG(7, "copy %u bytes in a 16-bytes stripe", (unsigned)length);
                     /* We don't need to check oend, since we check it once for each loop below */
-                    if (ip > iend-(16 + 1/*max lit + offset + nextToken*/)) { goto safe_literal_copy; }
+                    if (unlikely(ip > iend-(16 + 1/*max lit + offset + nextToken*/))) { goto safe_literal_copy; }
                     /* Literals can only be 14, but hope compilers optimize if we copy by a register size */
                     memcpy(op, ip, 16);
                 } else {  /* LZ4_decompress_fast() */
@@ -1707,7 +1707,7 @@ LZ4_decompress_generic(
             /* get matchlength */
             length = token & ML_MASK;
 
-            if (length == ML_MASK) {
+            if (unlikely(length == ML_MASK)) {
               variable_length_error error = ok;
               if ((checkOffset) && (unlikely(match + dictSize < lowPrefix))) { goto _output_error; } /* Error : offset outside buffers */
               length += read_variable_length(&ip, iend - LASTLITERALS + 1, endOnInput, 0, &error);
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1717,6 +1717,7 @@ LZ4_decompress_generic(
                 if (op + length >= oend - FASTLOOP_SAFE_DISTANCE) {
                     goto safe_match_copy;
                 }
+                goto skip_dup_check;
             } else {
                 length += MINMATCH;
                 if (op + length >= oend - FASTLOOP_SAFE_DISTANCE) {
@@ -1743,6 +1744,7 @@ LZ4_decompress_generic(
             }   }   }
 
             if ((checkOffset) && (unlikely(match + dictSize < lowPrefix))) { goto _output_error; } /* Error : offset outside buffers */
+skip_dup_check:
             /* match starting within external dictionary */
             if ((dict==usingExtDict) && (match < lowPrefix)) {
                 if (unlikely(op+length > oend-LASTLITERALS)) {
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1725,7 +1725,16 @@ LZ4_decompress_generic(
 
                 /* Fastpath check: Avoids a branch in LZ4_wildCopy32 if true */
                 if (!(dict == usingExtDict) || (match >= lowPrefix)) {
+#if defined(__aarch64__) && defined(__GNUC__) && !defined(__clang__)
+                    if (offset >= 16) {
+                        memcpy(op, match, 16);
+                        memcpy(op+16, match+16, 2);
+                        op += length;
+                        continue;
+                     } else if (offset >= 8) {
+#else
                     if (offset >= 8) {
+#endif
                         memcpy(op, match, 8);
                         memcpy(op+8, match+8, 8);
                         memcpy(op+16, match+16, 2);


lib/lz4.c
```diff
@@ -1725,7 +1725,16 @@ LZ4_decompress_generic(
 
                 /* Fastpath check: Avoids a branch in LZ4_wildCopy32 if true */
                 if (!(dict == usingExtDict) || (match >= lowPrefix)) {
+#if defined(__aarch64__) && defined(__GNUC__)
```
Cyan4973: This change will also be enabled for `clang`, since it also defines `__GNUC__`. 
Is that the intention ?

parheliamm: Currently, FAST_DEC_LOOP only enabled on AARCH64 with GCC.
So I will change like this
#if defined(__aarch64__) && defined(__GNUC__) && !defined(__clang__)

Cyan4973: yes

parheliamm: @Cyan4973 
Fixed.
Please review latest commit

parheliamm:
This change is depend on the GCC version.
GCC6 has obvious improvement, but GCC7 no improvements.
GCC4.9 for kernel has downgrade after apply this patch.
So drop it.

parheliamm:
Still need to review because verification/unit test case need to be reviewed.

parheliamm:
@Cyan4973 
Please review these 2 patches.
I update the verification case in kernel side, after apply these 2 patches the overall decompress speed is 2% better than 1.8.3

For lzbench, please refer to below result
dev branch with 2 patches.
lz4 1.9.x                 131 MB/s   587 MB/s   302640041  47.60 /sdcard/silesia

dev branch with 16 bytes copy patch
lz4 1.9.x                 126 MB/s   585 MB/s   302640041  47.60 /sdcard/silesia

1.8.3
lz4 1.8.3                 130 MB/s   569 MB/s   302640041  47.60 /sdcard/silesia

Cyan4973:
Thanks @parheliamm .
I'll look into them.
I'm a bit overwhelmed right now by many other activities, but don't worry, this PR is on my list, I'll get to it.

parheliamm:
> Thanks @parheliamm .
> I'll look into them.
> I'm a bit overwhelmed right now by many other activities, but don't worry, this PR is on my list, I'll get to it.

Thanks for review my ticket, I will continue updating this ticket for my progress both on kernel and userspace.

parheliamm:
Hi @Cyan4973:
Here are 2 commits to improve decompress performance on ARM64/X86.
1. FAST_DEC_LOOP: prefer to lerverage 16 bytes copy if offset >= 16 
This patch is only affect ARM64 which has better 16 bytes copy performance compare with 8 bytes.
After this change, the 16 bytes copy invovled if offest >= 16.
2. FAST_DEC_LOOP: Add hints for better code generation. 
This could be generate better instructions for hot path.
3. FAST_DEC_LOOP: skip duplicate offset check if length == ML_MASK 
Reduce duplicate offset check number by 40% to improve the decompress speed.

Test via lzbench:
dev branch with 3 patches.
lz4 1.9.x                 127 MB/s   **586** MB/s   302640041  47.60 /sdcard/silesia
vanilla:
lz4 1.8.3                 128 MB/s   569 MB/s   302640041  47.60 /sdcard/silesia
v1.9.x-348e10(restored FORCE_INLINE)
lz4 1.9.x                 128 MB/s   577 MB/s   302640041  47.60 /sdcard/silesia

parheliamm:
@Cyan4973 
I've finished the kernel side benchmark test.
ARM64 with GCC build show improved as below:
Patched:
    READ: bw=505MiB/s (529MB/s)
Vanilla:
    READ: bw=521MiB/s (546MB/s)

Unfortunately, Clang for ARM64 in kernel has big downgrade.
So I prefer to only enable FAST_DEC_LOOP with GCC on ARM64 for safety.

Chenxi

Cyan4973:
Hi @parheliamm ,

I tried this patch on a bunch of platforms, and here are some results.

phone, Qualcomm MSM8996, Android, clang v8.0 : 
calgary : 890 MB/s => 850 MB/s (-4.7%)
silesia : 930 MB/s => 930 MB/s (-0.0%)

laptop, Core i7-8559U, macOS, llvm10 : 
calgary : 4450 MB/s => 4300 MB/s (-3.5%)
silesia : 4310 MB/s => 4240 MB/s (-1.5%)

laptop, Core i7-8559U, macOS, gcc v9.1.0 : 
calgary : 4320 MB/s => 4200 MB/s (-2.8%)
silesia : 4260 MB/s => 4230 MB/s (-0.7%)

desktop, Core i7-9700K, Ubuntu Linux, clang v8.0 : 
calgary : 3570 MB/s => 3490 MB/s (-2.3%)
silesia : 3640 MB/s => 3605 MB/s (-1.0%)

desktop, Core i7-9700K, Ubuntu Linux, gcc v8.3.0 : 
calgary : 3660 MB/s => 3630 MB/s (-0.8%)
silesia : 3760 MB/s => 3780 MB/s (+0.5%)

I understand this patch was mostly intended for `gcc` for `ARM64`, 
unfortunately this combination doesn't compile anymore on my test system, so I could not test it.

For other platforms, as one can see, the results vary depending on platform, while being generally slightly negative.

Unfortunately, it's not a clear positive signal.

parheliamm:
@Cyan4973:
Sorry for late reply.
I am back to vocation just now.
I checked your test result, I have some questions and comment for your refernce.
First of all, my chances didn't affect clang at all, so below test result is intresting:
phone, Qualcomm MSM8996, Android, clang v8.0 :
calgary : 890 MB/s => 850 MB/s (-4.7%)
silesia : 930 MB/s => 930 MB/s (-0.0%)
desktop, Core i7-9700K, Ubuntu Linux, clang v8.0 :
calgary : 3570 MB/s => 3490 MB/s (-2.3%)
silesia : 3640 MB/s => 3605 MB/s (-1.0%)
Based on my understandings of my changes, the reliable result should be clang test result is same level.
However, your result has downgade.

Your X86 test result, it also intresting, the reason is below:
1. from code analysis, the commit https://github.com/lz4/lz4/pull/725/commits/3d56642c5bc032d382717eafca7aa791e7ec9be8
should reduce lots of condition checks during decompression. 
So the test result should not like this.

So I request you to double check my changes from code level then try the benchmark.
Meanwhile, I will try X86 tomorrow on my desktop to double check as well.

Any feedback would be appreciate and looking forward to reopening this ticket.

Chenxi.

Cyan4973:
>First of all, my chances didn't affect clang at all, 

I agree that the patch regarding copy-by-16, is not supposed to be applied for `clang`.
But the code is not completely neutral beyond that part. I can see that :

- a few `unlikely()` compiler hints were added, and they will be applied by `clang`. I'm unsure of the impact of these compiler hints. I prefer to keep such statements for situations which are very rare or very common, such as checking for an error case. Here, these situations will happen. It's not clear to me that they are going to be "rare". Hence it might actually hurt performance ? A would be to benchmark such change in isolation.

- There is a new `jmp` instruction, to skip a further test. I'm unable to tell how it translates into binaries, but it could end up changing the code layout in more or less unpredictable ways.

I can indeed make new tests, to check again.
But it seems this topic is also part of your new patch #744 ?

PR #727, opened 2019-06-05
CircleCI : use custom Docker Image with pre-installed dependencies

ephiepark:
Instead of installing dependencies before test, use custom Docker Image with pre-installed dependencies.


Cyan4973:
minor : I believe some of these commits could be squashed,
it's better when one commit does one thing,
not just for doc, but also for later investigations (thing about bisection process) when need be.

Cyan4973:
Thanks @ephiepark !

Issue #738, opened 2019-06-27
A heap-buffer-overflow vulnerability in v1.9.1


cxht:
Hi!
When I run your decompression function of v1.9.1, I use ASAN to run the program and it reports a heap-buffer-overflow bug in lz4.c:1729  `memcpy(op, match, 8);
`

And here is the information printed by ASAN:  
```
==9677==ERROR:AddressSanitizer:heap-buffer-overflow on address 0x7f78cb4f3e21 at pc 0x00000046e1b2 bp 0x7ffc41c17ae0 sp 0x7ffc41c17ad0
READ of size 8 at 0x7f78cb4f3e21 thread T0
0x46e1b1 in LZ4_decompress_generic /home/cx/workspace/target_group/lz4/lib/lz4.c:1729
0x46e1b1 in LZ4_decompress_safe      /home/cx/workspace/target_group/lz4/lib/lz4.c:1990
0x5190e8 in LZ4IO_decodeLegacyStream /home/cx/workspace/target_group/lz4/programs/lz4io.c:925
0x5190e8 in selectDecoder /home/cx/workspace/target_group/lz4/programs/lz4io.c:1123
0x5190e8 in LZ4IO_decompressSrcFile /home/cx/workspace/target_group/lz4/programs/lz4io.c:1168
0x521ea7 in LZ4IO_decompressDstFile /home/cx/workspace/target_group/lz4/programs/lz4io.c:1202
0x521ea7 in LZ4IO_decompressFilename /home/cx/workspace/target_group/lz4/programs/lz4io.c:1223
0x405397 in main /home/cx/workspace/target_group/lz4/programs/lz4cli.c:741
0x7f78cf53a82f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
0x409178 in _start (/home/cx/workspace/target_group/lz4/lz4+0x409178)
0x7f78cb4f3e21 is located 2527 bytes to the left of 8388608-byte region [0x7f78cb4f4800,0x7f78cbcf4800)
allocated by thread T0 here:
0x7f78cf97c602 in malloc (/usr/lib/x86_64-linux-gnu/libasan.so.2+0x98602)
0x518df6 in LZ4IO_decodeLegacyStream /home/cx/workspace/target_group/lz4/programs/lz4io.c:902
0x518df6 in selectDecoder /home/cx/workspace/target_group/lz4/programs/lz4io.c:1123
[...]

PR #739, opened 2019-06-28
oss-fuzz: Add a fuzzing target that compiles in the oss-fuzz environment

### LINKED ISSUE #731: Integrating with OSS-Fuzz
Greetings lz4 developers and contributors,

We’re reaching out because your project is an important part of the open source ecosystem, and we’d like to invite you to integrate with our fuzzing service, OSS-Fuzz. OSS-Fuzz is a free fuzzing infrastructure you can use to identify security vulnerabilities and stability bugs in your project. OSS-Fuzz will:

- Continuously run all the fuzzers you write.
- Alert you when it finds issues.
- Automatically close issues after they’ve been fixed by a commit.

Many widely used open source projects like OpenSSL, FFmpeg, LibreOffice, and ImageMagick, ZSTD are fuzzing via OSS-Fuzz, which helps them find and remediate critical issues.

Even though typical integrations can be done in < 100 LoC, we have a reward program in place which aims to recognize folks who are not just contributing to open source, but are also working hard to make it more secure.

We want to stress that anyone who meets the eligibility criteria and integrates a project with OSS-Fuzz is eligible for a reward.

We have a sample fuzz target that you are welcome to use directly or modify, that can be used as a starting point. 

If you're not interested in integrating with OSS-Fuzz, it would be helpful for us to understand why—lack of interest, lack of time, or something else—so we can better support projects like yours in the future.

If we’ve missed your question in our FAQ, feel free to reply or reach out to us at oss-fuzz-outreach@googlegroups.com.

Thanks!
The OSS-Fuzz Team

cmeister2:
Fixes #731 

This is the initial integration of a fuzz target into lz4. Currently all this does is try to compress whatever input has been given; this code was given by @Google-Autofuzz.

Once the initial code is in both here and the oss-fuzz project, it's easy to extend to multiple fuzzers to do multiple operations (i.e. one which just compresses the input, one which decompresses the input, one which uses the API in a more exciting way...)

Would appreciate any comments or feedback as it's my first commit here! Specifically looking for feedback on the makefiles and the layout of the code - `make -C` was giving me particular pain for calling into the ossfuzz makefile, so in the end I had to drop it. (Specifically, `CFLAGS` wasn't being appended to when being called with `make -C` from the top level makefile - this might be expected behaviour but to me it was a bit weird)


ossfuzz/ossfuzz.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash -eu
+
+# This script is called by the oss-fuzz main project when compiling the fuzz
+# targets. This script is regression tested by travisoss.sh.
+
+# Save off the current folder as the build root.
+export BUILD_ROOT=$PWD
+
+# lz4 uses CPPFLAGS rather than CXX flags.
+export CPPFLAGS="${CXXFLAGS}"
```
Cyan4973: Not sure what is meant here.
The proper "equivalent" to `CXXFLAGS` is `CFLAGS`,
although they are named differently for a reason : `CFLAGS` is for C compiler, and `CXXFLAGS` is for C++ compiler. Although some flags are common, there are also C compiler flags which make no sense for C++, and vice versa.

On the other hand `CPPFLAGS` is about "preprocessor flags", for example the include directories. And this concept is common for both C and C++.

`CPPFLAGS` is implied by default for both C andC++ compilations (but not linking), so it should not be necessary to "repeat" its content inside `CXXFLAGS`.

.travis.yml
```diff
@@ -194,5 +194,11 @@ matrix:
         - pushd build
         - DESTDIR=./staging ninja install
         - tree ./staging
+
+    # oss-fuzz compilation test
```
Cyan4973: Just for my understanding :
it's just a quick test which ensures that `oss-fuzz` still compiles correctly,
it doesn't run the test nor deploy anything on Google's oss-fuzz, right ?

I noticed it fails without a "red flag" notification :
https://travis-ci.org/lz4/lz4/jobs/551658718#L455

I suspect that's because `lz4` is not yet part of Google's ossfuzz ?
Also, it means that the test will have to be modified in some way so that it triggers a red flag when it fails ?

ossfuzz/Makefile
```diff
@@ -0,0 +1,54 @@
...
+FLAGS    = $(CFLAGS) $(CPPFLAGS) $(LDFLAGS)
+
+include ../Makefile.inc
+
+# Include a rule to build the static library if calling this target
+# directly.
+$(LZ4DIR)/liblz4.a:
+	$(MAKE) -C $(LZ4DIR) CFLAGS="$(CFLAGS)" liblz4.a
+
+%.o: %.cc
```
Cyan4973: I first thought "isn't this rule already defined by default ?", 
but it uses `CXXFLAGS` for `.cc` files, instead of `CFLAGS`.

In which case, it might be cleaner to have `CXXFLAGS += CFLAGS` instead
(although there remains a risk that some C-specific flags might be invalid for C++, but it's no different in this rule).

ossfuzz/compress_fuzzer.cc
```diff
@@ -0,0 +1,22 @@
...
+#include <stdint.h>
+#include <stdlib.h>
+#include "lz4.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+  size_t const compressed_dest_size = LZ4_compressBound(size);
+  char *const dest_buffer = (char *)malloc(compressed_dest_size);
+
+  int result = LZ4_compress_default((const char*)data, dest_buffer,
```
Cyan4973: Thanks for this excellent starting point.

I will likely upgrade it so that it can achieve at least a round-trip (successful compression, followed by successful decompression, and src-dst comparison). But I can take over from this point.

ossfuzz/compress_fuzzer.cc
```diff
@@ -0,0 +1,22 @@
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include "lz4.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+  size_t const compressed_dest_size = LZ4_compressBound(size);
+  char *const dest_buffer = (char *)malloc(compressed_dest_size);
```
Cyan4973: nit: note that any invocation of `malloc()` can fail.
Therefore, it should be tested for success (`dest_buffer != NULL`).

ossfuzz/travisoss.sh
```diff
@@ -0,0 +1,24 @@
...
+set -ex
+
+# Clone the oss-fuzz repository
+git clone https://github.com/google/oss-fuzz.git /tmp/ossfuzz
+
+if [[ ! -d /tmp/ossfuzz/projects/lz4 ]]
+then
+    echo "Could not find the lz4 project in ossfuzz"
+
+    # Exit with a success code while the lz4 project is not expected to exist
```
Cyan4973: 👍

ossfuzz/compress_fuzzer.cc
```diff
@@ -0,0 +1,22 @@
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include "lz4.h"
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
```
terrelln: This project is otherwise entirely in `C`, and this file doesn't use any C++. Can you please just make this a `.c` file, and remove `extern "C"`.

Making the fuzzers `C` also makes fuzzing with MSAN much, much easier, since we don't have to worry about compiling the C++ standard library in MSAN mode.

ossfuzz/compress_fuzzer.cc
```diff
@@ -0,0 +1,22 @@
...
+{
+  size_t const compressed_dest_size = LZ4_compressBound(size);
+  char *const dest_buffer = (char *)malloc(compressed_dest_size);
+
+  int result = LZ4_compress_default((const char*)data, dest_buffer,
+                                    size, compressed_dest_size);
+
+  if (result == 0)
```
terrelln: Can we have a `CHECK()` macro here, so we can use it above as well?

```
#define CHECK(p) ((cond) ? (void)0 : abort())

File: ossfuzz/ossfuzz.sh (hunk 1 above)
Comment (terrelln) [inline ossfuzz/ossfuzz.sh]: We should be free to ignore the `CXXFLAGS` if we move the fuzz-targets to C.

File: ossfuzz/standaloneengine.cc [hunk 9]
```diff
@@ -0,0 +1,74 @@
...
+
+#include "testinput.h"
+
+/**
+ * Main procedure for standalone fuzzing engine.
+ *
+ * Reads filenames from the argument array. For each filename, read the file
+ * into memory and then call the fuzzing interface with the data.
+ */
+int main(int argc, char **argv)
```
terrelln: I think we'll want to take the one from zstd, since that works recursively on directories. Then we can easily run regression tests on CI.

But that can be done later.

ossfuzz/ossfuzz.sh
cmeister2: Ah, fair enough. I possibly got the wrong end of the stick here. Will take a look

.travis.yml
cmeister2: Yep, you got it :smile:

ossfuzz/Makefile
cmeister2: I'll take another look here.

ossfuzz/compress_fuzzer.cc
cmeister2: So what I envisaged was having separate fuzzers; this simple compress one, one that just tries to decompress the input, and any other ones which would exercise interesting paths.

It's fine to just copy any number of fuzzers into the $OUT directory.

ossfuzz/compress_fuzzer.cc
cmeister2: Agreed, will fix up.

ossfuzz/compress_fuzzer.cc
cmeister2: So there is method to my madness, in that when I've tried this in pure C before (in curl) it just plain didn't work. Happy to explore going pure C though!!

ossfuzz/compress_fuzzer.cc
cmeister2: Sure.

ossfuzz/standaloneengine.cc
cmeister2: Happy to use whatever you want :smile: though maybe would be appropriate in a followup commit?

cmeister2:
Thanks for the reviews @Cyan4973 and @terrelln, will try and apply this feedback this evening.

ossfuzz/compress_fuzzer.cc
cmeister2: Yep, trying again with just `clang`, we start running into fun.
```
# Now compile the actual fuzzer.
clang -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -I../lib -g -DLZ4_DEBUG=1  -DXXH_NAMESPACE=LZ4_  -fsanitize=fuzzer compress_fuzzer.o ../lib/liblz4.a -o compress_fuzzer
/usr/bin/ld: cannot find -lstdc++
```
However, installing libstdc++-5-dev seems to get rid of those errors. So I think I'm happy to change the scripts to do that. Worth noting that I haven't explicitly tested all the other options for building images.

cmeister2: Actually, retesting with AFL, it doesn't obviously compile with clang. Will test if C++ actually improves that situation.

cmeister2: Yep, with C++ it works fine. So if you want more fuzzer coverage I'd stick with the C++ version

cmeister2:
Ok - so I think the C++ version is probably the best version right now. It means you get `afl` and `honggfuzz` as well as `libfuzzer`.

cmeister2:
FYI I wrote a simple decompress fuzz tester and it hit #738 in about 10 seconds. I'll add this into this commit.

terrelln:
AFL works with C and C++, please make the fuzzers C.

Yeah, I also found that a simple decompression fuzzer finds that bug in seconds. We're not sure why we haven't ran into it with AFL-fuzz on the command line... But it is great to have the coverage here.

ossfuzz/compress_fuzzer.cc
terrelln: The zstd OSS-fuzz fuzzers are written in C, so I know it can work.

ossfuzz/standaloneengine.cc
terrelln: Yeah, we can do it later.

cmeister2:
@terrelln - The problem here is that the OSS-Fuzz build system builds AFL in C++ mode (as per https://github.com/google/oss-fuzz/blob/master/infra/base-images/base-builder/compile_afl#L27), so if you write your fuzz target in C and then compile and link it using just `clang`, then you run into the problems I'm describing whereby it fails to find the C++ symbols it's expecting.

I've copied the gist output from this here: https://gist.github.com/cmeister2/c60fc827ca15e225dbe1c3813ed55339

I'm not aiming to be annoying here by saying "C++ is the easiest way forward"; it's just that from experience it's been the easiest way to do this. If you have any other options I'm eager to hear them!

terrelln:
@cmeister2 you can compile the object file with C, and then link with OSS-fuzzes `LIB_FUZZING_ENGINE` with C++. See how zstd does it here https://github.com/facebook/zstd/blob/dev/tests/fuzz/Makefile#L79.

cmeister2:
ACK. Ok, not a problem. Will just use $(CXX) for linking then like those examples do :smile:

cmeister2:
Ok, all fixed up.

ossfuzz/Makefile
```diff
@@ -0,0 +1,55 @@
...
+# ##########################################################################
+
+LZ4DIR  := ../lib
+LIB_FUZZING_ENGINE ?= standaloneengine.o
+
+DEBUGLEVEL?= 1
+DEBUGFLAGS = -g -DLZ4_DEBUG=$(DEBUGLEVEL)
+
+CFLAGS   += -I$(LZ4DIR) $(DEBUGFLAGS) $(MOREFLAGS)
+CXXFLAGS += -I$(LZ4DIR) $(DEBUGFLAGS) $(MOREFLAGS)
```
terrelln: `-I$(LZ4DIR)` and `$(DEBUGFLAGS)` belong in `CPPFLAGS`, since they both modify how the C-preprocessor works. `CPPFLAGS` are shared between C++ and C, so they don't have to be duplicated.

ossfuzz/compress_fuzzer.cc
```diff
@@ -0,0 +1,26 @@
...
+#include "lz4.h"
+
+#define CHECK(COND)   if (!(COND)) { abort(); }
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+  size_t const compressed_dest_size = LZ4_compressBound(size);
+  char *const dest_buffer = (char *)malloc(compressed_dest_size);
+
+  if (dest_buffer != NULL)
```
terrelln: We can `CHECK(dest_buffer != NULL)`. It's okay to fail the fuzzer if we can't allocate memory.

ossfuzz/decompress_fuzzer.cc
```diff
@@ -0,0 +1,28 @@
...
+#include "lz4.h"
+
+#define CHECK(COND)   if (!(COND)) { abort(); }
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+  size_t const buffer_size = 10 * 1024 * 1024;
+  char *const dest_buffer = (char *)malloc(buffer_size);
+
+  if (dest_buffer != NULL)
```
terrelln: We can just `CHECK()` this.

I'd rather know that allocations are failing, then have the fuzzers silently not fuzz anything.

ossfuzz/decompress_fuzzer.cc
```diff
@@ -0,0 +1,28 @@
+#include <stddef.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include "lz4.h"
+
+#define CHECK(COND)   if (!(COND)) { abort(); }
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+  size_t const buffer_size = 10 * 1024 * 1024;
```
terrelln: Can you add a comment `// TODO: Size input buffer pseudo-randomly based on seed extracted from input`

In the future, I will want to randomly size the buffer based on a seed we extract from the input. Some bugs are only exposed when the output buffer is too small. When the output buffer is too large, we hide potential buffer-overruns beyond the end of the output buffer.

ossfuzz/testinput.h
```diff
@@ -0,0 +1,3 @@
+#include <inttypes.h>
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size);
```
terrelln: Lets make this header work in both C and C++ modes.

```c
#if defined (__cplusplus)
extern "C" {
#endif 

int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size);

#if defined(__cplusplus)
}
#endif 
```

ossfuzz/testinput.h
```diff
@@ -0,0 +1,3 @@
+#include <inttypes.h>
```
terrelln: This header is missing an include guard.

ossfuzz/compress_fuzzer.cc
cmeister2: Yep, as discussed below, all fine if you just link with C++ rather than compile.

ossfuzz/Makefile
```diff
@@ -0,0 +1,55 @@
...
+# decompress_fuzzer : OSS Fuzz test tool
+# ##########################################################################
+
+LZ4DIR  := ../lib
+LIB_FUZZING_ENGINE ?= standaloneengine.o
+
+DEBUGLEVEL?= 1
+DEBUGFLAGS = -g -DLZ4_DEBUG=$(DEBUGLEVEL)
+
+CFLAGS   += -I$(LZ4DIR) $(DEBUGFLAGS) $(MOREFLAGS)
```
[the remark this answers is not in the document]
cmeister2: Sure. I was basing this off the Makefiles in `examples` and `test` which do override the variables, but I don't mind if you want to do a different method.

ossfuzz/decompress_fuzzer.cc
cmeister2: Seems like a good plan :smile:

cmeister2:
All new comments addressed.

terrelln:
Looks good to me, thanks for the patch!

I'll wait for CI to finish, and let Yann take a look before merging it.

cmeister2:
@terrelln @Cyan4973 can I use these credentials from the zstd project in the PR to google?

primary_contact: "NickRTerrell@gmail.com"
auto_ccs:
  - "Yann.collet.73@gmail.com"
  - "terrelln@fb.com"
- "cyan@fb.com"

terrelln:
I would switch it around, and make the primary contact `Yann.collet.73@gmail.com`, and add my gmail as an `auto_ccs`.

cmeister2:
Ok, just waiting on review from @Cyan4973 i think!

ossfuzz/Makefile
```diff
@@ -0,0 +1,61 @@
...
+# decompress_fuzzer : OSS Fuzz test tool
+# ##########################################################################
+
+LZ4DIR  := ../lib
+LIB_FUZZING_ENGINE ?= standaloneengine.o
+
+DEBUGLEVEL?= 1
+DEBUGFLAGS = -g -DLZ4_DEBUG=$(DEBUGLEVEL)
+
+LZ4_CFLAGS  = $(CFLAGS) $(DEBUGFLAGS) $(MOREFLAGS)
```
Cyan4973: I guess this means `liblz4` and `%_fuzzer` will be compiled without any optimization (`-O0`) by default. Is that the intention ?

ossfuzz/Makefile
```diff
@@ -0,0 +1,61 @@
...
+LZ4_CXXFLAGS = $(CXXFLAGS) $(DEBUGFLAGS) $(MOREFLAGS)
+LZ4_CPPFLAGS = $(CPPFLAGS) -I$(LZ4DIR) -DXXH_NAMESPACE=LZ4_
+
+include ../Makefile.inc
```
Cyan4973: What do you need from `Makefile.inc` ?
Please comment.

Cyan4973:
Looks mostly good to me !
Just some remaining minor clarification requests, especially around optimization flag.

ossfuzz/Makefile
cmeister2: When attempting to compile just in the lz4 codebase without any OSS-Fuzz shenanigans? Yeah, no optimizations.

The standard here though will be for OSS-Fuzz to pass in flags, usually something like
```
CFLAGS: -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link
CXXFLAGS: -O1 -fno-omit-frame-pointer -gline-tables-only -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=fuzzer-no-link -stdlib=libc++
```
so that'll be the standard way of compiling.

ossfuzz/Makefile
cmeister2: Hmm. Probably nothing! I'll remove this if unneeded.

cmeister2:
Alright, hopefully there now!

Cyan4973:
Thanks @cmeister2 !

PR #741, opened 2019-06-29
meson: Fix deprecated warnings on build options

lzutao:
Meson now reserves `build_` prefix options.


Cyan4973:
Thanks @lzutao , looks good to me !

PR #740, opened 2019-06-29
Fix out-of-bounds read

terrelln:
The shortcut added in lz4-1.9.0 added an out-of-bounds read in the case where the match length is less than `ML_MASK` when we're not in ext-dict mode.

* Check the offset in the shortcut when we aren't in `withPrefix64k` mode.
* Always check the offset in the `safe_match_copy` label, even though it is already checked for longer matches.

Tested by running a fuzzer equivalent to the decompression fuzzer in https://github.com/lz4/lz4/pull/739. It crashed in ASAN mode in under 10 seconds with both bugs. After it doesn't crash in 10 minutes of running in ASAN, UBSAN, and MSAN mode.

That doesn't guarantee there aren't more bugs, but at least there aren't 10 second bugs.

@Cyan4973 is going to follow up and see why we didn't catch this bug with our own fuzzer tests, and AFL fuzz on the CLI.

Fixes #738.

PR #744, opened 2019-07-01
FAST_DEC_LOOP: Performance improvements.

parheliamm:
@Cyan4973 :
Sorry for bother you many times.

This pull request is based on previous commit e72d44230 and include 2 changes for decompress performance improvements.
1. removed needless condition check because of e72d44230 introduced.
2. intodroduce built-in mempy for 16 bytes if offset >= 16 both on X86 and ARM64.
3. My verification is based on X86 and ARM64 with GCC 8.x. MSM8996 also introduced for verification.

detailed test result is include in commit message.

Please review the test result and commits.

Any feedback would be appreciate.

Chenxi

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1710,7 +1710,6 @@ LZ4_decompress_generic(
 
             if (length == ML_MASK) {
               variable_length_error error = ok;
-              if ((checkOffset) && (unlikely(match + dictSize < lowPrefix))) { goto _output_error; } /* Error : offset outside buffers */
               length += read_variable_length(&ip, iend - LASTLITERALS + 1, endOnInput, 0, &error);
               if (error != ok) { goto _output_error; }
                 if ((safeDecode) && unlikely((uptrval)(op)+length<(uptrval)op)) { goto _output_error; } /* overflow detection */
@@ -1726,7 +1725,16 @@ LZ4_decompress_generic(
 
                 /* Fastpath check: Avoids a branch in LZ4_wildCopy32 if true */
                 if ((dict == withPrefix64k) || (match >= lowPrefix)) {
-                    if (offset >= 8) {
+                    if (offset >= 16) {
+                        assert(match >= lowPrefix);
+                        assert(match <= op);
+                        assert(op + 18 <= oend);
+
+                        memcpy(op, match, 16);
+                        memcpy(op+16, match+16, 2);
+                        op += length;
+                        continue;
+                     } else if (offset >= 8) {
                         assert(match >= lowPrefix);
                         assert(match <= op);
                         assert(op + 18 <= oend);


Cyan4973:
Thanks @parheliamm.

I haven't yet tested this patch's performance, but reading it (it's short enough), I have some comments : 
- It seems to bundle 2 different changes, which both have a chance to improve performance, yet are independent. Maybe it's better to separate them into 2 patches, in order to better evaluate the benefit of each in isolation.
- The first change is a removal of a check, on the ground that it becomes useless. This seems sensible, and from a performance perspective, an almost certain performance benefit. However, we just discovered a vulnerability introduced by such a strategy from an earlier patch, hence we'd better make sure the check is indeed not needed. A typical strategy is to `assert()` it instead. You can omit the `assert()` if : 
    + The condition is obviously guaranteed to be true
    + The condition is obviously not needed
    + The condition is redundant with a prior/later test which checks the same thing (kind of special case of previous points)
    + For any of the above conditions, please insert a code comment, either in code or in commit message, instead of the `assert()`. Unfortunately, the default "diff view" of github doesn't make such conclusion obvious, because it only shows a few lines around the diff itself.  Some conditions need a larger scope to be appreciated. This is one of them.
- The second change is a copy-by-16 branch. This one looks more like a bet, on both `offset` values distribution and target compiler/cpu performance regarding copy-by-16 performance and cost of an additional branch. An unfavorable convoluted case is when there are a lot of offsets in the [8-15] range, and a lot of offsets >= 16, in which case the branch will be highly unpredictable, and it's almost certain to end badly due to pipeline flushes. But hopefully, it can be argued that such case is probably rare, and that offsets >= 16 are generally much more common. Yet, rare is not "never", and now it also depends on the cost of such a branch miss. Hence it might actually be detrimental in some cases. It would be better to analyze these effects in isolation, in order to not confuse challenges specific to this change with more straightforward benefits from the previous one.

parheliamm:
Hi @Cyan4973 :
Sorry for the short information in pull request. I put the detail information to the commit message.
For the 1st change:

> The first change is a removal of a check, on the ground that it becomes useless. 

The background is below:
Previous commit e72d442 (#740 )move the check offset form line 1876 to 1898.
So we could removed needless check offset in FAST_DEC_LOOP. Because if 
If https://github.com/lz4/lz4/blob/dev/lib/lz4.c#L1718 condition check is false, the checkoffset will leveraged at https://github.com/lz4/lz4/blob/dev/lib/lz4.c#L1898
Based on abvoe, we could say https://github.com/lz4/lz4/blob/dev/lib/lz4.c#L1713 is needless or duplicate.

That is my 1st change's idea.

>The second change is a copy-by-16 branch. 

Yup, it is a bet, based on test sample silesia, the offset >=16 is more than [8,15]. I will submit the hit/miss data later.
You are correct if offset is between 8 to 15, the performance is worse than vanilla beause of branch miss.
However, I didn't found this kind of test sample. I will double check more today. If you have this kind of samples, would you please share with me ? That would be appreciate.

Finally, I will sperate this change to 2 commits and create a new request.

Chenxi

Cyan4973:
> You are correct if offset is between 8 to 15, the performance is worse than vanilla beause of branch miss.
However, I didn't found this kind of test sample. I will double check more today. If you have this kind of samples, would you please share with me ? 

I don't think this situation is common. I believe that offset in the [8;15] range are pretty rare, save very specific cases, so it's a reasonable bet.

It was more a trope, to underline that, compared to previous change, this one is more variable, and less certain to be a performance win. Hence it shouldn't be mixed.

To be fair, I'm more concerned by the impact on the branch strategy, which will be selected by the compiler. But we'll see, we'll run some tests and measure.

parheliamm:
> I don't think this situation is common. I believe that offset in the [8;15] range are pretty rare, save very specific cases, so it's a reasonable bet.

Yup, I did a quick test 5 mins ago, the silesia test sample proved that [8;15] is much less than 16.
I will submit copy 16 commit right now and have a quick test on X86

parheliamm:
Hi @Cyan4973 :
I will close this ticket and subit a new one for discussion.
Stay tuned.

One more udpated info.
For branch strategy, I found below change might be better
+                    if (likely(offset >= 16)) {
I am running benchmark now...

Chenxi.

parheliamm:
Close this ticket and submit a new one to dicuss.

Cyan4973:
Strangely enough, when I tried removing the `checkoffset` test, on the ground that it's redundant,
it resulted in a loss of performance in a lot of scenarios.
Small wins for some others.

It doesn't make sense. Removing a useless test should be positive or neutral.

I'm suspecting a random impact on code layout, which shows how modern cpus are sensitive to that kind of things. Unfortunately, this impact is relatively uncontrollable from C source code, which is a problem.

It also shows that one should be cautious with performance gains : sometimes, it's just random luck on instruction alignment, and it's not necessarily reproducible with a different compiler, or when associated with another small change.

PR #752, opened 2019-07-13
 [ossfuzz] Improve the fuzzers

terrelln:
* Run more decompression variants
* Round trip the compression fuzzer and do partial decompression as well
* Add a compression fuzzer that compresses into a smaller output buffer
  and test the destSize variant

These fuzzers caught 2 bugs that were fixed in the previous commit.
* Input buffer over-read in partial decompress
* Partial decompress fails if output size is 0


lib/lz4.c
```diff
@@ -1852,19 +1856,19 @@ LZ4_decompress_generic(
             {
                 if (partialDecoding) {
                     if (cpy > oend) { cpy = oend; assert(op<=oend); length = (size_t)(oend-op); }  /* Partial decoding : stop in the middle of literal segment */
-                    if ((endOnInput) && (ip+length > iend)) { goto _output_error; } /* Error : read attempt beyond end of input buffer */
+                    if ((endOnInput) && (ip+length>iend)) { goto _output_error; } /* Error : read attempt beyond end of input buffer */
+                    if ((endOnInput) && (cpy != oend) && (ip+length>iend-(2+1+LASTLITERALS)) && (ip+length != iend)) { goto _output_error; } /* Error : input must be consumed */
```
terrelln: I believe this is correct, but maybe it could be written better? Please carefully check this fix, since I'm not 100% confident in it.

It passes the round trip and decompression fuzzers, which gives me some confidence that its right.

ossfuzz/compress_fuzzer.c
```diff
@@ -1,25 +1,51 @@
+/**
+ * This fuzz target attempts to comprss the fuzzed data with the simple
```
Cyan4973: nit: "compress"

PR #753, opened 2019-07-15
bugfix: correctly control the offset < LZ4_DISTANCE_MAX,when change t…

Hitatm:
when change the value of  LZ4_DISTANCE_MAX macro，and set -DLZ4_DEBUG=1 during compile lz4；use the compression parameter  -  [lz4hc,9],it will trigger an assert ，because it can not correctly contrl offset smaller than the value of LZ4_DISTANCE_MAX.   this bug can be fixed by set lowestMatchIndex according to LZ4_DISTANCE_MAX ,but not fixed 64K.


Cyan4973:
Yes, that's a sensible change.

Maybe we should add some tests with custom `LZ4_DISTANCE_MAX`, to ensure it's properly applied.

Hitatm:
Thx for your timely response and I saw your other changes，this looks much better

PR #755, opened 2019-07-15
ensure conformance with custom LZ4_DISTANCE_MAX

Cyan4973:
Follow up from #753 : 

It's now possible to select a custom `LZ4_DISTANCE_MAX` at compile time,
provided it's `<= 65535`.

However, in some cases (compression using `byU16` mode),
the custom distance wasn't always respected,
as offsets used to be implied as "in range" in this mode, and therefore wasn't checked.

Added a distance check for this case.
The check is only active when `LZ4_DISTANCE_MAX < 65535`.
Otherwise, it's a no-op (default).

Also : added a new TravisCI test which ensures that
custom `LZ4_DISTANCE_MAX` compiles and compresses correctly,
relying on `assert()` to find outsized offsets.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,10 @@ matrix:
       script:
         - CC=clang MOREFLAGS=-fsanitize=address make -C tests test-frametest test-fuzzer
 
+    - name: Custom LZ4_DISTANCE_MAX
+      script:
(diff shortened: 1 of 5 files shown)


Cyan4973:
> Does the test also catch the LZ4HC bug?

This is the standard `make check`, which also includes a few HC tests, invoking typically `lz4 -9`.

PR #756, opened 2019-07-17
[LZ4_compress_destSize + multi-blocks streaming] Fix rare data corruption bug

terrelln:
* Allow 2 extra match code bytes (line 1033).
* Fix a rare data corruption bug in LZ4_compress_destSize

When we had to shorten the match to fit in the output buffer, and the new match is before the `ip` where we found the match because of the "catch up" there will be positions in the table that are beyond the new `ip`. We must zero those positions so we don't corrupt the hash table.

Credit to OSS-Fuzz

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1027,7 +1027,7 @@ _next_match:
             }
 
             if ((outputDirective) &&    /* Check output buffer overflow */
-                (unlikely(op + (1 + LASTLITERALS) + (matchCode>>8) > olimit)) ) {
+                (unlikely(op + (1 + LASTLITERALS) + (matchCode+240)/255 > olimit)) ) {
                 if (outputDirective == fillOutput) {
                     /* Match description too long : reduce it */
                     U32 newMatchCode = 15 /* in token */ - 1 /* to avoid needing a zero byte */ + ((U32)(olimit - op) - 1 - LASTLITERALS) * 255;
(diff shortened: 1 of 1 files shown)


lib/lz4.c
```diff
@@ -1038,6 +1066,8 @@ LZ4_FORCE_INLINE int LZ4_compress_generic(
             } else
                 *token += (BYTE)(matchCode);
         }
+        /* Ensure we have enough space for the last literals. */
+        assert(!(outputDirective == fillOutput && op + 1 + LASTLITERALS > olimit));
```
terrelln: This assert isn't quite correct (or is exposing a bug) so I'll look into it.

terrelln: I believe it is triggered because the condition on line 1030 isn't quite right. As I understand it `matchCode>>8` has an off by one error in the division. Also, we could need one extra zero byte.

I think we need something like `(matchCode+240)/255`.

Cyan4973:
Looks good to me !
It's great you made this new test tool able to stress `LZ4_compress_destSize()` more thoroughly.

lib/lz4.c
terrelln: This assert was failing when `matchCode == 7665`. `matchCode >> 8 == 29`, but it actually needs 31 bytes since `(matchCode + 240) / 255 == 31` (the extra byte is needed because 7665 - 15 is a multiple of 255).

PR #757, opened 2019-07-18
[fuzz] Add streaming round trip fuzzer

terrelln:
Fuzzers
* Streaming compression with prefix
* Streaming compression with extDict
* Streaming compression with loaded dictionary
* Streaming compression with attached dictionary
* Streaming compression with all of the above in HC mode
* LZ4HC round trip fuzzer
* LZ4HC compress fuzzer

Fixes:
* Fixed `LZ4_attach_dictionary()` with an empty dictionary. It would find matches **beyond** the end of the dictionary causing asserts to fail. CC @felixhandte.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1429,7 +1429,7 @@ void LZ4_attach_dictionary(LZ4_stream_t* workingStream, const LZ4_stream_t* dict
      */
     LZ4_resetStream_fast(workingStream);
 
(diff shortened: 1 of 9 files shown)


terrelln:
The HC streaming round trip fuzzer is running into a bug, so we can't merge this yet.

terrelln:
Still buggy in in HC with an attached dictionary

PR #758, opened 2019-07-18
Added `BUNDLE	DESTINATION`

dooxe:
Added `BUNDLE	DESTINATION` to target `${LZ4_PROGRAMS_BUILT}` in the `CMakeLists.txt`.
Like that, it is compatible with newer versions of cmake (tested on 3.14 and 3.15).


Cyan4973:
Is it still compatible with `cmake` `v2.8.6` ?

dooxe:
I do not have an environment to test, but the `BUNDLE DESTINATION` option is already present in the cmake 2.8.6 documentation: 
https://cmake.org/cmake/help/v2.8.6/cmake.html#command:install

Cyan4973:
Thanks @dooxe !

dooxe:
My pleasure :)

PR #766, opened 2019-07-23
Add option '--best' to lz4(1)

Low-power:
This option exists in **gzip(1)**, **bzip2(1)**, **lzma(1)** and similar commands.


Cyan4973:
Thanks @Low-power !
I believe it's a nice addition to the CLI

PR #768, opened 2019-07-24
[LZ4HC] Speed up pattern compression with external dictionary

### LINKED ISSUE #761: [LZ4HC] Speed up pattern compression with external dictionary
LZ4HC detects patterns in the source data, like a huge run of 0s, and instead of becoming an N^2 algorithm, special cases it. However, this special case is only applied to matches in the current segment. Matches into the external dictionary skip this optimization. See https://github.com/lz4/lz4/blob/dev/lib/lz4hc.c#L317.

This task is to extend this optimization to apply to matches in the external dictionary.

terrelln:
Fixes #761.

Tested by running the ossfuzz round trip fuzzers overnight without any errors (but they caught bugs during development and in the initial version quickly).

### DIFF
--- a/lib/lz4hc.c
+++ b/lib/lz4hc.c
@@ -151,6 +151,19 @@ int LZ4HC_countBack(const BYTE* const ip, const BYTE* const match,
     return back;
 }
 
+#if defined(_MSC_VER)
+#  define LZ4HC_rotl32(x,r) _rotl(x,r)
+#else
+#  define LZ4HC_rotl32(x,r) ((x << r) | (x >> (32 - r)))
+#endif
+
+
+static U32 LZ4HC_rotatePattern(size_t const length, U32 const pattern)
+{
+    size_t const bitsToRotate = (length & (sizeof(pattern) - 1)) << 3;
+    return LZ4HC_rotl32(pattern, (int)bitsToRotate);
+}
+
 /* LZ4HC_countPattern() :
  * pattern32 must be a sample of repetitive pattern of length 1, 2 or 4 (but not 3!) */
 static unsigned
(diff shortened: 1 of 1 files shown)


lib/lz4hc.c
```diff
@@ -313,34 +326,41 @@ LZ4HC_InsertAndGetWiderMatch (
...
+                        const BYTE* const dictStart = dictBase + hc4->lowLimit;
+                        const BYTE* const iLimit = extDict ? dictBase + dictLimit : iHighLimit;
+                        size_t forwardPatternLength = LZ4HC_countPattern(matchPtr+sizeof(pattern), iLimit, pattern) + sizeof(pattern);
+                        if (extDict && ip + forwardPatternLength == iLimit) {
+                            U32 const rotatedPattern = LZ4HC_rotatePattern(forwardPatternLength, pattern);
+                            forwardPatternLength += LZ4HC_countPattern(lowPrefixPtr, iHighLimit, rotatedPattern);
+                        }
+                        {   const BYTE* const lowestMatchPtr = extDict ? dictStart : lowPrefixPtr;
+                            size_t const backLengthRaw = LZ4HC_reverseCountPattern(matchPtr, lowestMatchPtr, pattern);
+                            size_t const backLength = matchCandidateIdx - MAX(matchCandidateIdx - (U32)backLengthRaw, lowestMatchIndex);
```
terrelln: This is the last changed line, after this everything is the same, just indented one more level.

Cyan4973:
Is there some way to measure the speed improvement ?

terrelln:
I'll measure the speed on zeros and normal data

terrelln:
```
make clean round_trip_stream_fuzzer DEBUGLEVEL=0
time ./round_trip_stream_fuzzer clusterfuzz-testcase-minimized-round_trip_stream_fuzzer-5658930090344448
```

This file (attached) takes 9 seconds to compress before the patch and 0.1 seconds to compress after.

[clusterfuzz-testcase-minimized-round_trip_stream_fuzzer-5658930090344448.zip](https://github.com/lz4/lz4/files/3428652/clusterfuzz-testcase-minimized-round_trip_stream_fuzzer-5658930090344448.zip)

Also the following command takes 6 seconds on dev and 0.01 seconds after this patch.

```
time head -c 8M /dev/zero | ./lz4 -BD -12 > /dev/null
```

I don't see a noticeable speed difference when running the benchmark tool at level 12 with and without `-BD` before and after the patch.

Cyan4973:
Excellent, thanks @terrelln !

PR #769, opened 2019-07-30
[lz4] Fixed conditional compilation when using memory stack option (L…

P1119r1m:
[lz4] Fixed conditional compilation when using memory stack option (LZ4_HEAPMODE = 0)

* When LZ4_HEAPMODE is equal to 0, several functions use missing malloc/calloc/free functions

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1326,7 +1326,13 @@ int LZ4_compress_destSize(const char* src, char* dst, int* srcSizePtr, int targe
 
 LZ4_stream_t* LZ4_createStream(void)
 {
+#if (LZ4_HEAPMODE)
     LZ4_stream_t* const lz4s = (LZ4_stream_t*)ALLOC(sizeof(LZ4_stream_t));
+#else
+    static LZ4_stream_t lz4sBody;
+    LZ4_stream_t* lz4s = &lz4sBody;
+#endif
+
     LZ4_STATIC_ASSERT(LZ4_STREAMSIZE >= sizeof(LZ4_stream_t_internal));    /* A compilation error here means LZ4_STREAMSIZE is not large enough */
     DEBUGLOG(4, "LZ4_createStream %p", lz4s);
     if (lz4s == NULL) return NULL;
@@ -1374,7 +1380,10 @@ int LZ4_freeStream (LZ4_stream_t* LZ4_stream)
 {
     if (!LZ4_stream) return 0;   /* support free on NULL */
     DEBUGLOG(5, "LZ4_freeStream %p", LZ4_stream);
+#if (LZ4_HEAPMODE)
     FREEMEM(LZ4_stream);
+#endif
+
     return (0);
 }
 
@@ -2155,7 +2164,14 @@ int LZ4_decompress_fast_doubleDict(const char* source, char* dest, int originalS
 
 LZ4_streamDecode_t* LZ4_createStreamDecode(void)
 {
+#if (LZ4_HEAPMODE)
     LZ4_streamDecode_t* lz4s = (LZ4_streamDecode_t*) ALLOC_AND_ZERO(sizeof(LZ4_streamDecode_t));
+    if (lz4s == NULL) return NULL;
+#else
+    static LZ4_streamDecode_t lz4sBody;
+    LZ4_streamDecode_t* const lz4s = &lz4sBody;
+#endif
+
     LZ4_STATIC_ASSERT(LZ4_STREAMDECODESIZE >= sizeof(LZ4_streamDecode_t_internal));    /* A compilation error here means LZ4_STREAMDECODESIZE is not large enough */
     return lz4s;
 }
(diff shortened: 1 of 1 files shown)


lib/lz4.c
```diff
@@ -1326,7 +1326,13 @@ int LZ4_compress_destSize(const char* src, char* dst, int* srcSizePtr, int targe
 
 LZ4_stream_t* LZ4_createStream(void)
 {
+#if (LZ4_HEAPMODE)
     LZ4_stream_t* const lz4s = (LZ4_stream_t*)ALLOC(sizeof(LZ4_stream_t));
+#else
+    static LZ4_stream_t lz4sBody;
```
Cyan4973: `static` is unlikely to be correct here.
This will make the function impossible to multi-thread.

lib/lz4.c
```diff
@@ -2155,15 +2164,24 @@ int LZ4_decompress_fast_doubleDict(const char* source, char* dest, int originalS
 
 LZ4_streamDecode_t* LZ4_createStreamDecode(void)
 {
+#if (LZ4_HEAPMODE)
     LZ4_streamDecode_t* lz4s = (LZ4_streamDecode_t*) ALLOC_AND_ZERO(sizeof(LZ4_streamDecode_t));
+    if (lz4s == NULL) return NULL;
+#else
+    static LZ4_streamDecode_t lz4sBody;
```
Cyan4973: Same issue with `static` here

Cyan4973:
`HEAPMODE` vs `STACKMODE` : 
 this must be seen as an optimization, without ever modifying the semantic of the function.

It's only applicable when using one-shot compression functions without explicit context,
`LZ4_compress_default()` being a good example.

However, one cannot change the meaning of `LZ4_createStream()` to transparently deliver a static buffer instead of a dynamic one. It makes it impossible to use in a multi-thread environment, or even just with multiple contexts ! `LZ4_createStream()` is clearly a constructor, followed by a corresponding destructor, this logic _cannot_ change.

Is the issue about compiling the source code without ever invoking `malloc/calloc/free` ?
In which case, the currently available solution is to modify this set of macros : https://github.com/lz4/lz4/blob/dev/lib/lz4.c#L176 .

P1119r1m:
> 
> 
> `HEAPMODE` vs `STACKMODE` :
> this must be seen as an optimization, without ever modifying the semantic of the function.
> 
> It's only applicable when using one-shot compression functions without explicit context,
> `LZ4_compress_default()` being a good example.
> 
> However, one cannot change the meaning of `LZ4_createStream()` to transparently deliver a static buffer instead of a dynamic one. It makes it impossible to use in a multi-thread environment, or even just with multiple contexts ! `LZ4_createStream()` is clearly a constructor, followed by a corresponding destructor, this logic _cannot_ change.
> 
> Is the issue about compiling the source code without ever invoking `malloc/calloc/free` ?
> In which case, the currently available solution is to modify this set of macros : https://github.com/lz4/lz4/blob/dev/lib/lz4.c#L176 .

In general, I agree with you, but on embedded devices there is no alloc/calloc/free routines, so yes " Issue is about compiling the source code without ever invoking `malloc/calloc/free`"

So, generally we need to find some solution to satisfy embedded applications that doesn't have memory allocator too.

Anyway, current solution works for single context application and it is better than nothing IMHO.
What do you think about compile-time warning message?

felixhandte:
I don't think I agree that it's better than nothing. Doing so would enormously violate the principle of least surprise. If those functions cannot be made to work on the target platform, they should fail loudly, not silently.

I would recommend, as an alternative, defining your own stubs:

```
void* stub_malloc(size_t) {
  return NULL;
}

void* stub_calloc(size_t, size_t) {
  return NULL;
}

void stub_free(void*) {}
```

And building LZ4 with something like "`-Dmalloc=stub_malloc -Dcalloc=stub_calloc -Dfree=stub_free`". LZ4 could be improved so that it doesn't `#include <stdlib.h>` at all if those symbols are already defined. I'd be happy to submit a patch that does that.

P1119r1m:
Frankly, I've also changed my mind about "... better than nothing" :)
Now I think it is "better than nothing" if to add definition (for conditional compilation) like "#define SINGLE_THREAD_NO_HEAP".
I'm not sure if stub functions will solve the problem. They will do nothing, but must allocate/free memory.
One more solution is to create simple allocator, but this will lead to the increase of memory size.
Moreover it looks ugly. :(
So, I'm looking for another ideas.

felixhandte:
What do you mean "they will do nothing, but must allocate/free memory"?

P1119r1m:
I was talking about stub functions you've gave before. They are just dummy functions, but must be able to allocate/free memory.

Cyan4973:
I believe this topic needs a more detailed discussion, 
about the targeted needs, potential solutions, and risks,
before moving to a code implementation.

I'm sympathetic to the need to compile LZ4 in a freestanding environment without `stdlib`,
this is a common use case, which has already been deployed multiple times,
however, this must be done in a way which doesn't break compatibility with the rest of the ecosystem.

If all one need is a statically allocated space for the context,
the current solution is : do not use allocator function.
Just statically allocate a context.
That's why `LZ4_stream_t` and `LZ4_streamDecode_t` are published.

Then, there is still a need to link the library, but `malloc`/`free` stubs are a relatively common issue with freestanding environments. If the experience can be improved, then okay, let's discuss it and do it.

PR #770, opened 2019-07-31
util.h: Remove deprecated utime for non-Windows

neheb:
utime was deprecated in POSIX 2008.

### DIFF
--- a/programs/util.h
+++ b/programs/util.h
@@ -37,12 +37,17 @@ extern "C" {
 #include <assert.h>
 #include <sys/types.h>    /* stat, utime */
 #include <sys/stat.h>     /* stat */
-#if defined(_MSC_VER)
+#if defined(_WIN32)
 #  include <sys/utime.h>  /* utime */
 #  include <io.h>         /* _chmod */
 #else
 #  include <unistd.h>     /* chown, stat */
+# if PLATFORM_POSIX_VERSION < 200809L
 #  include <utime.h>      /* utime */
+# else
+#  include <fcntl.h>      /* AT_FDCWD */
+#  include <sys/stat.h>   /* for utimensat */
+# endif
 #endif
 #include <time.h>         /* time */
 #include <limits.h>       /* INT_MAX */
@@ -287,14 +292,23 @@ UTIL_STATIC int UTIL_isRegFile(const char* infilename);
 UTIL_STATIC int UTIL_setFileStat(const char *filename, stat_t *statbuf)
 {
     int res = 0;
-    struct utimbuf timebuf;
 
     if (!UTIL_isRegFile(filename))
         return -1;
 
-    timebuf.actime = time(NULL);
-    timebuf.modtime = statbuf->st_mtime;
-    res += utime(filename, &timebuf);  /* set access and modification times */
+    {
+#if defined(_WIN32) || (PLATFORM_POSIX_VERSION < 200809L)
+        struct utimbuf timebuf;
+        timebuf.actime = time(NULL);
+        timebuf.modtime = statbuf->st_mtime;
+        res += utime(filename, &timebuf);  /* set access and modification times */
+#else
+        struct timespec timebuf[2] = {};
+        timebuf[0].tv_nsec = UTIME_NOW;
+        timebuf[1].tv_sec = statbuf->st_mtime;
+        res += utimensat(AT_FDCWD, filename, timebuf, 0);  /* set access and modification times */
+#endif
+    }
 
 #if !defined(_WIN32)
     res += chown(filename, statbuf->st_uid, statbuf->st_gid);  /* Copy ownership */
--- a/programs/platform.h
+++ b/programs/platform.h
@@ -86,7 +86,7 @@ extern "C" {
 #  else
 #    if defined(__linux__) || defined(__linux)
 #      ifndef _POSIX_C_SOURCE
-#        define _POSIX_C_SOURCE 200112L  /* use feature test macro */
+#        define _POSIX_C_SOURCE 200809L  /* use feature test macro */
 #      endif
 #    endif
 #    include <unistd.h>  /* declares _POSIX_VERSION */


Cyan4973:
Same thing here : we are planning a new `v1.9.2` release soon, and this is a bugfix release, so we want to minimize the nb of variables, and this change looks like it could impact some platforms.
So we'll consider it again after the release.

Among the discussion points : what happens to posix platforms which are < 2008 ?

programs/util.h
```diff
@@ -37,12 +37,13 @@ extern "C" {
 #include <assert.h>
 #include <sys/types.h>    /* stat, utime */
 #include <sys/stat.h>     /* stat */
-#if defined(_MSC_VER)
+#if defined(_WIN32)
 #  include <sys/utime.h>  /* utime */
 #  include <io.h>         /* _chmod */
 #else
+#  include <fcntl.h>      /* AT_FDCWD */
+#  include <sys/stat.h>   /* for utimensat */
```
Cyan4973, line 49: What are the minimum requirements to ensure these files can be included without risk ?

neheb:
Updated pull request to match the zstd one. Should work with other platforms as well.

Having said that, some of those look like they can use POSIX 2008 instead of 2001. Talking about this: https://github.com/lz4/lz4/blob/dev/programs/platform.h#L83

neheb:
Actually... I'll just leave it.

```
warning: 'utimensat' is only available on macOS 10.13 or newer [-Wunguarded-availability-new]
```

programs/util.h
```diff
@@ -287,14 +292,24 @@ UTIL_STATIC int UTIL_isRegFile(const char* infilename);
 UTIL_STATIC int UTIL_setFileStat(const char *filename, stat_t *statbuf)
 {
     int res = 0;
+#if defined(_WIN32) || (PLATFORM_POSIX_VERSION < 200809L)
```
Cyan4973: minor: coding style : 

You could create a dedicated block `{ ... }` below
where these variables are declared, alongside the code using them.
Since they are not used outside of the block, it's also great to end their lifetime at block boundary.

The benefit is that changes are well localized in one place.

programs/util.h
Cyan4973: So my understanding is that, 
if `PLATFORM_POSIX_VERSION >= 200809L`,
then there is some kind of "guarantee" that these `#include` _must_ be successful ?
A link to a doc source stating such guarantee would be great.

programs/util.h
```diff
@@ -37,12 +37,17 @@ extern "C" {
 #include <assert.h>
 #include <sys/types.h>    /* stat, utime */
 #include <sys/stat.h>     /* stat */
-#if defined(_MSC_VER)
+#if defined(_WIN32)
 #  include <sys/utime.h>  /* utime */
 #  include <io.h>         /* _chmod */
 #else
 #  include <unistd.h>     /* chown, stat */
+#if PLATFORM_POSIX_VERSION < 200809L
```
Cyan4973: minor : coding style
since this `#if` is nested inside the `#else`, it would be nice to indent it to the right, alongside its content.

Cyan4973:
Let's tackle this PR now that we are past the risky window.
I believe it's almost there, just some minor nit-picking.

The nature of the change is a bit stressful, 
because this kind of modification tend to impact less common platforms,
which are outside of our automated CI test suite, so we are blind.
However, this PR seems reasonably well scoped and guarded,
so I suppose it's worth the risk.

programs/util.h
neheb: I'll indent just the if. I've seen in some other projects ```# if``` works but ```#  if``` does not.

programs/util.h
neheb: The downside is that it will throw C89 warnings.

programs/util.h
neheb: Single space vs. double space. GitHub doesn't show it well it seems.

programs/util.h
neheb: No idea. Here's a list of headers that uClibc-ng optionally removes:

```
Makefile.in:HEADERS_RM- := \
Makefile.in:HEADERS_RM- += sgtty.h
Makefile.in:HEADERS_RM-$(HAVE_SHARED)                    += dlfcn.h bits/dlfcn.h
Makefile.in:HEADERS_RM-$(PTHREADS_DEBUG_SUPPORT)         += thread_db.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_ARGP)                += argp.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_BSD_ERR)             += err.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_CRYPT)               += crypt.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_EPOLL)               += sys/epoll.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_FENV)                += fenv.h bits/fenv.h bits/fenvinline.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_FLOATS)              += complex.h fpu_control.h ieee754.h \
Makefile.in:HEADERS_RM-$(findstring y,$(UCLIBC_HAS_FTW)$(UCLIBC_HAS_NFTW))  += ftw.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_FTS)                 += fts.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_GLIBC_CUSTOM_PRINTF) += printf.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_GLOB)                += glob.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_GNU_ERROR)           += error.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_GETOPT_LONG)         += getopt.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_IPV6)                += netinet/ip6.h netinet/icmp6.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_BACKTRACE)           += execinfo.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_LIBICONV)            += iconv.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_LIBINTL)             += libintl.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_LOCALE)              += bits/uClibc_ctype.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_PTY)                 += pty.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_REALTIME)            += mqueue.h bits/mqueue.h sched.h \
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_REGEX)               += regex.h regexp.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_SHADOW)              += shadow.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_SOCKET)              += sys/socket.h bits/socket.h sys/socketvar.h bits/socket_type.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_SYSLOG)              += syslog.h sys/syslog.h bits/syslog*.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_THREADS)             += *thread*.h semaphore.h \
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_THREADS_NATIVE)      += atomic.h bits/atomic.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_UTMP)               += bits/utmp.h utmp.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_UTMPX)               += bits/utmpx.h utmpx.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_WCHAR)               += wchar.h wctype.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_WORDEXP)             += wordexp.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_XATTR)               += sys/xattr.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_XLOCALE)             += xlocale.h
Makefile.in:HEADERS_RM-$(UCLIBC_LINUX_SPECIFIC)          += \
Makefile.in:HEADERS_RM-$(UCLIBC_SUPPORT_AI_ADDRCONFIG)   += ifaddrs.h
Makefile.in:HEADERS_RM-$(UCLIBC_SV4_DEPRECATED)          += ustat.h sys/ustat.h bits/ustat.h
Makefile.in:HEADERS_RM-$(UCLIBC_SUSV3_LEGACY)            += sys/timeb.h regexp.h
Makefile.in:HEADERS_RM-$(UCLIBC_SUSV4_LEGACY)            += utime.h ucontext.h
Makefile.in:HEADERS_RM-$(UCLIBC_HAS_ADVANCED_REALTIME)   += spawn.h
```

programs/util.h
Cyan4973: Not if you create a nested block.
Declaration of variables are still allowed in `C89` at the top of any sub-nested block.

neheb: ehhh nvm. This project passes -std=c99. Made the change.

neheb:
error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]

I thought this passes -std=c99.

Cyan4973:
`C90` conformity is still an important property of this project.

In this case, it's enough to create a sub-block, so that variable declaration happen at the top of the block.

programs/util.h
neheb: Not sure what you mean. Got an example?

felixhandte:
```
void foo() {
  int a;
  a = 1;
  int b;
  b = 2;
}
```

Fails ANSI C.

```
void foo() {
  int a;
  a = 1;
  {
    int b;
    b = 2;
  }
}
```

This passes, because all declarations are at the beginning of an enclosing scope.

neheb: Used this suggestion.

PR #773, opened 2019-08-06
Make Attaching an Empty Dict Behave the Same as Using it Directly

felixhandte:
When using an empty dictionary, we bail out of loading or attaching it in
ways that leave the working context in potentially slightly different states.
In particular, in some paths, we will cause the currentOffset to be non-zero,
while in others we would allow it to remain 0.

This difference in behavior is perfectly harmless, but in some situations, it
can produce slight differences in the compressed output. For sanity's sake,
we currently try to maintain a strict correspondence between the behavior of
the dict attachment and the dict loading paths. This patch restores them to
behaving identically.

This shouldn't have any negative side-effects, as far as I can tell. When
writing the dict attachment code, I tried to preserve zeroed currentOffsets
when possible, since they benchmarked as very slightly faster. However, the
case of attaching an empty dictionary is probably rare enought that it's
acceptable to minisculely degrade performance in that corner case.

This split in behavior could previously be observed with:

```
make -C tests fuzzer
tests/fuzzer -s240 -t284906 -i284907
tests/fuzzer -s1330 -t3796126 -i3796127
```

These tests now pass.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1429,6 +1429,10 @@ int LZ4_loadDict (LZ4_stream_t* LZ4_dict, const char* dictionary, int dictSize)
 }
 
 void LZ4_attach_dictionary(LZ4_stream_t* workingStream, const LZ4_stream_t* dictionaryStream) {
+    DEBUGLOG(4, "LZ4_attach_dictionary (%p, %p, size %d)",
+        workingStream, dictionaryStream, dictionaryStream != NULL ?
+        dictionaryStream->internal_donotuse.dictSize : 0);
+
     /* Calling LZ4_resetStream_fast() here makes sure that changes will not be
      * erased by subsequent calls to LZ4_resetStream_fast() in case stream was
      * marked as having dirty context, e.g. requiring full reset.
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks for explanation @felixhandte .

Is this issue supposed to be present since a long time ? Or has it been introduced recently ?

The only difference I could spot is : 
`if (dictionary_stream != NULL) {` 
which became
` if (dictionaryStream != NULL && dictionaryStream->internal_donotuse.dictSize > 0) {`
which was introduced in 1f236e0790ab28a223e89bbb8e65d3ed067adc89
likely as a fix from a failed oss-fuzz test.

But it's unclear to me if such a change, which targets empty dictionaries, could have produced this impact.

I tried the same `fuzzer` test on v1.9.1, and it passed flawlessly.
Though it's unclear if using the same `seed` and test nb results in generating the exact same test,
since it's enough to change one previous test in the loop to modify random parameters of later tests in the same loop...

felixhandte:
Yeah, I think that was the culprit. I think these additional changes should fix the test failure.

terrelln:
I'm surprised those tests weren't failing before, since I observed out-of-bounds indices causing assert failures in oss-fuzz when attaching an empty dictionary in oss-fuzz.

Cyan4973:
Let's merge it, so that oss-fuzz can pick it up, and check it's working fine.

PR #774, opened 2019-08-08
FAST_DEC_LOOP: performance improvements on X86/X64

parheliamm:
Hi @Cyan4973:

This pull request include 2 commits.
1. FAST_DEC_LOOP:Handle all offset condition.
2. FAST_DEC_LOOP: skip needless offset check if length == ML_MASK

Here is the benchmark result.

File name | lz4 | Patched#747 | Patched#747+skip offset check
-- | -- | -- | --
dickens | 3231 | 3342 | 3351
mozilla | 3231 | 3272 | 3284
mr | 3719 | 3810 | 3830
nci | 4592 | 4738 | 4700
ooffice | 3265 | 3293 | 3321
osdb | 3325 | 3533 | 3567
reymont | 2833 | 2882 | 2894
samba | 3663 | 3798 | 3804
sao | 4433 | 4431 | 4492
webster | 3143 | 3296 | 3306
xml | 3908 | 4049 | 4054
X-ray | 7500 | 7614 | 7569

Chenxi

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1825,9 +1825,10 @@ LZ4_decompress_generic(
                         continue;
                     }
 #endif
-            }   }
+                }
+                if ((checkOffset) && (unlikely(match + dictSize < lowPrefix))) { goto _output_error; } /* Error : offset outside buffers */
+            }
 
-            if ((checkOffset) && (unlikely(match + dictSize < lowPrefix))) { goto _output_error; } /* Error : offset outside buffers */
             /* match starting within external dictionary */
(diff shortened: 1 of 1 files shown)


Cyan4973:
I've been benchmarking this PR against current `dev` branch,
and it turns out to be detrimental to performance : 
decompression speed goes down from 3330 MB/s to 3270 MB/s.

Now, it's true that this PR is pretty old by now, and that LZ4 code base has evolved a bit since.
But the specific code block targeted is still identical : 
https://github.com/lz4/lz4/blob/dev/lib/lz4.c#L1787
and when applying the patch directly against it, it leads to the same performance drop.

PR #779, opened 2019-08-16
Adding fuzz data producer for uint32 and using in decompress_fuzzer

bimbashrestha:
Consuming bytes from the end of data instead of from the front to prevent "all-in-one" decisions.

### DIFF
--- a/ossfuzz/compress_frame_fuzzer.c
+++ b/ossfuzz/compress_frame_fuzzer.c
@@ -10,14 +10,12 @@
 #include <string.h>
 
 #include "fuzz_helpers.h"
-#include "fuzz_data_producer.h"
 #include "lz4.h"
 #include "lz4frame.h"
 #include "lz4_helpers.h"
 
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
-    FUZZ_dataProducer_t *producer = FUZZ_dataProducer_create(data, size);
     uint32_t seed = FUZZ_seed(&data, &size);
     LZ4F_preferences_t const prefs = FUZZ_randomPreferences(&seed);
     size_t const compressBound = LZ4F_compressFrameBound(size, &prefs);
--- a/ossfuzz/compress_frame_fuzzer.c
+++ b/ossfuzz/compress_frame_fuzzer.c
@@ -10,12 +10,14 @@
 #include <string.h>
 
 #include "fuzz_helpers.h"
+#include "fuzz_data_producer.h"
 #include "lz4.h"
 #include "lz4frame.h"
 #include "lz4_helpers.h"
 
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
+    FUZZ_dataProducer_t *producer = FUZZ_dataProducer_create(data, size);
     uint32_t seed = FUZZ_seed(&data, &size);
     LZ4F_preferences_t const prefs = FUZZ_randomPreferences(&seed);
     size_t const compressBound = LZ4F_compressFrameBound(size, &prefs);
--- a/ossfuzz/decompress_fuzzer.c
+++ b/ossfuzz/decompress_fuzzer.c
@@ -15,7 +15,8 @@
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
     FUZZ_dataProducer_t *producer = FUZZ_dataProducer_create(data, size);
-    size_t const dstCapacity = FUZZ_dataProducer_uint32(producer, 0, 4 * size);
+    size_t const dstCapacity = FUZZ_dataProducer_uint32(
+      producer, 0, 4 * size);
     size_t const smallDictSize = size + 1;
     size_t const largeDictSize = 64 * 1024 - 1;
     size_t const dictSize = MAX(smallDictSize, largeDictSize);
--- a/ossfuzz/decompress_fuzzer.c
+++ b/ossfuzz/decompress_fuzzer.c
@@ -9,13 +9,12 @@
 #include <string.h>
 
 #include "fuzz_helpers.h"
+#include "fuzz_data_producer.h"
 #include "lz4.h"
 
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
-
-    uint32_t seed = FUZZ_seed(&data, &size);
-    size_t const dstCapacity = FUZZ_rand32(&seed, 0, 4 * size);
+    size_t const dstCapacity = FUZZ_produceUint32Range(data, size, 0, 4 * size);
     size_t const smallDictSize = size + 1;
     size_t const largeDictSize = 64 * 1024 - 1;
     size_t const dictSize = MAX(smallDictSize, largeDictSize);
--- a/ossfuzz/decompress_fuzzer.c
+++ b/ossfuzz/decompress_fuzzer.c
@@ -14,7 +14,8 @@
 
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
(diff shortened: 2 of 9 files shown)


ossfuzz/fuzz_data_producer.h
```diff
@@ -0,0 +1,27 @@
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+FUZZ_STATIC uint32_t FUZZ_produceUint32Range(uint8_t *data, size_t size,
```
terrelln: I think we need to make this a stateful struct and take ownership of the buffer. Otherwise two calls to this function will produce the same data. We want to use this API for the rest of the fuzzers too. For example, we'll need to implement [FUZZ_randomPreferences()](https://github.com/lz4/lz4/blob/dev/ossfuzz/lz4_helpers.h#L8) using the `FUZZ_dataProducer_t` instead of the seed.

```c
// Header
typedef struct {
  uint8_t* data;
  size_t size;
} FUZZ_buffer_t;

typedef struct FUZZ_dataProducer_s FUZZ_dataProducer_t;
FUZZ_dataProducer_t* FUZZ_dataProduer_create(uint8_t* data, size_t size);
void FUZZ_dataProducer_free(FUZZ_dataProducer_t* producer);
uint32_t FUZZ_dataProducer_uint32(FUZZ_dataProducer_t* producer, uint32_t min, uint32_t max);
FUZZ_buffer_t FUZZ_dataProducer_remaining(FUZZ_dataProducer_t* producer);
// Whatever other functions you need...

// C file
struct FUZZ_producer_s {
  uint8_t* data;
  size_t size;
};
```

ossfuzz/fuzz_data_producer.h
```diff
@@ -3,8 +3,22 @@
...
+  producer->data = data;
+  producer->size = size;
+  return producer;
+}
+
+void FUZZ_dataProducer_free(FUZZ_dataProducer_t *producer) { free(producer); }
+
+uint32_t FUZZ_dataProducer_uint32(FUZZ_dataProducer_t *producer, uint32_t min,
+                                  uint32_t max) {
   if (min > max) {
```
bimbashrestha: Should I be using an assert instead?

terrelln: Use `FUZZ_ASSERT` from `fuzz_helpers.h`. We want to make sure that we still crash even if `NDEBUG` is defined.

ossfuzz/decompress_fuzzer.c
```diff
@@ -9,13 +9,13 @@
...
+#include "fuzz_data_producer.h"
 #include "lz4.h"
 
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
-
-    uint32_t seed = FUZZ_seed(&data, &size);
-    size_t const dstCapacity = FUZZ_rand32(&seed, 0, 4 * size);
+    FUZZ_dataProducer_t *producer = FUZZ_dataProducer_create(data, size);
+    size_t const dstCapacity = FUZZ_dataProducer_uint32(producer, 0, 4 * size);
```
terrelln: After this we'll want:

```c
    FUZZ_buffer_t const buffer = FUZZ_dataProducer_remaining(producer);
    size_t const smallDictSize = buffer.size + 1;
    // use buffer instead of (data, size)...
```

Right now, we are using the last few bytes twice. Once to get `dstCapacity`, and once as the input to the decompression functions. We should only use each byte in one way. We can do that with a `FUZZ_dataProducer_remaining()` function that returns the unused bytes.

We can use the remaining bytes in each of the `LZ4_decompress*()` functions because they interpret the bytes in the same way, since they all accept the same format.

bimbashrestha: I think only the remaining bytes are being fed into the decompression functions because I change size here: https://github.com/lz4/lz4/pull/779/files/fad8c97532f74d92f6aa4427a739610035fcbbd1..a9ac05645644a0615b558f6ac655c4ae46c4a926#diff-7d201b931de5a277000ee0a58cff72abR29

Or maybe you're just saying it would be clearer to do that in a function FUZZ_dataProducer_remaining()?

ossfuzz/compress_frame_fuzzer.c
```diff
@@ -10,12 +10,14 @@
...
 
 #include "fuzz_helpers.h"
+#include "fuzz_data_producer.h"
 #include "lz4.h"
 #include "lz4frame.h"
 #include "lz4_helpers.h"
 
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
+    FUZZ_dataProducer_t *producer = FUZZ_dataProducer_create(data, size);
```
bimbashrestha: This file should not have been changed. Will do this fuzzer later. I'm going to undo the changes in the next commit.

ossfuzz/fuzz_data_producer.c
```diff
@@ -0,0 +1,32 @@
+#include "fuzz_data_producer.h"
+
```
terrelln: Let's hide the definition of the struct. 

```c
struct FUZZ_dataProducer_s {
  const uint8_t *data;
  size_t size;
};
```

Then where you are using `producer->size`, instead add a function `size_t FUZZ_dataProducer_remainingBytes(FUZZ_dataProducer_t* producer) { return producer->size; }`

ossfuzz/fuzz_data_producer.c
```diff
@@ -0,0 +1,32 @@
...
+FUZZ_dataProducer_t *FUZZ_dataProducer_create(const uint8_t *data, size_t size) {
+  FUZZ_dataProducer_t *producer = malloc(sizeof(FUZZ_dataProducer_t));
```
terrelln: You should always check that the result of `malloc()` is not `NULL`.
```c
FUZZ_ASSERT(producer != NULL);
```

ossfuzz/fuzz_data_producer.h
```diff
@@ -3,40 +3,16 @@
 #include <stdio.h>
 #include <stdlib.h>
 
+#include "fuzz_helpers.h"
+
 typedef struct {
```
terrelln: Then here, you can just do:

```c
typedef struct FUZZ_dataProducer_s FUZZ_dataProducer_t;
```

This is how we hide struct internals in C where we don't have private members.

ossfuzz/fuzz_data_producer.h
```diff
@@ -3,40 +3,16 @@
...
-  producer->data = data;
-  producer->size = size;
-  return producer;
-}
+FUZZ_dataProducer_t *FUZZ_dataProducer_create(const uint8_t *data, size_t size);
 
-void FUZZ_dataProducer_free(FUZZ_dataProducer_t *producer) { free(producer); }
+void FUZZ_dataProducer_free(FUZZ_dataProducer_t *producer);
 
 uint32_t FUZZ_dataProducer_uint32(FUZZ_dataProducer_t *producer, uint32_t min,
```
terrelln: Can you add documentation to these functions?

ossfuzz/decompress_fuzzer.c
terrelln: Good point, I missed that.

ossfuzz/compress_frame_fuzzer.c
```diff
@@ -10,14 +10,12 @@
 #include <string.h>
 
 #include "fuzz_helpers.h"
-#include "fuzz_data_producer.h"
 #include "lz4.h"
 #include "lz4frame.h"
 #include "lz4_helpers.h"
```
bimbashrestha: Just removing accidentally made changes here:)

Cyan4973:
Thanks @bimbashrestha !

PR #785, opened 2019-08-30
Making fuzzers use dataProducer api instead of random seed for decisions

bimbashrestha:
Had planned on moving the remaining fuzzers over to consuming bytes from the end of data for decisions. Doing so now. Previous PR: https://github.com/lz4/lz4/pull/779. 

I've added a new dataProducer method (FUZZ_dataProducer_preferences) to get fuzzed preferences using bytes from the end of the stream instead of rand32().

Still have yet to move round_trip_steam_fuzzer.c. Will do that one separately because it's more complex.


Cyan4973 (approved):
LGTM !

terrelln (approved):

PR #789, opened 2019-09-13
Removing size modification after data producer call 

bimbashrestha:
The assert below was failing because the data producer generates partialCapacity between (0, n)  inclusive but then we shrink the size of the data to be n-1 right after so when the data producer selects n as the partialCapacity, the assert fails. 

Of course, the current solution means that we're reusing the data we used to make the decision for the fuzzer. Maybe this is okay? Maybe not? thoughts? 

This means we need to do this for all the other fuzzers that I moved too. I'll do that once I get some feedback on the above decision.


bimbashrestha:
Failing assert

https://github.com/lz4/lz4/blob/dev/ossfuzz/round_trip_fuzzer.c#L45

PR #790, opened 2019-09-13
Separating the seed generation and use in FUZZ_dataProducer api 

bimbashrestha:
The problem: The data producer was used to generate values for buffer sizes. But the data producer also consumes bytes from the data stream, making it smaller. This isn't an issue when values less than the original size are produced but this isn't always the case. 

Solution: Instead of using the value from the producer directly, use that as a seed and then have a secondary function do the actual data generation within the correct range (for example between 0 and the truncated size).

### DIFF
--- a/ossfuzz/compress_frame_fuzzer.c
+++ b/ossfuzz/compress_frame_fuzzer.c
@@ -19,8 +19,10 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
     FUZZ_dataProducer_t *producer = FUZZ_dataProducer_create(data, LZ4_compressBound(size));
     LZ4F_preferences_t const prefs = FUZZ_dataProducer_preferences(producer);
-    size_t const compressBound = LZ4F_compressFrameBound(size, &prefs);
-    size_t const dstCapacity = FUZZ_dataProducer_uint32(producer, 0, compressBound);
+    size_t const compressBound = LZ4F_compressFrameBound(FUZZ_dataProducer_remainingBytes(producer), &prefs);
+    size_t const dstCapacitySeed = FUZZ_dataProducer_uint32_seed(producer, 0, compressBound);
+    size_t const dstCapacity = FUZZ_dataProducer_uint32(dstCapacitySeed,
+        0, FUZZ_dataProducer_remainingBytes(producer));
     char* const dst = (char*)malloc(dstCapacity);
     char* const rt = (char*)malloc(size);
 
--- a/ossfuzz/compress_frame_fuzzer.c
+++ b/ossfuzz/compress_frame_fuzzer.c
@@ -19,13 +19,11 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
     FUZZ_dataProducer_t *producer = FUZZ_dataProducer_create(data, LZ4_compressBound(size));
     LZ4F_preferences_t const prefs = FUZZ_dataProducer_preferences(producer);
-
+    size_t const dstCapacitySeed = FUZZ_dataProducer_retrieve32(producer);
     size = FUZZ_dataProducer_remainingBytes(producer);
+
     size_t const compressBound = LZ4F_compressFrameBound(size, &prefs);
-    size_t const dstCapacitySeed = FUZZ_dataProducer_uint32(producer, 0, compressBound);
-
-    size = FUZZ_dataProducer_remainingBytes(producer);
-    size_t const dstCapacity = FUZZ_getRange_from_uint32(dstCapacitySeed, 0, size);
+    size_t const dstCapacity = FUZZ_getRange_from_uint32(dstCapacitySeed, 0, compressBound);
 
     char* const dst = (char*)malloc(dstCapacity);
     char* const rt = (char*)malloc(size);
--- a/ossfuzz/compress_frame_fuzzer.c
+++ b/ossfuzz/compress_frame_fuzzer.c
@@ -19,18 +19,19 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
     FUZZ_dataProducer_t *producer = FUZZ_dataProducer_create(data, LZ4_compressBound(size));
     LZ4F_preferences_t const prefs = FUZZ_dataProducer_preferences(producer);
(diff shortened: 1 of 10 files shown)


ossfuzz/fuzz_data_producer.c
```diff
@@ -17,7 +17,7 @@ FUZZ_dataProducer_t *FUZZ_dataProducer_create(const uint8_t *data, size_t size)
 
 void FUZZ_dataProducer_free(FUZZ_dataProducer_t *producer) { free(producer); }
 
-uint32_t FUZZ_dataProducer_uint32(FUZZ_dataProducer_t *producer, uint32_t min,
+uint32_t FUZZ_dataProducer_uint32_seed(FUZZ_dataProducer_t *producer, uint32_t min,
```
Cyan4973: API consistency :
I would rather recommend this prototype to be named `FUZZ_dataProducer_uint32()`, because it retrieves a `uint32` from `producer`. This is clear and consistent with `FUZZ_dataProducer_preferences()`, which retrieves `preferences` from `producer`, and `FUZZ_dataProducer_frameInfo()`, which ... well, I guess you get the point.

ossfuzz/fuzz_data_producer.c
```diff
@@ -32,24 +32,36 @@ uint32_t FUZZ_dataProducer_uint32(FUZZ_dataProducer_t *producer, uint32_t min,
...
   }
 
-  if (range == 0xffffffff) {
-    return result;
-  }
+  return result;
+}
 
-  return min + result % (range + 1);
+uint32_t FUZZ_dataProducer_uint32(uint32_t seed, uint32_t min, uint32_t max)
```
Cyan4973: In contrast, this one could be called (for example) `FUZZ_getRange_from_uint32()` or something equivalent.
Note that it doesn't act on any `*producer`.

ossfuzz/fuzz_data_producer.c
```diff
@@ -32,24 +32,36 @@ uint32_t FUZZ_dataProducer_uint32(FUZZ_dataProducer_t *producer, uint32_t min,
...
+{
+    uint32_t range = max - min;
+    if (range == 0xffffffff) {
+      return seed;
+    }
+    return min + seed % (range + 1);
+}
+
+uint32_t FUZZ_dataProducer_uint32NonAdaptive(FUZZ_dataProducer_t* producer,
```
Cyan4973: Same idea, try to preserve API consistency.
What about `FUZZ_dataProducer_range32()` ?

ossfuzz/compress_frame_fuzzer.c
```diff
@@ -19,8 +19,10 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
     FUZZ_dataProducer_t *producer = FUZZ_dataProducer_create(data, LZ4_compressBound(size));
     LZ4F_preferences_t const prefs = FUZZ_dataProducer_preferences(producer);
-    size_t const compressBound = LZ4F_compressFrameBound(size, &prefs);
-    size_t const dstCapacity = FUZZ_dataProducer_uint32(producer, 0, compressBound);
+    size_t const compressBound = LZ4F_compressFrameBound(FUZZ_dataProducer_remainingBytes(producer), &prefs);
+    size_t const dstCapacitySeed = FUZZ_dataProducer_uint32_seed(producer, 0, compressBound);
+    size_t const dstCapacity = FUZZ_dataProducer_uint32(dstCapacitySeed,
+        0, FUZZ_dataProducer_remainingBytes(producer));
```
Cyan4973: minor :
I would rather recommend to move `size = FUZZ_dataProducer_remainingBytes(producer);` up, and use `size` directly. It makes it very clear that `size` is the same value for `dstCapacity` determination and as argument of `LZ4F_compressFrame()`.
In contrast, `FUZZ_dataProducer_remainingBytes(producer)` is a function call, and may (or may not) provide 2 different results, or generate side effects, for 2 consecutive operations.

Cyan4973:
[Naming is the hardest problem in computer science](https://hilton.org.uk/blog/why-naming-things-is-hard).

ossfuzz/fuzz_data_producer.h
```diff
@@ -17,14 +17,14 @@ FUZZ_dataProducer_t *FUZZ_dataProducer_create(const uint8_t *data, size_t size);
 void FUZZ_dataProducer_free(FUZZ_dataProducer_t *producer);
 
 /* Returns a seed value for the function after this one to consume */
-uint32_t FUZZ_dataProducer_uint32_seed(FUZZ_dataProducer_t *producer, uint32_t min,
+uint32_t FUZZ_dataProducer_uint32(FUZZ_dataProducer_t *producer, uint32_t min,
```
Cyan4973: I don't get it,
if all you want is to extract a `uint32` to consume it as a `seed` later on,
why are there `min` and `max` bounds part of the interface ?
This seems the exact role of `FUZZ_dataProducer_range32()`.

ossfuzz/compress_frame_fuzzer.c
```diff
@@ -19,19 +19,20 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
     FUZZ_dataProducer_t *producer = FUZZ_dataProducer_create(data, LZ4_compressBound(size));
     LZ4F_preferences_t const prefs = FUZZ_dataProducer_preferences(producer);
-    size_t const compressBound = LZ4F_compressFrameBound(FUZZ_dataProducer_remainingBytes(producer), &prefs);
-    size_t const dstCapacitySeed = FUZZ_dataProducer_uint32_seed(producer, 0, compressBound);
-    size_t const dstCapacity = FUZZ_dataProducer_uint32(dstCapacitySeed,
-        0, FUZZ_dataProducer_remainingBytes(producer));
+
+    size = FUZZ_dataProducer_remainingBytes(producer);
```
Cyan4973, line 23: `size` should be set only once, and after all extraction of `producer` are completed.
`compressBound` should depend on final `size`, not an intermediate version that will be updated later on.

I suspect the issue is you want to provide `compressBound` to retrieve `dstCapacitySeed` with `FUZZ_dataProducer_uint32()`. See my comment on this function. Without a need to provide `bounds` as parameters, it removes the need to determine `compressBound` before `dstCapacitySeed`.

Cyan4973 (approved):

Issue #792, opened 2019-09-15
spec doesn't say how to handle "x += 255" overflowing


nigeltao:
The spec (`doc/lz4_Block_format.md`) says that a 255 byte means to add that to an accumulator and read another byte. It does not specify overflow behavior: what to do if e.g. decoding 17 million consecutive 0xFF bytes (`17000000 * 255` is around 4.3 billion, which is more than `1 << 32`).

Of course, a reasonable encoder shouldn't do that, but not every encoder is reasonable, and a malicious actor passing around a 17MB file is very feasible. It would be great if the spec and various implementations (whether C, Java, JavaScript, Python etc which all have different 'integer' semantics) could all agree on how to treat this (e.g. is it an error, or should it always be `uint32_t` modular arithmetic)?

It is possibly a security issue if two different programs disagree on the decoding of some LZ4-compressed data, especially if the first program verifies some sort of safety property before passing it on to the second.

If I understand the canonical *implementation* correctly, `read_variable_length` in `lz4.c` uses an `unsigned`. The function does have comments that say "overflow detection", but this checks overflowing the input buffer, not overflowing the accumulator. Technically, `sizeof(unsigned int)` depends on the platform. A mere 260 copies of an 0xFF byte could overflow on a 16-bit platform where `sizeof(int) == 2`.

Cyan4973:
The compression format doesn't define any limit to length fields. Enforcing a limit is considered an implementation limitation.

So, for example, if an encoder considers that a length could be encoded using 64-bit registers, and starts sending a huge length with a huge amount of `255`, while the recipient decoder only supports 16-bit fields for length, and is unable to represent such a big number, then it's up to the decoder to define a behavior for such case. A general idea is that it should state that it's not compatible with the received block.
That does not make the block incorrect : it's a decoder limitation, but the block may still be compliant with `lz4` specification.

I agree that this distinction between valid format and implementation limitation is not present in the document. That's because the initial intention was to make the document about the format only, avoiding to be too reliant on any particular implementation. However, thinking about it now, it might be better for this case to signal that most implementations will necessarily have limits and must be prepared to deal with them.

For reference, the implementation in this repository uses 32-bit fields for length. If it overflows, it will generate a decoding error down the line. The error code is not specific though, it merely tells that a decoding error happened if return value is `< 0`, not which error exactly happened.

[...]

PR #793, opened 2019-09-15
Have read_variable_length use fixed size types

nigeltao:
Otherwise, the output from decoding LZ4-compressed input could be
platform dependent.

Also add a compile-time check to confirm the existing code's assumptions
that, if <stdint.h> isn't used, then sizeof(int) == 4.

Updates #792

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -249,6 +249,10 @@ static int g_debuglog_enable = 1;
   typedef uint64_t U64;
   typedef uintptr_t uptrval;
 #else
+# include <limits.h>
+# if UINT_MAX != 4294967295UL
+#   error "LZ4 code (when not C++ or C99) assumes that sizeof(int) == 4"
+# endif
   typedef unsigned char       BYTE;
   typedef unsigned short      U16;
   typedef unsigned int        U32;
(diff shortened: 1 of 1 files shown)


lib/lz4.c
```diff
@@ -249,6 +249,9 @@ static int g_debuglog_enable = 1;
   typedef uint64_t U64;
   typedef uintptr_t uptrval;
 #else
+#if defined(__SIZEOF_INT__) && (__SIZEOF_INT__ != 4)
```
Cyan4973: Yes, this static test is welcomed.

nit picking here, but I think that `__SIZEOF_INT__` is `gcc`, and not part of the `C` standard. There is no problem if this macro is not defined, your code protects this situation, and just disables this test.

Another way to check the same thing would be to probe that `UINT_MAX`, defined in `<limits.h>`, has value `4294967295`. It's equivalent, but `<limits.h>` is `C89`, hence more portable.

nigeltao: Changed the check to look at UINT_MAX.

Cyan4973 (approved):

Cyan4973:
Thanks @nigeltao !

PR #794, opened 2019-09-23
Using size instead of LZ4_compressBound(size) <- causes heap overflow

bimbashrestha:
Should be the cause of the build failure. Passing LZ4_compressBound(size) makes the data producer try and access things outside the data buffer.

### DIFF
--- a/ossfuzz/compress_frame_fuzzer.c
+++ b/ossfuzz/compress_frame_fuzzer.c
@@ -17,7 +17,7 @@
 
 int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 {
-    FUZZ_dataProducer_t *producer = FUZZ_dataProducer_create(data, LZ4_compressBound(size));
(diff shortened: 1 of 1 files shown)


terrelln (approved):

Cyan4973 (approved):

Cyan4973:
I was thinking :
we currently have a test `oss-fuzz`, which ensures that the fuzz suite can be compiled properly : https://github.com/lz4/lz4/blob/dev/.travis.yml#L214 .

But it doesn't run the scenarios themselves, we have to wait for `oss-fuzz` to pick them up.

I was wondering if we could extend the test, to also do some minimum amount of runtime, just to ensure that tests are running properly. 

The goal isn't to find bugs from fuzzing (this is what `oss-fuzz` is for). The objective is merely to ensure that there is no "simple" runtime errors left in the fuzz suite (such as the one fixed in this PR) that would prevent `oss-fuzz` from working properly.

This could be part of another PR, so that we can merge this one early.

PR #795, opened 2019-09-23
Minor: adding forgotten assert

bimbashrestha:


ossfuzz/fuzz_data_producer.c
```diff
@@ -20,6 +20,7 @@ void FUZZ_dataProducer_free(FUZZ_dataProducer_t *producer) { free(producer); }
 uint32_t FUZZ_dataProducer_retrieve32(FUZZ_dataProducer_t *producer) {
     const uint8_t* data = producer->data;
     const size_t size = producer->size;
+    FUZZ_ASSERT(size >= 0);
```
Cyan4973, line 23: If `size` is of type `size_t`, it necessarily is `>= 0`.

bimbashrestha: Ah I see. okay! closing

PR #796, opened 2019-09-26
meson: move one layer deeper to allow easy construction of a wrap file

jcaesar:
When using meson's wrapdb, it is necessary for the `meson.build` file containing the `project` statement to be on the folder root project. While the file can be copied, this currently requires path adjustments I'd like to avoid with this PR.

See also: this discussion on me trying to add lz4 to wrapdb: https://github.com/mesonbuild/wrapweb/issues/103


jcaesar:
@lzutao: what do you think, is this worth the trouble?

tesuji (approved):
I don't have enough knowledge about how wrap work but if this PR doesn't
cause build failure so why not.

Cyan4973 (approved):

PR #798, opened 2019-10-04
Adding cirrus test for FreeBSD

bimbashrestha:
Gmake test runs on init on my fork: 
https://github.com/bimbashrestha/lz4/pull/3

### DIFF
--- a/tests/test_install.sh
+++ b/tests/test_install.sh
@@ -1,7 +1,14 @@
 #/usr/bin/env sh
 set -e
 
-make="$(MAKE) -C $lz4_root"
+
+make="make -C $lz4_root"
+if [ "$unamestr" = 'Linux' ]; then
+  make="make -C $lz4_root"
+elif [ "$unamestr" = 'FreeBSD' ]; then
+  make="gmake -C $lz4_root"
+fi
+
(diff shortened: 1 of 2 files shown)


tests/test_install.sh
```diff
@@ -1,7 +1,7 @@
 #/usr/bin/env sh
 set -e
 
-make="make -C $lz4_root"
+make="gmake -C $lz4_root"
```
Cyan4973: `gmake` is for FreeBSD, but Linux still uses `make`.

tests/test_install.sh
```diff
@@ -1,7 +1,14 @@
 #/usr/bin/env sh
 set -e
 
+
 make="make -C $lz4_root"
+if [ "$unamestr" = 'Linux' ]; then
```
Cyan4973, line 7: where is defined `$unamestr` ?

bimbashrestha: Ahh good catch. Accidentally removed it before pushing.

Cyan4973 (approved):

Cyan4973:
Thanks @bimbashrestha , everything seems to work fine on Cirrus CI !

PR #808, opened 2019-11-04
Make benchmark compatible with dictionary compression

rkoradi:
Support the -D command line option for running benchmarks. The
benchmark code was slightly restructured to factor out the calls
that need to be different for each benchmark scenario. Since there
are now 4 scenarios (all combinations of fast/HC and with/without
dictionary), the logic was getting somewhat convoluted otherwise.

This was done by extending the compressionParameters struct that
previously contained just a single function pointer. It now
contains 4 function pointers for init/reset/compress/cleanup,
with the related state. The functions get a pointer to the
structure as their first argument (inspired by C++), so that they
can access the state values in the struct.

### DIFF
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -45,17 +45,176 @@
 
 #include "datagen.h"     /* RDG_genBuffer */
 #include "xxhash.h"
+#include "bench.h"
 
-
+#define LZ4_STATIC_LINKING_ONLY
 #include "lz4.h"
-#define COMPRESSOR0 LZ4_compress_local
-static int LZ4_compress_local(const char* src, char* dst, int srcSize, int dstSize, int clevel) {
-  int const acceleration = (clevel < 0) ? -clevel + 1 : 1;
-  return LZ4_compress_fast(src, dst, srcSize, dstSize, acceleration);
-}
+#define LZ4_HC_STATIC_LINKING_ONLY
 #include "lz4hc.h"
-#define COMPRESSOR1 LZ4_compress_HC
-#define DEFAULTCOMPRESSOR COMPRESSOR0
+
+
+/* *************************************
+*  Compression parameters and functions
+***************************************/
+
+struct compressionParameters
+{
+    int cLevel;
+    const char* dictBuf;
+    int dictSize;
+
+    LZ4_stream_t* LZ4_stream;
+    LZ4_stream_t* LZ4_dictStream;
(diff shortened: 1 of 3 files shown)


Cyan4973:
Quick question :
Is there a difference in measured performance
for "normal" compression scenarios
before and after this patch ?

rkoradi:
> Is there a difference in measured performance
> for "normal" compression scenarios
> before and after this patch ?

Yes, I do see differences. Even with using LZ4_attach_dictionary, compression is slower, particularly for smaller file sizes. Not surprisingly, decompression is quite a bit slower (also more pronounced for smaller files and larger dictionaries), since the dictionary is read for each iteration.

I can provide some numbers if you let me know what sizes/options should be tested. I tested with a 64K dictionary and a 450K file, and everything was roughly 5% slower.

Cyan4973:
The question was more about comparison of measured performance
when _not_ using a dictionary.

Cyan4973:
The general concept of providing function pointers in a `struct` payload looks good to me.

Now, we mostly need to care for performance.

Also, the code should be `C90` compatible, I guess the majority of compilation warnings come from this requirement.

rkoradi:
Sorry, I misunderstood the question. The same functions as before are used for the non-dictionary case. So I wouldn't expect any differences there. Comparing the runs (for a 450K file again), there are minor differences that do appear repeatable. For level 3, compression is reported as typically about 1% slower than before the change. For level 1 it tends to be a tiny bit higher, though.

rkoradi:
Finally got all the CI builds running cleanly. Sorry about the multiple attempts.

Cyan4973:
I can confirm a very small compression speed loss when comparing `dev` branch with your PR.
On my test system, it's < 1% at level 1, though well reproducible.

It's probably not a big blocker. But let's use a bit of time to understand why this happens.

rkoradi:
The updated version I just pushed runs at least as fast as the original version for me. It actually ran slightly faster in my test runs.

Based on experimentation, I found that calling LZ4_freeStream/LZ4_freeStreamHC unnecessarily for the cases where no streams are used was correlated to the slowdown. Frankly, I have no idea why, because they do nothing when a null stream is passed in. But with such a small difference, it's difficult to root cause it.

Cyan4973 (approved):

Cyan4973:
Thanks @rkoradi ! Great work and great investigation !

PR #809, opened 2019-11-05
Replace obsolete postgres extension on github pages

zilder:
Hi,

Sorry if this isn't the right place to bring this up. I've noticed that you mentioned one of my postgres extensions on your github pages, [pg_lz4](https://github.com/zilder/pg_lz4). I just wanted to say that this extension was implemented as a demo for the postgres patch that my colleague @ildus was working on and which unfortunately didn't make it into the postgres core. Therefore this extension is not useable with vanilla postgres. I see people keep coming from your page and then get confused because they can't build it.

As an alternative i can suggest you another extension i'm currently working on and which can actually be used with the latest postgres 12 release. This is an append-only compressed pluggable storage, [pg_cryogen](https://github.com/adjust/pg_cryogen). Here is a patch for your gh-pages index.html in case you consider it reasonable replacement.

Thanks!


Cyan4973:
Thanks for notification @zilder .
Yes, it's the right place to bring this up.

PR #810, opened 2019-11-07
bring support for Haiku up to date.

korli:
programs/lz4cli.c: only checks recursive when UTIL_HAS_CREATEFILELIST is defined.

### DIFF
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -658,7 +658,11 @@ int main(int argc, const char** argv)
         if (!output_filename) output_filename = stdoutmark;
     }
     else{
+#ifdef UTIL_HAS_CREATEFILELIST
         if (!recursive && !UTIL_isRegFile(input_filename)) {
+#else
+        if (!UTIL_isRegFile(input_filename)) {
+#endif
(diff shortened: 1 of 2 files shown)


programs/lz4cli.c
```diff
@@ -658,7 +658,11 @@ int main(int argc, const char** argv)
         if (!output_filename) output_filename = stdoutmark;
     }
     else{
+#ifdef UTIL_HAS_CREATEFILELIST
```
Cyan4973, line 661: I'm surprised about this need.
`UTIL_HAS_CREATEFILELIST` is used above to ensure that `recursive` can never be set to `1`. So this particular test should not need to be gated ?

korli: `recursive` is not declared when `UTIL_HAS_CREATEFILELIST` isn't used: see [here](https://github.com/lz4/lz4/blob/dev/programs/lz4cli.c#L330)

Cyan4973: oh , right !
It's designed differently from `zstd`, where `recursive` is always declared, and is just set to `0`, avoiding further preprocessor selection.

Anyway, that's probably a topic for another change.

Cyan4973 (approved):

PR #817, opened 2019-12-03
faster decoding speed with Visual

Cyan4973:
by enabling the fast decoder path.
Visual requires a different set of macro constants to detect x86 / x64.

On my laptop, decoding speed on x64 went up from 3.12 to 3.45 GB/s.
32-bit is less impressive, though still favorable,
with speed increasing from 2.55 to 2.60 GB/s.

So both cases are now enabled.

Suggested by Bartosz Taudul (@wolfpld).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,9 @@ _codelite_lz4/
 bin/
 *.zip
 
+# analyzers
+infer-out
+
 # Mac
 .DS_Store
 *.dSYM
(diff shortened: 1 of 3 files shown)


terrelln (approved):

Cyan4973:
CI found a completely unrelated bug,
in "LZ4_compress_HC_continue with an external dictionary",
that now requires further investigation.

ffulin:
@Cyan4973 - I made this same change locally against 1.9.2, and while preparing the PR I found this was already fixed in head by this change :). I thought I'd share my timings in case they are interesting to you.

I saw:
 - some cases improved across the board at all compression levels (like the numbers reported here)
 - some cases not improved at all
 - some cases, when using high acceleration values, improved by almost 2x!

(all numbers are x64 MSVC 2019 v16.5.1 on Ryzen 3900X. Each test repeated 2000 times)

image

PR #820, opened 2019-12-06
lz4cli: fix build w/out createfilelist

vapier:


Cyan4973:
Thanks @vapier,
the issue was actually fixed in a recently merged patch.

PR #823, opened 2019-12-22
Add aarch64, ppc64le, and s390x to Travis CI for Xenial targets

adamretter:
This PR allows testing on multiple architectures via Travis-CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,13 +48,6 @@ matrix:
       script:
         - CC=clang MOREFLAGS=-fsanitize=address make -C tests test-frametest test-fuzzer
 
-    - name: ASAN tests with fuzzer and frametest (s390x)
-      arch: s390x
-      install:
-        - sudo sysctl -w vm.mmap_min_addr=4096
-      script:
-        - CC=clang MOREFLAGS=-fsanitize=address make -C tests test-frametest test-fuzzer
-
     - name: Custom LZ4_DISTANCE_MAX (amd64)
       arch: amd64
       script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -274,20 +274,5 @@ matrix:
       script:
         - ./ossfuzz/travisoss.sh
 
-    - name: Compile OSS-Fuzz targets (arm64)
-      arch: arm64
-      script:
-        - ./ossfuzz/travisoss.sh
-
-    - name: Compile OSS-Fuzz targets (ppc64le)
-      arch: ppc64le
-      script:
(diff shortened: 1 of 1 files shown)


Cyan4973:
Thanks @adamretter , it's a great add to the test system !

I noticed a few remaining test errors on travis.
I resumed the `ppc64` ones, but they fail the same : 
> An error occurred while generating the build script.

Maybe there is an issue in the .travis file ?

Then, I'm unsure if [this one](https://travis-ci.org/lz4/lz4/jobs/628401861#L44) is revealing of a larger problem.

adamretter:
Regards the:

> An error occurred while generating the build script.

There is so little information that comes back from Travis, it's hard to know what is wrong. We should open a ticket with them I think...

adamretter:
> Then, I'm unsure if this one is revealing of a larger problem.

It looks like maybe ASAN is only supported for ARM64 on Android. Is that how you also interpret the guidance here? - https://github.com/google/sanitizers/wiki/AddressSanitizer#introduction

Cyan4973:
Yes.

It also seems that `ppc64le` support is broken on travisCI since recently.

Maybe we don't need _all_ these tests after all.
Simply remove those that don't pass.

Too long testing time is taking a toll on the developer's feedback loop, so let's not overdo it.

adamretter:
@Cyan4973 did you see something written somewhere about broken ppc64le?

Cyan4973:
Yes, in a thread, on travis support forum.

That being said, the thread concluded with a recent answer stating that the situation is now fixed.
So, `ppc64le` builds should work again.

Cyan4973:
It seems there is a remaining issue with `asan` on `ppc64le` : https://travis-ci.org/lz4/lz4/jobs/633740598#L45

adamretter:
@Cyan4973 Indeed. I am not quite sure where to start with it though... perhaps I should consult the ASAN people?

Cyan4973:
We can probably remove this test, as both `asan` and `ppc64le` are tested separately already.
Maybe just leave a note explaining why this test was removed from the script.

adamretter:
@Cyan4973 I would like to understand why it fails before deciding to remove it ;-)

ojeda:
I would suggest creating an issue for that while adding the rest of the tests to benefit from them in the meantime.

Cyan4973:
It seems this PR was never completed.

I'll try to salvage what I can from it into a feature branch before next release.

Cyan4973:
Superceded by #902

PR #839, opened 2020-02-01
Add new type sptrval (equvalent for ssize_t)

remittor:
It is necessary in some places to replace `int` to `sptrval`

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -240,18 +240,56 @@ static const int LZ4_minLength = (MFLIMIT+1);
 /*-************************************
 *  Types
 **************************************/
-#if defined(__cplusplus) || (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) /* C99 */)
+#if defined(__VMS)
+# /* nothing */
+#elif defined(_MSC_VER)
+# if (_MSC_VER >= 1600) /* MSVC 2010 */
+#  define LZ4_HAVE_STDINT_H
+# endif
+#elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) /* C99 */
+# define LZ4_HAVE_STDINT_H
+#elif defined(__cplusplus)
+# define LZ4_HAVE_STDINT_H
+#else
+# /* nothing */
+#endif
+
+#include <limits.h>
+
+#if defined(LZ4_HAVE_STDINT_H)
 # include <stdint.h>
+# define LZ4_STDINT_H_INCLUDED
+#else
+# if defined(INT_LEAST32_MAX) && defined(INT8_MAX) && defined(INT32_MAX) && defined(INTPTR_MAX)
+#  define LZ4_STDINT_H_INCLUDED   /* stdint.h already included */
+# endif
+#endif
+
+#if defined(SIZE_MAX)
+# define LZ4_SIZE_MAX  SIZE_MAX
+#else
+# if defined(__WORDSIZE) && (__WORDSIZE == 64)
+#  define LZ4_SIZE_MAX  18446744073709551615ULL
+# endif
+# if defined(__WORDSIZE) && (__WORDSIZE == 32)
+#  define LZ4_SIZE_MAX  4294967295UL
+# endif
+#endif
+
+#ifdef LZ4_STDINT_H_INCLUDED
   typedef  uint8_t BYTE;
   typedef uint16_t U16;
   typedef uint32_t U32;
   typedef  int32_t S32;
   typedef uint64_t U64;
   typedef uintptr_t uptrval;
+  typedef  intptr_t sptrval;
 #else
-# include <limits.h>
 # if UINT_MAX != 4294967295UL
 #   error "LZ4 code (when not C++ or C99) assumes that sizeof(int) == 4"
+# endif
+# if !defined(LZ4_SIZE_MAX)
+#   error "LZ4 code (when not C++ or C99) assumes that macro constant SIZE_MAX has been defined"
 # endif
   typedef unsigned char       BYTE;
   typedef unsigned short      U16;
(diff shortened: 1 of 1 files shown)


lib/lz4.c
```diff
@@ -240,25 +240,70 @@ static const int LZ4_minLength = (MFLIMIT+1);
...
+#else
+# if defined(INT_LEAST32_MAX) && defined(INT8_MAX) && defined(INT32_MAX) && defined(INTPTR_MAX)
+#  define LZ4_STDINT_H_INCLUDED   /* stdint.h already included */
+# endif
+#endif
+
+#if defined(SIZE_MAX)
+# define LZ4_SIZE_MAX  SIZE_MAX
+#else
+# if defined(__WORDSIZE) && (__WORDSIZE == 64)
```
Cyan4973, line 271: I guess `__WORDSIZE` is a compiler extension.
Which compilers define it ?

lib/lz4.c
```diff
@@ -240,25 +240,70 @@ static const int LZ4_minLength = (MFLIMIT+1);
...
   typedef uint64_t U64;
   typedef uintptr_t uptrval;
+  typedef  intptr_t sptrval;
 #else
-# include <limits.h>
 # if UINT_MAX != 4294967295UL
 #   error "LZ4 code (when not C++ or C99) assumes that sizeof(int) == 4"
+# endif
+# if !defined(LZ4_SIZE_MAX)
+#   error "LZ4 code (when not C++ or C99) assumes that macro constant SIZE_MAX has been defined"
```
Cyan4973, line 292: It seems this will become an issue.
In essence, `C90` compilers have to support `__WORDSIZE`, or compilation will fail.
This seems a too strong restriction.

lib/lz4.c
```diff
@@ -240,25 +240,70 @@ static const int LZ4_minLength = (MFLIMIT+1);
...
   typedef unsigned int        U32;
   typedef   signed int        S32;
   typedef unsigned long long  U64;
   typedef size_t              uptrval;   /* generally true, except OpenVMS-64 */
+# if defined(ULLONG_MAX) && (LZ4_SIZE_MAX == ULLONG_MAX)
+  typedef   signed long long  sptrval;
+# elif defined(ULONG_MAX) && (LZ4_SIZE_MAX == ULONG_MAX)
+  typedef   signed long       sptrval;
+# else
+#   error "Can't determine limit for size_t"
```
Cyan4973, line 305: I'm confused, is all this about `size_t` or `intptr_t` ?

lib/lz4.c
```diff
@@ -240,25 +240,70 @@ static const int LZ4_minLength = (MFLIMIT+1);
...
+#endif
+
+#ifdef LZ4_STDINT_H_INCLUDED
   typedef  uint8_t BYTE;
   typedef uint16_t U16;
   typedef uint32_t U32;
   typedef  int32_t S32;
   typedef uint64_t U64;
   typedef uintptr_t uptrval;
+  typedef  intptr_t sptrval;
```
Cyan4973, line 286: `sptrval` is declared, but not used.
In which places would it be useful ?

remittor:
> In essence, C90 compilers have to support __WORDSIZE, or compilation will fail.
> This seems a too strong restriction.

It really is.

In vain you made support C90:
https://github.com/lz4/lz4/commit/e185b252f0c62a46929bb3fa3be4ab4d1d15d5d5#diff-75d74151998eff44aa2768946f8d3c0eL153-R169
If the compiler supports `size_t` and `long long`, then it cannot be considered as C90. And the MSVC 2005/2008 compilers support `size_t` and `long long`.

For a very long time, all users of compilers, for which initially do not have the `<stdint.h>` file, themselves put its analogues when compiling.
It looks something like this:
```
#include <my_stdint.h>
#include "external/lz4/lib/lz4.h"
```
Here is a real example:
https://github.com/remittor/paxz.wcx/blob/8c10a4673c4e613383f7c61041e4116007ba032d/src/stdafx.h#L22-L25
https://github.com/remittor/paxz.wcx/tree/8c10a4673c4e613383f7c61041e4116007ba032d/src/win32common

My belief: A true C90 compiler should not support type `long long`!

lib/lz4.c
remittor: A small part of such places:
https://github.com/lz4/lz4/blob/e55548b0bd553cc25d8c0994db6f5d49663fb3f5/lib/lz4hc.c#L144
https://github.com/lz4/lz4/blob/e55548b0bd553cc25d8c0994db6f5d49663fb3f5/lib/lz4hc.c#L317
https://github.com/lz4/lz4/blob/e55548b0bd553cc25d8c0994db6f5d49663fb3f5/lib/lz4hc.c#L615
https://github.com/lz4/lz4/blob/e55548b0bd553cc25d8c0994db6f5d49663fb3f5/lib/lz4hc.c#L647

lib/lz4.c
remittor: Same as for `uptrval`.
https://github.com/lz4/lz4/blob/e55548b0bd553cc25d8c0994db6f5d49663fb3f5/lib/lz4.c#L261

lib/lz4.c
remittor: If you need to do it perfectly, then you need to oblige the developers to include `<stdint.h>` before include `"lz4.h"`

lib/lz4.c
remittor: it's a hack for `gсс` (Travis CI).
In older versions of `gcc` this is not. 
It's much more efficient to use `__LP64__`

lib/lz4.c
remittor: From `gcc` 3.0 up to `gcc` 4.3, the programmer was required to create the `stdint.h` file himself.
And we are trying to cancel this tradition.

remittor:
https://github.com/lz4/lz4/blob/e55548b0bd553cc25d8c0994db6f5d49663fb3f5/lib/lz4.c#L243-L244
This is actually incorrect implementation!
`gcc` from v3.0 to v4.4 may not have a file `stdint.h` (see: https://gcc.gnu.org/c99status.html ). And may be indicated `-std=c99` or using `g++` compiler.

Therefore, according to all the particularities, need to do this: https://github.com/remittor-pr/lz4/blob/e5a01a77da09bdcd5948d785cfbe31f5ad7ab84f/lib/lz4.c#L243-L259

remittor:
Possible rejection of using type `intptr_t`: https://github.com/remittor-pr/lz4/commit/fdf60e2b3f70159efc860e91bc1b232c9f913ebd

PR #841, opened 2020-02-06
Turning on fast dec for visual

bimbashrestha:
Benchmarked compression on Windows vm with ./lz4.exe -b1 -e9

Level 1: +3%
Level 2: +2% 
Level 3: +2% 
Level 4: +3%
Level 5: -1%
Level 6: +0%
Level 7: +6%
Level 8: +6%
Level 9: +5%

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -374,7 +374,7 @@ static const int      dec64table[8] = {0, 0, 0, -1, -4,  1, 2, 3};
 
 
 #ifndef LZ4_FAST_DEC_LOOP
-#  if defined __i386__ || defined _M_IX86 || defined __x86_64__ || defined _M_X64
+#  if defined _MSC_VER || defined __i386__ || defined _M_IX86 || defined __x86_64__ || defined _M_X64
 #    define LZ4_FAST_DEC_LOOP 1
 #  elif defined(__aarch64__) && !defined(__clang__)
      /* On aarch64, we disable this optimization for clang because on certain


lib/lz4.c
```diff
@@ -374,7 +374,7 @@ static const int      dec64table[8] = {0, 0, 0, -1, -4,  1, 2, 3};
 
 
 #ifndef LZ4_FAST_DEC_LOOP
-#  if defined __i386__ || defined _M_IX86 || defined __x86_64__ || defined _M_X64
+#  if defined _MSC_VER || defined __i386__ || defined _M_IX86 || defined __x86_64__ || defined _M_X64
```
Cyan4973, line 377: 2 things :

`_MSC_VER` checks if Visual Studio is being used.

But Visual Studio can be used to compile multiple targets, such as ARM, and the request was limited to enabling it for normal `x86`/`x64` targets only.

Extending beyond that scope is not forbidden, but it's a different topic, and will have to be backed by a dedicated set of benchmarks.

Second, I note that the `defined` line already tests Visual's own macros `_M_X64` and `_M_IX86`.
It means that the task has already been fulfilled : the fast decoder is already unlocked for Visual Studio.

Cyan4973 (changes requested):

Cyan4973:
So apparently the related task has already been completed, it was just a left-over in the system.
Sorry for that. 

Now I must say I'm puzzled by the measured performance improvement, since the outcome of the change is supposed to be a non-event. If it's not, then there is something wrong with the detection macro.

This could be checked quickly by comparing the `md5` checksum of both generated files (with and without the patch, and of course, after removal of the source code from the binary).

PR #843, opened 2020-02-07
Implement -m option with legacy format on cli

filipecalasans:
This PR implements the cli option -m with legacy compression format.


Cyan4973:
Thanks @filipecalasans !

It would great to add at least one test case that no only proves that the functionality works, but as importantly will be used to regularly check that the functionality is not broken by some future change.

For simple CLI commands, tests are listed directly into [`tests/Makefile`](https://github.com/lz4/lz4/blob/dev/tests/Makefile#L226).

Cyan4973 (changes requested):
add 1 test case in `tests/Makefile`

Cyan4973 (approved):

Cyan4973:
That's perfect, thanks @filipecalasans !

PR #853, opened 2020-04-16
Add WindowsTargetPlatformVersion of 10.0 and lz4 executable to VS2017…

mingwandroid:
… Visual Studio solution

Hi Yann, please give this your consideration. Many thanks!

### DIFF
--- a/visual/VS2017/lz4.sln
+++ b/visual/VS2017/lz4.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.28307.271
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.30011.22
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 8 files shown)


visual/VS2017/lz4.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.28307.271
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.30011.22
```
Cyan4973, line 3: Isn't `Version 16` actually Visual Studio 2019 ?

Cyan4973:
Manually maintaining multiple versions of Visual Studio projects is a real pain for maintenance. Any little change must be reflected across all build scripts.

I would very much prefer a different strategy, where these VS projects are generated from a single source, typically a `cmake` script.
We could even pre-generate them inside the repository, so that users don't have to run `cmake` locally.
But the important part is : this should be automated, with a script. Maintainers should not have to manually tweak these files for each version of Visual Studio.

PR #854, opened 2020-04-21
Add some functionality to contrib cmake_unofficial

hjmallon:
I am updating lz4 version in the hunter package manager, which relies a lot on CMake features and currently has a fork with it's own CMake implementation for lz4. I have instead made the minimal changes on the current contrib CMake support to work with hunter (and anyone else using a CMake package based system).

There are 2 commits, one adds support for CMake packages (https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html)
The other just adds an option to skip installation of binaries

### DIFF
--- a/contrib/cmake_unofficial/CMakeLists.txt
+++ b/contrib/cmake_unofficial/CMakeLists.txt
@@ -170,14 +170,17 @@ endforeach (flag)
 
 if(NOT LZ4_BUNDLED_MODE)
   include(GNUInstallDirs)
+  set(TARGETS_EXPORT_NAME "${PROJECT_NAME}Targets")
 
   install(TARGETS ${LZ4_PROGRAMS_BUILT}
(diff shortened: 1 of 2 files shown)


Cyan4973:
The updated `cmake` script apparently fails the CI `cmake` test : 
```
CMake Error at CMakeLists.txt:182 (install):
  install TARGETS given unknown argument "INCLUDES".
CMake Error at CMakeLists.txt:247 (include):
  include could not find load file:
    CMakePackageConfigHelpers
CMake Error at CMakeLists.txt:248 (write_basic_package_version_file):
  Unknown CMake command "write_basic_package_version_file".
```
Possibilities include : 
- it may require capabilities availably only on versions higher than the one present in CI. Note that `cmake` minimum version is currently `2.8.6`. On first look, the proposed script requires at minimum `v3.0.2`.
- it may require a dependency, which is not satisfied on CI system, and by extension, is not guaranteed to be present on user's systems

Issue #783, opened 2019-08-23
LZ4_decompress_safe_partial is broken in 1.9.2


jfkthame:
I'm seeing what I believe is incorrect behavior from `LZ4_decompress_safe_partial` in the latest release.

Specifically, in our use case we have a buffer of compressed data that may have a few bytes of padding at the end. We know the expected uncompressed size, but only the rounded-up (possibly padded) size of the compressed data. Therefore, we've been using `LZ4_decompress_safe_partial` to decompress, because it would ignore residual input once it has decompressed the expected `targetOutputSize` bytes (unlike `LZ4_decompress_safe` which expects that "compressedSize : is the exact complete size of the compressed block").

With the update from 1.9.1 to 1.9.2, however, the behavior of `LZ4_decompress_safe_partial` has changed, and it now returns an error if there are a *small* number of extra bytes at the end of the source. It works as expected if there's a lot of extra data; in this case, it stops successfully after decompressing the requested `targetOutputSize`. But when there are around 1-7 extra bytes -- exactly the sort of case that's likely to arise if we're working with a padded block -- it fails. I don't think this is expected behavior for this function.

See attached test file, which compares the behavior of `LZ4_decompress_safe` and `LZ4_decompress_safe_partial` when the given length of the source is slightly wrong. As expected, both fail if the source is truncated, and `LZ4_decompress_safe` also fails if the given `compressedSize` is too large. `LZ4_decompress_safe_partial`, however, fails for values of `srcSize` that are just slightly larger than needed (1-7 bytes), and then starts succeeding again for values that are substantially larger:

```
Length of source: 1313
Compressed size: 992
Trying input size 984:
  decompress_safe -> FAILED (-978)
  decompress_safe_partial -> FAILED (-978)
Trying input size 985:
  decompress_safe -> FAILED (-978)
  decompress_safe_partial -> FAILED (-978)
Trying input size 986:
  decompress_safe -> FAILED (-978)
  decompress_safe_partial -> FAILED (-978)
Trying input size 987:
  decompress_safe -> FAILED (-978)
  decompress_safe_partial -> FAILED (-978)
Trying input size 988:
  decompress_safe -> FAILED (-984)
  decompress_safe_partial -> FAILED (-984)
Trying input size 989:
  decompress_safe -> FAILED (-984)
  decompress_safe_partial -> FAILED (-984)
Trying input size 990:
  decompress_safe -> FAILED (-984)
  decompress_safe_partial -> FAILED (-984)
[...]

PR #861, opened 2020-05-08
fix issue #783

BellaXlp:
fix issue #783


BellaXlp:
There's a new pull request somewhere else.

PR #862, opened 2020-05-08
fix issue #783

BellaXlp:
fix issue #783

### DIFF
--- a/tests/decompress-partial.c
+++ b/tests/decompress-partial.c
@@ -31,13 +31,13 @@ int main(void)
   int srcLen = (int)strlen(source);
   char cmpBuffer[BUFFER_SIZE];
   char outBuffer[BUFFER_SIZE];
-  int cmpSize, result;
+  int cmpSize;
   int i;
   
   cmpSize = LZ4_compress_default(source, cmpBuffer, srcLen, BUFFER_SIZE);
 
   for (i = cmpSize; i < cmpSize + 10; ++i) {
-    result = LZ4_decompress_safe_partial(cmpBuffer, outBuffer, i, srcLen, BUFFER_SIZE);
+    int result = LZ4_decompress_safe_partial(cmpBuffer, outBuffer, i, srcLen, BUFFER_SIZE);
     if ((result < 0) || (result != srcLen) || memcmp(source, outBuffer, srcLen)) {
       printf("test decompress-partial error \n");
       return -1;
--- a/tests/decompress-partial.c
+++ b/tests/decompress-partial.c
@@ -28,7 +28,7 @@ const char source[] =
 
 int main(void)
 {
-  size_t srcLen = strlen(source);
+  int srcLen = (int)strlen(source);
   char cmpBuffer[BUFFER_SIZE];
   char outBuffer[BUFFER_SIZE];
   int cmpSize, result;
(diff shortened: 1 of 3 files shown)


Cyan4973:
Thanks for looking into this issue @BellaXlp . 
I really like that you implemented a test case for it.

I will have a deeper look into the issue later this month, and notably increase testing capabilities to catch thus situation in the future.
I'm likely to merge your patch at this point.

BellaXlp:
@Cyan4973 I will add use cases in these 2 days

BellaXlp:
@Cyan4973  OK. I will add use cases in these 2 days.

tests/decompress-partial.c
```diff
@@ -0,0 +1,49 @@
...
+  int srcLen = (int)strlen(source);
+  char cmpBuffer[BUFFER_SIZE];
+  char outBuffer[BUFFER_SIZE];
+  int cmpSize;
+  int i;
+  
+  cmpSize = LZ4_compress_default(source, cmpBuffer, srcLen, BUFFER_SIZE);
+
+  for (i = cmpSize; i < cmpSize + 10; ++i) {
+    int result = LZ4_decompress_safe_partial(cmpBuffer, outBuffer, i, srcLen, BUFFER_SIZE);
```
Cyan4973, line 40: I'm trying to understand this test case, as I'm attempting to strengthen the test suite to generate failure scenarios within the `fuzzer`, in order to catch similar bugs in the future.

My understanding is the following : 
- buffer `source,srcLen` is compressed into `cmpBuffer,cmpSize `
- then the compressed buffer is `cmpBuffer,i` decompressed into `outBuffer,srcLen` with `LZ4_decompress_safe_partial()` using varying `i` input size.

However, according to this [function documentation](https://github.com/lz4/lz4/blob/dev/lib/lz4.h#L220) , `LZ4_decompress_safe_partial()` is still expected to receive as input a _complete compressed block_. What is expected to be "partial" is the output size.

Therefore, it looks to me that this test shouldn't work. As the compressed block is incomplete, the function should detect that it hits the end of the compressed block way too soon, and return an error code.

Yet it seems that, within this PR, the test is included as part of `make test`, and since I don't see a test failure message, I presume the test is passed successfully. That's confusing ...

_edit_ : I don't find the string `test decompress-partial` in any CI log, so it seems the test is just never played.
_edit 2_: Played the test locally, it seems to pass. Investigating ...
_edit 3_: I missed that point : the test checks what happens when provided input size is _too large_ (it's initialized at `cmpSize`, not `0`). Which is the exact scenario described in #783. `LZ4_decompress_safe_partial()` should indeed succeed in this case, at least as long as `targetOutputSize <= srcSize`, which is the tested scenario.

_edit 4_ : excellent, this example made it possible to improve the fuzzer test suite, now generating new scenarios able to trigger this issue. This will make it straightforward to integrate this PR.

_edit 5_ : bad news, while this PR improves the situation, it doesn't fix _all_ problems related to extraneous bytes in the input buffer. There are still some use cases that fail the test.

More investigation needed. I'm likely going to merge this PR into a feature branch, where additional investigation will be done to detect and fix remaining issues.

Cyan4973:
Merging into feature branch `extraInput` for further investigation

tests/decompress-partial.c
BellaXlp: ok

PR #863, opened 2020-05-11
Reducing stack usage in _t_alignment checks

Devernua:
Reduce stack usage during checks by removing unnecessary struct allocations on the stack.

A crash occurred when using a small stack size. This can be avoided by not allocating the struct during checks.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1349,8 +1349,8 @@ LZ4_stream_t* LZ4_createStream(void)
                      while actually aligning LZ4_stream_t on 4 bytes. */
 static size_t LZ4_stream_t_alignment(void)
 {
-    struct { char c; LZ4_stream_t t; } t_a;
-    return sizeof(t_a) - sizeof(t_a.t);
+    typedef struct { char c; LZ4_stream_t t; } t_a;
+    return sizeof(t_a) - sizeof(LZ4_stream_t);
 }
 #endif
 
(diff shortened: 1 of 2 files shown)


Cyan4973:
Interesting,
I would have expected any compiler to understand that the declared variable is never used, 
thus the function merely produces a compile-time constant, simplifying it drastically. 

This seems confirmed when [looking at produced assembly](https://godbolt.org/z/8xz664), even with no optimization (`-O0`) setting.

What kind of compiler, or compiler settings, were you using to observe the stack usage issue ?

Devernua:
@Cyan4973 clang(11.0.3) -O0

Cyan4973:
`clang-11` has not been released yet.
Do you mean Apple's `clang` ? Be aware they have their own numbering scheme. 
They also have their own defaults and carved their own dedicated scope, meaning it could generate something different from "normal" `clang`.

As far as I can tell, I'm unable to reproduce the issue with any version of `clang`, including development versions, using `-O0` optimization setting.

Devernua:
Yes, clang version on macOS. 
The problem occurs not only on clang, https://godbolt.org/z/szT96x, just need add array

Cyan4973:
Thanks for the reproduction case.
It's indeed much clearer with a large array.

Anyway, your proposed fix looks totally harmless, and doesn't require any dce cleverness from the compiler, so it's pretty good.

Cyan4973 (approved):

PR #879, opened 2020-06-17
Fix issue 878 - Bitwise & instead of logical && in LZ4HC_FindLongerMatch

YuryChaikou:


Cyan4973:
As already explained in the related issue, this was not a mistake

YuryChaikou:
Sacrificing readability for dubious optimization is a mistake itself.

PR #885, opened 2020-07-21
meson: remove build requirement for distutils

neheb:
Tested on a default installation of Debian 10.

Same as https://github.com/facebook/zstd/pull/2197

Signed-off-by: Rosen Penev <rosenp@gmail.com>


bimbashrestha (approved):
This is the same patch we accepted in zstd (https://github.com/facebook/zstd/pull/2197) and there haven't been problems there so this seems good to go to me.

Cyan4973:
Thanks @neheb  and @bimbashrestha  !

PR #890, opened 2020-08-03
Call LZ4_memcpy() instead of memcpy()

terrelln:
`LZ4_memcpy()` uses `__builtin_memcpy()` to ensure that clang/gcc
can inline the `memcpy()` calls in freestanding mode.

This is necessary for decompressing the Linux Kernel with LZ4.
Without an analogous patch decompression ran at 77 MB/s, and with
the patch it ran at 884 MB/s.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -277,6 +277,21 @@ typedef enum {
 /*-************************************
 *  Reading and writing into memory
 **************************************/
+
+/**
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Panchajanya1999:
To all the guys picking this, y'all getting battery life hits coz of more CPU cycles??

PR #891, opened 2020-08-05
Add support for shared library build on MSVC (dll build)

emmenlau:
This PR adds the possibility to build lz4 as a shared library also on MSVC. This requires an update to the xxhash library in order to have the definition `XXH_EXPORT` available.


Cyan4973:
Updating `xxhash` should probably be left to a different PR.
It introduces new issues, which block the main objective of this PR (`cmake`).

emmenlau:
> Updating `xxhash` should probably be left to a different PR.
> It introduces new issues, which block the main objective of this PR (`cmake`).

Conceptually I agree. But the current `xxhash` does not feature the `dllexport`. I think it may be moderately easy to add, but I would not feel confident doing this myself.

Therefore the best I could do is to suggest updating `xxhash` _before_ considering this PR, or somebody more knowledgable than me looking into adding `dllexport` to the current `xxhash`.

By the way, it seems newer `xxhash` has quite a number of code improvements, so maybe it would be worth a look anyways?

Cyan4973:
ping @emmenlau ,
please consider removing xxhash update from this PR

emmenlau:
> ping @emmenlau ,
> please consider removing xxhash update from this PR

I can do that, but then the PR will no longer build successfully due to missing dllexport in xxhash. Can you clarify if this is what you want?

Cyan4973:
Sorry, I did not realize that `__declspec(dllexport)` was so important to the formula.
I'm not too familiar with Visual library compilation, so I may miss something.

I would expect, from the name of it, that `__declspec(dllexport)` means something along these lines : 
"`XXH_*` symbols are publicly accessible (invocable) by users of the dynamic library".

Is it a correct interpretation ?
If yes, why do you believe it's necessary ?

My own expectation is that `XXH_*` symbols should remain strictly internal to the library, hence not be accessible nor visible by user processes of the dynamic library.

emmenlau:
> I would expect, from the name of it, that `__declspec(dllexport)` means something along these lines :
> "`XXH_*` symbols are publicly accessible (invocable) by users of the dynamic library".
> 
> Is it a correct interpretation ?
> If yes, why do you believe it's necessary ?

Yes, this is perfectly correct.

> My own expectation is that `XXH_*` symbols should remain strictly internal to the library, hence not be accessible nor visible by user processes of the dynamic library.

I am under the same impression. But during creation of the shared library build, the tool `lz4.exe` was complaining about one or two missing symbols from `XXH_*`. I don't remember the details from the top of my head, but updating `xxhash` and setting the option for `dllexport` fixed the issue for me. If you want, I can try to reproduce the issue?

Cyan4973:
> during creation of the shared library build, the tool lz4.exe was complaining about one or two missing symbols from XXH_*.

I believe these are 2 different topics.

The shared library has nothing to do with `lz4.exe` CLI. 
It shall compile irrespective of any issue encountered or not during `lz4.exe` compilation.

Now, mentioning that 
> `lz4.exe` was complaining about one or two missing symbols from `XXH_*`

this is an issue centered entirely on `lz4.exe`.

I presume an implied context is that `lz4.exe` was created by a recipe linking to the previously mentioned dynamic library.
If that's the case, then it's the recipe of `lz4.exe` which is incorrect. It should not rely on the shared lz4 library to provide `XXH_*` symbols, which are out of its scope (note: the CLI probably uses some `XXH_*` symbols in the benchmark module typically).

Building `lz4.exe` entirely from source, or by linking CLI code to the shared lz4 library, are 2 distinct builds.
It should not be assumed that they are trivially interchangeable. Recipe must be adjusted.

emmenlau:
> > during creation of the shared library build, the tool lz4.exe was complaining about one or two missing symbols from XXH_*.
> 
> I believe these are 2 different topics.
> 
> The shared library has nothing to do with `lz4.exe` CLI.
> It shall compile irrespective of any issue encountered or not during `lz4.exe` compilation.

Yes, this makes perfect sense. I have indeed mixed the topics. During testing of the build of the shared library, I used the executables as a validation for undefined symbols. This may not be the best possible validation. But I was assuming that it may be a good start, i.e. if all tools and tests can be compiled and used against the shared library, then at least a number of relevant symbols must be correctly exported.

> > `lz4.exe` was complaining about one or two missing symbols from `XXH_*`
> 
> this is an issue centered entirely on `lz4.exe`.
> 
> I presume an implied context is that `lz4.exe` was created by a recipe linking to the previously mentioned dynamic library.

Yes, correct.

> If that's the case, then it's the recipe of `lz4.exe` which is incorrect. It should not rely on the shared lz4 library to provide `XXH_*` symbols, which are out of its scope (note: the CLI probably uses some `XXH_*` symbols in the benchmark module typically).

Ok, I did not know that, but it makes perfect sense.

> Building `lz4.exe` entirely from source, or by linking CLI code to the shared lz4 library, are 2 distinct builds.
> It should not be assumed that they are trivially interchangeable. Recipe must be adjusted.

I've just checked which symbols was missing. Maybe I was wrong about `lz4.exe`? Here is the actual error message from the linker:
```
bench.c.obj : error LNK2019: unresolved external symbol XXH64 referenced in function BMK_benchMem
```

Cyan4973:
> I've just checked which symbols was missing.

Yes, that's confirmation that the CLI requires `XXH64()` symbol for the benefit of its benchmark unit.

In order to get access to these symbols, the recipe building the `lz4.exe` variant linked to the shared library must also add `xxhash.c` to the list of locally compiled unit (hence statically linked). This should fix the reported link error.

Cyan4973:
For information,
I added a target [`lz4-wlib`](https://github.com/lz4/lz4/blob/dev/programs/Makefile#L93) in `Makefile` able to build a version of `lz4` CLI linked to the dynamic library.

`XXH64()` is not the main issue, it can be easily solved.
However, the CLI also requires access to several unstable symbols, which can be discovered by using `LZ4_STATIC_LINKING_ONLY` during compilation, but are not exposed by the dynamic library by default. Therefore, it requires usage of special build macros during library creation time to expose them, creating additional risks of issues at runtime link stage.

In short, beyond very specific setups where the version of the dynamic library can be guaranteed at runtime, it's really not recommended.

Cyan4973:
For information, we are heading towards a new release in the near future,
so if you believe any part of this PR is worth merging,
I recommend simplifying its content into small (yet wholly contained) unitary elements,
that can be more easily reviewed and merged.

Cyan4973:
Is it still an active PR ?

emmenlau:
I'm sorry for not getting back to this... I lost a bit track of what I need to do next? Can you say again what you think should be done (best would be bulletpoints) :-) Thanks!

Cyan4973:
Well, I may have misread the PR.

I initially thought that this PR was providing a way to create `liblz4` as a shared library on MSVC via `cmake`.

> This PR adds the possibility to build lz4 as a shared library also on MSVC

But reading it again, since the message mentions `lz4` rather than `liblz4`,
it seems that what is actually meant is that 
this PR tries to build the `lz4` CLI itself, by linking it against an existing `liblz4` shared library.

In which case, this goal is currently not easily achievable,
due to usage of `LZ4_STATIC_LINKING_ONLY` declarations, 
that require to lock the library version (hence _static_ linking).
This limitation is independent from MSVC.

emmenlau:
Dear @Cyan4973 I'm sorry if this has caused a misunderstanding. Most of the confusion is probably coming from the fact that I'm not very involved in the build process of this repository. I'll try to clarify: my goal is only to build the lz4 library shared, not anything else. I assume this is `liblz4`.

What I did in the process was to link the `lz4` executable against this shared library. My reason to do so was for testing purposes. The idea was that a working shared library build could be well tested against the executable. But I'm not at all keen on this aspect, and very much value your input here.

So, to get back to this: I would very much like to help in setting up a shared library build for `liblz4`. But I lost a bit track about a number of questions:
 - which symbols should or should not be exported from the shared library?
 - how should the shared library build deal with the executable, if the `lz4` executable can _not_ link against the shared library. Should the executable just be disabled, or should both the shared and static library be built?

If you could help me get this back on track it would be great to have the shared library build ready and merged soon.

Cyan4973:
The `lz4` executable must be linked against the _static_ library, so the static library must necessarily be built too.

Since your goal is only to deal with the _shared_ library, you can probably leave that question aside. Building the `lz4` executable is likely already being taken care of by the current `cmake` script.

That being said, the part where I'm lost is that [the shared library seems also taken care of by the `cmake` script](https://github.com/lz4/lz4/blob/dev/build/cmake/CMakeLists.txt#L70).
I have no experience of using `cmake` to generate `MSVC` builds, but I would assume that the current `cmake` recipe already takes care of building the shared library for all targets, thus including `MSVC` ?

emmenlau:
> That being said, the part where I'm lost is that [the shared library seems also taken care of by the `cmake` script](https://github.com/lz4/lz4/blob/dev/build/cmake/CMakeLists.txt#L70).
> I have no experience of using `cmake` to generate `MSVC` builds, but I would assume that the current `cmake` recipe already takes care of building the shared library for all targets, thus including `MSVC` ?

The cmake script allows to _enable_ a shared library build on MSVC. But the build fails. I am under the impression that it can not work in the current form, and I assume it was never tested, at least not with cmake. This was the very reason for starting this PR in the first place.

Cyan4973:
OK !
So I guess that's the value of the proposed work : ensure the `cmake` script _actually_ builds successfully a shared library on MSVC. Limit the PR to this single objective.

If you also want to add a test to ensure it does work as intended,
consider `examples/`, as most mini-projects there are meant to link against the shared `liblz4` library.

Issue #822, opened 2019-12-19
LZ4_resetStream_fast in LZ4_attach_dictionary doesn't make sense any more


buweilv:
function "LZ4_resetStream_fast" is added in LZ4_attach_dictionary at commit "f2ae385c2f" in order to make sure that changes will not be erased by subsequent calls to LZ4_resetStream_fast() in case stream was marked as having dirty context, e.g. requiring full reset.
Correspondingly， when LZ4_compress_generic goto failure cctx->dirtyContext is set to 1. However, now,  LZ4_compress_generic goto failure , no cctx->dirty would be set. So  `if (cctx->dirty)` in function LZ4_resetStream_fast makes no sense. No body will set this variable to 1.

felixhandte:
Hi @buweilv,

Thanks for bringing this to our attention! Sorry it took so long to get to this. I believe you're correct. `dirty` is never set by our codebase. It remains only to detect uninitialized memory passed in by users, for which it isn't really sufficient. So I've put up #892 to remove this.

PR #892, opened 2020-08-05
Remove dirty Flag in LZ4_stream_t

felixhandte:
Nothing internally sets dirty anymore. The only way to get that is if you use an uninitialized context, in which case your warranty is void anyways.

Addresses #822.


Cyan4973 (approved):

Cyan4973:
I guess if there is no more usage of `cctx->dirty`, it might be time to remove it.

Cyan4973:
It seems there is a (minor) cast warning left on Visual,
possibly related to the change of type for member `tableType` ?

Cyan4973 (approved):

PR #896, opened 2020-08-10
fix #832

### LINKED ISSUE #832: lz4.h is trying to use undefined __GNUC__
Super low priority issue.

When compiling `lz4` as part of another library with clang in Visual Studio, I'm getting a warning about undefined `__GNUC__`:
```
lz4/lz4.h(673,10): warning : '__GNUC__' is not defined, evaluates to 0
```
which comes from admittedly "obsolete" corner of the code and which probably ends up in correct branch anyway:
```c
#ifdef LZ4_DISABLE_DEPRECATE_WARNINGS
#  define LZ4_DEPRECATED(message)   /* disable deprecation warnings */
#else
#  define LZ4_GCC_VERSION (__GNUC__ * 100 + __GNUC_MINOR__)
#  if defined (__cplusplus) && (__cplusplus >= 201402) /* C++14 or greater */
#    define LZ4_DEPRECATED(message) [[deprecated(message)]]
#  elif (LZ4_GCC_VERSION >= 405) || defined(__clang__)
#    define LZ4_DEPRECATED(message) __attribute__((deprecated(message)))
#  elif (LZ4_GCC_VERSION >= 301)
#    define LZ4_DEPRECATED(message) __attribute__((deprecated))
#  elif defined(_MSC_VER)
#    define LZ4_DEPRECATED(message) __declspec(deprecated(message))
#  else
#    pragma message("WARNING: You need to implement LZ4_DEPRECATED for this compiler")
#    define LZ4_DEPRECATED(message)
#  endif
#endif /* LZ4_DISABLE_DEPRECATE_WARNINGS */
```

Cyan4973:
does no longer rely on default 0-interpretation when `__GNUC__` is not defined

### DIFF
--- a/lib/lz4.h
+++ b/lib/lz4.h
@@ -665,18 +665,17 @@ union LZ4_streamDecode_u {
 #ifdef LZ4_DISABLE_DEPRECATE_WARNINGS
 #  define LZ4_DEPRECATED(message)   /* disable deprecation warnings */
 #else
-#  define LZ4_GCC_VERSION (__GNUC__ * 100 + __GNUC_MINOR__)
 #  if defined (__cplusplus) && (__cplusplus >= 201402) /* C++14 or greater */
 #    define LZ4_DEPRECATED(message) [[deprecated(message)]]
-#  elif (LZ4_GCC_VERSION >= 405) || defined(__clang__)
-#    define LZ4_DEPRECATED(message) __attribute__((deprecated(message)))
-#  elif (LZ4_GCC_VERSION >= 301)
-#    define LZ4_DEPRECATED(message) __attribute__((deprecated))
 #  elif defined(_MSC_VER)
 #    define LZ4_DEPRECATED(message) __declspec(deprecated(message))
+#  elif defined(__clang__) || (defined(__GNUC__) && (__GNUC__ >= 4) && (__GNUC_MINOR__ >= 5))
+#    define LZ4_DEPRECATED(message) __attribute__((deprecated(message)))
+#  elif defined(__GNUC__) && (__GNUC__ >= 3) && (__GNUC_MINOR__ >= 1)
+#    define LZ4_DEPRECATED(message) __attribute__((deprecated))
 #  else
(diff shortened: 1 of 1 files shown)


terrelln (approved):

lib/lz4.h
```diff
@@ -665,18 +665,17 @@ union LZ4_streamDecode_u {
...
-#  define LZ4_GCC_VERSION (__GNUC__ * 100 + __GNUC_MINOR__)
 #  if defined (__cplusplus) && (__cplusplus >= 201402) /* C++14 or greater */
 #    define LZ4_DEPRECATED(message) [[deprecated(message)]]
-#  elif (LZ4_GCC_VERSION >= 405) || defined(__clang__)
-#    define LZ4_DEPRECATED(message) __attribute__((deprecated(message)))
-#  elif (LZ4_GCC_VERSION >= 301)
-#    define LZ4_DEPRECATED(message) __attribute__((deprecated))
 #  elif defined(_MSC_VER)
 #    define LZ4_DEPRECATED(message) __declspec(deprecated(message))
+#  elif defined(__clang__) || (defined(__GNUC__) && (__GNUC__ >= 4) && (__GNUC_MINOR__ >= 5))
```
aqrit, line 672: I think `__GNUC_MINOR__` is being used incorrectly.
Wouldn't gcc 7.1.0 translate to `__GNUC__ == 7` & `__GNUC_MINOR__ == 1`, which would fail this test?

I believe it should read:
`#elif defined(__clang__) || (defined(__GNUC__) && ((__GNUC__ > 4) || ((__GNUC__ == 4) && (__GNUC_MINOR__ >= 5))))`

ojeda: Indeed. An easier way to write these tests that we use in the kernel is to write something like:

    #define GCC_VERSION (__GNUC__ * 10000		\
                         + __GNUC_MINOR__ * 100	\
                         + __GNUC_PATCHLEVEL__)

which can then be compared as an integer.

Cyan4973: Fixed in https://github.com/lz4/lz4/commit/28e5a4e0542820e81f8372c1a405310e9757cd7e

PR #899, opened 2020-08-13
Clarifies and fix EndMark

Cyan4973:
EndMark, the 4-bytes value indicating the end of frame,
must be `0x00000000`.

Previously, it was just specified as a `0-size` block.
But such definition could be understood as encompassing uncompressed blocks of size 0,
with a header value of `0x80000000`.

However, the intention always was to support uncompressed empty blocks.
Note that compressed empty blocks are already supported,
it's just that they have a size of 1 instead of 0 (for the `0` token).
Such block could be used as a kind of keep-alive signal for example.

Unfortunately, the decoder implementation was also wrong,
and would interpret a `0x80000000` block header as an endMark.

This issue evaded detection so far simply because
this situation never happens, as LZ4Frame always issues
a clean `0x00000000` value as a endMark.
It also does not flush and therefore does not produce empty blocks.

This is fixed in this PR.
The decoder can now deal with empty uncompressed blocks,
and do not confuse them with EndMark.
The specification is also clarified.
Finally, FrameTest is updated to randomly insert empty blocks during fuzzing.

### DIFF
--- a/doc/lz4_Frame_format.md
+++ b/doc/lz4_Frame_format.md
@@ -85,8 +85,8 @@ That’s where compressed data is stored.
 
 __EndMark__
 
-The flow of blocks ends when the last data block is announced with
-an invalid size of “0”, expressed as a 32-bits value (exactly `0x00000000`).
+The flow of blocks ends when the last data block is followed by
+the 32-bit value `0x00000000`.
 
 __Content Checksum__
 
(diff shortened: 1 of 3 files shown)


doc/lz4_Frame_format.md
```diff
@@ -85,14 +85,13 @@ That’s where compressed data is stored.
 
 __EndMark__
 
-The flow of blocks ends when the last data block has a size of “0”.
-The size is expressed as a 32-bits value.
+The flow of blocks ends when the last data block is announced with
+an invalid size of “0”, expressed as a 32-bits value (exactly `0x00000000`).
```
terrelln: Maybe reference the `__Block Size__` section, because I don't know what an "invalid size of 0" means until I read that section.

Cyan4973: Wording has been simplified, to only mention a "value of `0x00000000`",
thus making the concept of "size" irrelevant in the context of `endMark`.

terrelln (approved):

Issue #904, opened 2020-08-24
cli: Handling of "stdinmark" can lead to surprising results


nh2:
Wow, I got amazing compression! Compressed a 2 GB file into 2 Bytes!

```
$ lz4 --list /tmp/test.lz4
    Frames           Type Block  Compressed  Uncompressed     Ratio   Filename
         1       LZ4Frame   B7I       1.96G          2.00 105394020300.00%  test.lz4 
```

How did I do it? I just ran `lz4 --content-size < /tmp/test > /tmp/test.lz4`. But I accidentally also had a file called `stdin` in my directory that contained a single letter plus newline:

```
$ cat stdin
a
```

The reason:

https://github.com/lz4/lz4/blob/34fe7c9d7f4fbeeb42c22f4ad1a01c1771d14268/programs/lz4io.c#L682-L684

`UTIL_getFileSize()` does:

https://github.com/lz4/lz4/blob/34fe7c9d7f4fbeeb42c22f4ad1a01c1771d14268/programs/util.h#L373-L376

But `lz4` also supports passing the magic file name `stdin` instead of `-` for reading from stdin. For example, you can do `echo foo | lz4 stdin`.

https://github.com/lz4/lz4/blob/34fe7c9d7f4fbeeb42c22f4ad1a01c1771d14268/programs/lz4io.h#L43

That may be fine conceptually, but if that is done, lz4 should not try to afterwards `stat()` that file by name.

In general, using a single variable (`srcFileName`) for two different domains makes these kind of issues easy to occur. In programming languaes that support sum types, this would be an `Option` or similar. Perhaps lz4 should use `srcFileName = NULL` to make clear that stdin is used, or use a bool for it, or alternatively check whether it's `stdinmark` before calls to any function that operates on actual files (like `UTIL_getFileSize()`).

nh2:
PR #905 fixes with this specific case I encountered but there may be other similar occurences of the same problem in the code.

PR #905, opened 2020-08-24
Use fstat() to determine file size

nh2:
This allows us to get the file size even when the input file is passed
via stdin. This fixes `--content-size` not working in situations like

    $ lz4 -v --content-size < /tmp/test > /tmp/test.lz4
    Warning : cannot determine input content size

With this change, it works.

Also helps with #904.

---

Note I have only tested compilation and run on Linux, I modified the Windows `#if defined`s blindly.

### DIFF
--- a/programs/lz4io.c
+++ b/programs/lz4io.c
@@ -680,7 +680,7 @@ LZ4IO_compressFilename_extRess(LZ4IO_prefs_t* const io_prefs, cRess_t ress,
     prefs.frameInfo.contentChecksumFlag = (LZ4F_contentChecksum_t)io_prefs->streamChecksum;
     prefs.favorDecSpeed = io_prefs->favorDecSpeed;
     if (io_prefs->contentSizeFlag) {
-      U64 const fileSize = UTIL_getFileSize(srcFileName);
+      U64 const fileSize = UTIL_getOpenFileSize(srcFile);
       prefs.frameInfo.contentSize = fileSize;   /* == 0 if input == stdin */
       if (fileSize==0)
           DISPLAYLEVEL(3, "Warning : cannot determine input content size \n");
@@ -1472,7 +1472,7 @@ LZ4IO_getCompressedFileInfo(LZ4IO_cFileInfo_t* cfinfo, const char* input_filenam
     LZ4IO_infoResult result = LZ4IO_format_not_known;  /* default result (error) */
     unsigned char buffer[LZ4F_HEADER_SIZE_MAX];
     FILE* const finput = LZ4IO_openSrcFile(input_filename);
-    cfinfo->fileSize = UTIL_getFileSize(input_filename);
+    cfinfo->fileSize = (finput == NULL) ? 0 : UTIL_getOpenFileSize(finput);
 
     while (!feof(finput)) {
         LZ4IO_frameInfo_t frameInfo = LZ4IO_INIT_FRAMEINFO;
--- a/programs/util.h
+++ b/programs/util.h
@@ -143,6 +143,15 @@ extern "C" {
 #endif
 
 
+/*-****************************************
+*  fileno() function
+******************************************/
+#if defined(_MSC_VER)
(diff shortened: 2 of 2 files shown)


nh2:
I noticed that `make` does not notice when `util.h` gets changed, so nothing gets recompiled.

Could that be fixed independently? Thanks!

Cyan4973:
There are a few `C90`-compliance issues remaining, especially around declaration / statement order.

Cyan4973:
> get the file size even when the input file is passed via `stdin`

This is a new capability, that wasn't expected to be part of `stdin` interface.

Obviously, this can still be a nice addition, though it also means increased attention to potential limits,
be it from specific systems, or from specific use cases.

For portability, it seems you are following the recipe of `UTIL_getFileSize()`, so it may be equivalent. 
Although this additional `fileno()` invocation, just by itself, can end up being a portability concern.

Also, what happens for example when `stdin` comes from program generation, as opposed to file system ?

I guess a few more tests should be added to illustrate and control wanted behavior.

nh2:
> There are a few `C90`-compliance issues remaining, especially around declaration / statement order.

I've pushed a new commit that fixes it and also collects the `#ifdef`s into one place similar how it's done for other functions' portability. If you find that better / more readable, then I will squash it in.

> This is a new capability, that wasn't expected to be part of `stdin` interface.

Yes.

I think it's a good idea though, because many other unix programs also handle this, e.g. `pv < bigfile | gzip > /dev/null` lets `pv` show you a progress bar because it knows the absolute file size ([code here](https://github.com/icetee/pv/blob/8e1592bf5c4df88a1d2b7f6d1045b3a75accce15/src/pv/file.c#L44)).

> Also, what happens for example when `stdin` comes from program generation, as opposed to file system ?

Then `S_ISREG(st_mode)` is `false` as before, and `0` is returned as before.

If it comes from a file (if you do `lz4 < myfile`), then `S_ISREG(st_mode)` is `true`.

> For portability, it seems you are following the recipe of `UTIL_getFileSize()`, so it may be equivalent.
> Although this additional `fileno()` invocation, just by itself, can end up being a portability concern.

[`platform.h` uses `_fileno()`](https://github.com/lz4/lz4/blob/d64962a44bde352e24bf2e7553dc68c13568821e/programs/platform.h#L111) in some Windows ifdefs, probably because https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/posix-fileno?view=vs-2019

> The Microsoft-implemented POSIX function name `fileno` is a deprecated alias for the `_fileno` function. By default, it generates [Compiler warning (level 3) C4996](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996?view=vs-2019). The name is deprecated because it doesn't follow the Standard C rules for implementation-specific names. However, the function is still supported.

I see the Appveyor test [pointed this out](https://ci.appveyor.com/project/YannCollet/lz4-1lndh/builds/34831629/job/9j0818hnqb12vwpy); I'll try to fix it.

> I guess a few more tests should be added to illustrate and control wanted behavior.

I'm not really familiar with how the test suite works, and have rather limited time, so I'd like to pass that on to somebody else, but I imagine a good test would be to check that `lz4 < myfile` works exactly as `lz4 myfile`.

nh2:
> I see the Appveyor test [pointed this out](https://ci.appveyor.com/project/YannCollet/lz4-1lndh/builds/34831629/job/9j0818hnqb12vwpy); I'll try to fix it.

Looks like that is [green now](https://ci.appveyor.com/project/YannCollet/lz4-1lndh/builds/34831945) so the `fileno` fix should be working.

Please take another look!

Cyan4973:
PR itself looks good to me, but requires more testing.
Redirecting to `stdin_size` feature branch, in order to add relevant tests before merging into `dev`.

nh2:
Linking related PR #908

PR #919, opened 2020-09-23
Add CMake option LZ4_BUILD_CLI

attilaolah:
Defaults to `ON` so nothing changes, but it can be set to `OFF` so that the CLI is not built at all, only the library, i.e. like so:

```sh
$ cmake -DLZ4_BUILD_CLI:BOOL=OFF
```

This is useful in combination with `-DLZ4_BUILD_LEGACY_LZ4C:BOOL=OFF` for builds where only the library is needed.


Cyan4973 (approved):

Cyan4973:
Looks good !

Issue #784, opened 2019-08-25
Cannot compress long zeroed buffer properly with LZ4_compress_HC_destSize()


hsiangkao:
tested with the latest lz4 upstream library

tst.c
#include <lz4hc.h>
#include <string.h>
#include <stdio.h>

int main(void)
{

	int SrcSize = 1642496;
	char buf[1642496];
	char dst[4096];
	int compressed;

	void *ctx = LZ4_createStreamHC();

	memset(buf, 0, sizeof(buf));
	compressed = LZ4_compress_HC_destSize(ctx, buf, dst, &SrcSize, 4096, LZ4HC_CLEVEL_DEFAULT);

	LZ4_freeStreamHC(ctx);
	printf("compressed: %d SrcSize: %d", compressed, SrcSize);
	return 0;
}
gcc -I lz4-dev/lib -L lz4-dev/lib -o test tst.c lz4-dev/lib/liblz4.a

result:
compressed: 4095 SrcSize: 4078

and it seems goto _dest_overflow but I didn't dig into that. LZ4_compress_destSize seems fine.

hsiangkao:
and LZ4HC_CLEVEL_MAX (LZ4HC_compress_optimal) has the same issue.

hsiangkao:
it seems that LZ4HC_InsertAndFindBestMatch didn't consider the outputbufferlimit, and generate a much longer match with the matchlength is larger than the buffer size...

pos:      0 -- literals:  1, match:1642490, offset:    1, cost:6446 + 0
                                                                                                                ^ > 4096
hmmm...

remittor:
It seems to me that it's my fault: https://github.com/lz4/lz4/commit/f007153e3f0c401d51d6ba5f6ecf27b223591ef2
I'll try to fix it ...

remittor:
@hsiangkao, @Cyan4973 
> and it seems goto _dest_overflow but I didn't dig into that. LZ4_compress_destSize seems fine.

Seems to have fixed: https://github.com/remittor-pr/lz4/commits/hc_dstsize_3
Fuzzing test completed successfully!

Now should add new tests to the `fazzer`.

esxgx:
Hi remittor,
I have not enough time to do the full regression test since I'm mainly focusing on EROFS development now.
But I'm very happy it could be fixed in LZ4 upstream :)

Thanks,
Gao Xiang

Cyan4973:
Going back to this topic,
trying to get it fixed in time for next release,

[...]

PR #923, opened 2020-09-28
fix efficiency of LZ4_compress_HC_destSize()

Cyan4973:
LZ4_compress_HC_destSize() had a tendency
to discard its last match when this match overflowed specified dstBuffer limit.
The impact is generally moderate,
but occasionally huge,
typically when this last match is very large
(such as compressing a bunch of zeroes).

Issue #784 fixed for both Chain and Opt implementations.

Added a unit test suggested by @remittor checking this topic.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1185,8 +1185,8 @@ _last_literals:
             if (outputDirective == fillOutput) {
                 /* adapt lastRun to fill 'dst' */
                 assert(olimit >= op);
-                lastRun  = (size_t)(olimit-op) - 1;
-                lastRun -= (lastRun+240)/255;
+                lastRun  = (size_t)(olimit-op) - 1/*token*/;
+                lastRun -= (lastRun + 256 - RUN_MASK) / 256;  /*additional length tokens*/
             } else {
                 assert(outputDirective == limitedOutput);
                 return 0;   /* cannot compress within `dst` budget. Stored indexes in hash table are nonetheless fine */
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -247,6 +247,7 @@ static const int LZ4_minLength = (MFLIMIT+1);
 /*-************************************
 *  Types
 **************************************/
+#include <limits.h>
 #if defined(__cplusplus) || (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) /* C99 */)
 # include <stdint.h>
   typedef  uint8_t BYTE;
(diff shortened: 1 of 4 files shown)


remittor:
Compare code:
https://github.com/lz4/lz4/blob/7a75b045bde89e77e3573f91821d0314251aac1d/lib/lz4.c#L1154-L1155
https://github.com/lz4/lz4/blob/89736e4e27358ed463c2482553cfae1b1176c797/lib/lz4hc.c#L731-L732

Src commit: https://github.com/lz4/lz4/pull/923/commits/e7fe105ac6ed02019d34731d2ba3aceb11b51bb1#diff-bcf1f5402173f298047c4fa808056f3bL719-R726

`RUN_MASK = 15`  => ` llAdd = (lastRunSize + 241) / 256;`

Why?

Cyan4973:
Both formulaes are extremely close to each other,
and both are "correct" with regards to `lz4` guarantees,
aka they won't overwrite the destination buffer, 
and both try to pack as much as possible data into this space.

That being said, the new formula is _slightly_ more accurate, 
hence pack _a bit_ more data into remaining space,
though this is only visible on large or specific lengths.

Examples : 

| Space left after token | Old formula : Nb of Last Literals / Effective size | New formula : Nb of Last Literals / Effective size |
| --- | --- | --- |
| 5 | 5 / 5 | 5 / 5 |
| 14 | 14 / 14 | 14 / 14 |
| 15 | 14 / 14 | 14 / 14 |
| 16 | 15 / 16 | 15 / 16 |
| 269 | 268 / 269 | 268 / 269 |
| 270 | 268 / 269 | __269__  / 270 |
| 271 | 269 / 270 | 269 / 270 |
| 272 | 270 / 272 | 270 / 272 |
| 4094 | 4078 / 4094 | 4078 / 4094 |
| 4096 | 4079 / 4095 | __4080__  / 4096 |
| 4112 | 4095 / 4112 | 4095 / 4112 |
| 65536 | 65279 / 65535 | __65280__ / 65536 |
| 100000 | 99607 / 99998 | __99609__ / 100000 |

The old formula was a bit over-cautious.
It was calculating the amount of additional length bytes required to represent the last segment length,
and then subtracting that amount to produce the literal length. Thus, the literal length was actually a bit shorter,
and maybe could be represented using _less_ additional bytes.
Hence, it was a bit of an over-estimation, but felt "good enough" at the time.

The new formula is implemented into `lz4hc` because it was the center of this investigation. That being said, it can be extended to `lz4` as well.

remittor:
> That being said, the new formula is _slightly_ more accurate,
> hence pack _a bit_ more data into remaining space,

The new formula is not only more accurate, but also faster than the old one, since division by 256 is elementary.
I like the new version!

PR #930, opened 2020-10-06
Fix: The "inline" specifier do not use for LZ4_wildCopy8 and LZ4_wildCopy32

remittor:
This problem was reproduced on MSVC 2015 (32-bit, `-O2`). Both functions were called using the operator "call".

Src: https://github.com/lz4/lz4/commit/73bcf90e51b9837454b92265d30d4a214be38b02

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -162,7 +162,7 @@
 #  define LZ4_FORCE_O2_INLINE_GCC_PPC64LE __attribute__((optimize("O2"))) LZ4_FORCE_INLINE
 #else
 #  define LZ4_FORCE_O2_GCC_PPC64LE
-#  define LZ4_FORCE_O2_INLINE_GCC_PPC64LE static
+#  define LZ4_FORCE_O2_INLINE_GCC_PPC64LE  LZ4_FORCE_INLINE
 #endif
 
 #if (defined(__GNUC__) && (__GNUC__ >= 3)) || (defined(__INTEL_COMPILER) && (__INTEL_COMPILER >= 800)) || defined(__clang__)
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -299,7 +299,7 @@ typedef enum {
 #define LZ4_memcpy(dst, src, size) memcpy(dst, src, size)
 #endif
 
-static unsigned LZ4_isLittleEndian(void)
+LZ4_FORCE_INLINE unsigned LZ4_isLittleEndian(void)
 {
     const union { U32 u; BYTE c[4]; } one = { 1 };   /* don't use static : performance detrimental */
     return one.c[0];
@@ -309,12 +309,12 @@ static unsigned LZ4_isLittleEndian(void)
 #if defined(LZ4_FORCE_MEMORY_ACCESS) && (LZ4_FORCE_MEMORY_ACCESS==2)
 /* lie to the compiler about data alignment; use with caution */
 
-static U16 LZ4_read16(const void* memPtr) { return *(const U16*) memPtr; }
-static U32 LZ4_read32(const void* memPtr) { return *(const U32*) memPtr; }
-static reg_t LZ4_read_ARCH(const void* memPtr) { return *(const reg_t*) memPtr; }
+LZ4_FORCE_INLINE U16 LZ4_read16(const void* memPtr) { return *(const U16*) memPtr; }
+LZ4_FORCE_INLINE U32 LZ4_read32(const void* memPtr) { return *(const U32*) memPtr; }
+LZ4_FORCE_INLINE reg_t LZ4_read_ARCH(const void* memPtr) { return *(const reg_t*) memPtr; }
 
-static void LZ4_write16(void* memPtr, U16 value) { *(U16*)memPtr = value; }
-static void LZ4_write32(void* memPtr, U32 value) { *(U32*)memPtr = value; }
+LZ4_FORCE_INLINE void LZ4_write16(void* memPtr, U16 value) { *(U16*)memPtr = value; }
+LZ4_FORCE_INLINE void LZ4_write32(void* memPtr, U32 value) { *(U32*)memPtr = value; }
 
 #elif defined(LZ4_FORCE_MEMORY_ACCESS) && (LZ4_FORCE_MEMORY_ACCESS==1)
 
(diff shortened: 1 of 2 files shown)


Cyan4973:
Thanks for the PR @remittor ,
is there any measurable performance difference ?

Cyan4973:
I went ahead, and measured performance difference on a Desktop Windows 10 station with Visual 2017 installed.

With this PR, I measure a small loss of performance for `win32` platform, and a small gain of performance for `x64` one.
In both cases, it's small, basically within noise margin (~1 %).

To be fair, considering the importance of function `LZ4_wildCopy*()` to `LZ4` general performance, I was expecting that moving this invocation from a function call to direct inlining would result in a much more perceptible impact. So it's not clear if I'm measuring the same thing (after all, it's MSVC2017 vs MSVC2015).

Note that `x64` performance is more important than `win32` one these days, so even this limited impact qualifies as "positive",
though, to be more complete, one would also have to check performance impact on `clang` and `gcc` as well.

Cyan4973:
okay,
so now this PR enforces a much larger change.

I can understand that it is tempting to simplify the logic with "inlining is always better than a function call",
but in practice, this is not _always true_, for a variety of reasons and side-effects too long to list here.

On the other hand, the one thing that is always true,
is that force inlining always produce a larger (or at best equivalent) binary size than when letting the compiler do this choice automatically.

Because such change may impact performance in many different directions, depending on the compiler used, the set of flags and the platform, it can't be blindly accepted on the ground that one of these configurations experience some benefit. A larger test set must be provided, benchmarked and results listed.

remittor:
> but in practice, this is not _always true_, for a variety of reasons and side-effects too long to list here.

Functions that consist of no more than 4 assembly instructions should always be inlined (`LZ4_read*`, `LZ4_write*`).
The rest of the functions are also pretty short when compared to the functions that were previously marked for inline.
There are huge functions in the `LZ4` code that are marked for inline.
I forced `__forceinline` for those static functions that were automatically inlined by MSVC 2015 compiler.

remittor:
@Cyan4973 
Look at `XXHash`: https://github.com/Cyan4973/xxHash/blob/29356f590f4eb38c671606270abe213a5f317424/xxhash.h#L1129
`XXH_FORCE_INLINE xxh_u32 XXH_readLE32(const void* memPtr)`

Look at `Zstd`: https://github.com/facebook/zstd/blob/f7d49437885fe2ecafeb0d2663a5c891ba57231a/lib/common/bitstream.h#L338
`MEM_STATIC FORCE_INLINE_ATTR size_t BIT_getLowerBits(size_t bitContainer, U32 const nbBits)`

Look agane at `Zstd`: https://github.com/facebook/zstd/blob/f7d49437885fe2ecafeb0d2663a5c891ba57231a/lib/common/mem.h#L228
`#define MEM_STATIC static __inline __attribute__((unused))`
`MEM_STATIC void MEM_write32(void* memPtr, U32 value)`

remittor:
> I went ahead, and measured performance difference on a Desktop Windows 10 station with Visual 2017 installed.
> With this PR, I measure a small loss of performance for `win32` platform, and a small gain of performance for `x64` one.
> In both cases, it's small, basically within noise margin (~1 %).

This problem only occurred when using MSVC 2015 32-bit code generator (with `-O2`)!
MSVC 2008 (32-bit and 64-bit) all small static functions was inlined automatically (without `__forceinline`, with `-O2`).
MSVC 2015 64-bit all small static functions was inlined automatically (without `__forceinline`, with `-O2`).

I made a performance comparison (32-bit code by MSVC 2015). I did not notice any differences.

remittor:
When the compiler uses size optimization `-O1`, small static functions are not inlined into the code!

Here is a list of functions that are not inlined:

**MSVC 2008 (x86 32bit, `-O1 -Oi`)**
```
LZ4_wildCopy8
LZ4_memcpy_using_offset_base
LZ4_wildCopy32
LZ4_memcpy_using_offset
LZ4_NbCommonBytes
LZ4_compressBound           { return LZ4_COMPRESSBOUND(isize); } 
LZ4_getPositionOnHash
```

**Clang 9.0.8 (ARM64, `-O1`)**
```
LZ4_read16           (1 instruction)
LZ4_readLE16         (call LZ4_read16)
LZ4_write16          (1 instruction)
LZ4_read32           (1 instruction)
LZ4_write32          (1 instruction)
LZ4_wildCopy8        (4 instructions)
LZ4_writeLE16        (call LZ4_write16)
LZ4_read_ARCH        (1 instruction)
LZ4_NbCommonBytes    (3 instructions: RBIT, CLZ, UBFX)
```

It makes no sense to check the performance, because the result is known in advance.

Cyan4973:
> When the compiler uses size optimization -O1, small static functions are not inlined into the code!

It's funny because invoking small 1-instruction functions instead of inlining them
is more expensive size-wise (not just performance wise).
Well, I guess these compiler's heuristics are not that good ...

> When the compiler uses size optimization -O1

`-O1` rather means "cheap optimizations".
Size optimization is rather `-Os` for `gcc` and `clang`.
or even `-Oz` for more recent `clang` versions.

Cyan4973 (approved):

Cyan4973:
So apparently this PR breaks something.
It's difficult to understand what and how, but the fuzzer found an issue : 
`decompress_fuzzer: lz4.c:1777: int LZ4_decompress_generic(const char *const, char *const, int, int, endCondition_directive, earlyEnd_directive, dict_directive, const BYTE *const, const BYTE *const, const size_t): Assertion 'oend - op >= FASTLOOP_SAFE_DISTANCE' failed.`
I don't see an obvious link between this issue and the content of this PR, but bisecting clearly points at this PR.
Maybe something related to strict aliasing ?
More investigation, to be checked.

Cyan4973:
At this stage, I'm unconvinced that the issue found is related to the PR.
It's strange that the fuzzer points at it,
but it could be some kind of secondary side effect that guide the fuzzer into some chaotic yet reproducible behavior.

remittor:
> So apparently this PR breaks something.
> It's difficult to understand what and how, but the fuzzer found an issue :

I have checked all possible options. The bug could not be repeated.

I previously ran into bugs in `gcc 4.8` that occurred when using inlines (the stack broke).

> More investigation, to be checked.

There is not enough input data. We need to know the compiler, its version and compilation options.

Cyan4973:
The fuzzer issue is not solved yet, but as mentioned, I'm convinced it's just noise, unrelated to this PR. Investigation ongoing.

However, I also made some speed tests, and noticed a big speed degradation,
a large portion of which can be directly traced to the last commit of this PR.

The PR is fine up to 749bd91, but the last step, which generalizes `FORCE_INLINE` everywhere, goes too far, and performance regresses noticeably.

| file | commit | C.Speed | D.Speed |
| --- | --- | --- | --- |
| silesia.tar | 749bd91 | 532 | 3386 |
| silesia.tar | 0e3933e | 513 | 3263 |
| calgary.tar | 749bd91 | 443 | 3326 |
| calgary.tar | 0e3933e | 426 | 3190 |

It seems it would be best to cherry-pick in order to revert the last commit only
(something I'm not familiar with).

remittor:
> However, I also made some speed tests, and noticed a big speed degradation

Compiler? Version? Options? CPU?

Cyan4973:
It's in the PR

remittor:
> However, I also made some speed tests, and noticed a big speed degradation, ...

It's not the speed drop you've noticed, but the strange behavior of the compiler!
Here are the results of my testing (`fullbench -d1 silesia.tar`):
Compiler + CPU                       | w/o inline  | with inline | %
------------------------------------ | ----------- | ----------- |------
gcc  9.3.0 (mingw64) Xeon X5660@2664 | 1940.6 MB/s | 1863.6 MB/s | -4.0%
gcc 10.2.0 (mingw64) Xeon X5660@2664 | 1941.4 MB/s | 1941.4 MB/s |
gcc  9.3.0 (mingw64) Celeron J1900   | 1152.8 MB/s | 1146.5 MB/s | -0.5%
gcc 10.2.0 (mingw64) Celeron J1900   | 1166.8 MB/s | 1181.2 MB/s | +1.3%
gcc  4.8.4 (ubuntu14) Xeon X5660     | 2264.7 MB/s | 2266.1 MB/s | +0.1%

Therefore, I advise you to update the `gcc` compiler to the latest version.

I decided to compare the ASM code of the `LZ4_decompress_fast` function in `IDA` - both versions are 100% the same (from `gcc 9.3`). Some incredible things ....

Cyan4973:
Indeed, likely some differences in instruction alignments.
Basically a random impact ...

remittor:
> Indeed, likely some differences in instruction alignments.

It's really about the address where the function is located!
Before the function `LZ4_decompress_fast`, I added a `lz4_dummy_1` and the speed increased:
```
int lz4_dummy_1(int k)
{  
    return k * k + 1000;
}

LZ4_FORCE_O2
int LZ4_decompress_fast(const char* source, char* dest, int originalSize)
{
    return LZ4_decompress_generic(source, dest, 0, originalSize,
```

Compiler + CPU                       | w/o inline  | with inline 
------------------------------------ | ----------- | ----------- 
gcc  9.3.0 (mingw64) Xeon X5660@2664 | 1943.7 MB/s | 1944.5 MB/s 

Therefore, you need to do something about speed testing.

Cyan4973:
Trying to control instruction alignment from a C source code
is an entirely different topic,
and I prefer not to be weighted down by it for the purpose of the upcoming release.

This can still be worked upon though, for the benefit of future releases.

PR #932, opened 2020-10-18
Multi thread support for compression

yumeyao:


yumeyao:
using std

Cyan4973:
That's a nice attempt.
I'm certainly opened to such evolution,
though it will have to use `C90` coding standard if it wants to integrate the `lz4` CLI tool
which must remain highly portable.

Alternatively, it could become a separate project,
hosted either into its own repository or within this repository in `contrib/`
where there is no such constraint and C++ code is allowed.

yumeyao:
I created this PR by mistake so I closed it. 

let's use #933 for further disccusion.

PR #936, opened 2020-10-31
More alignment tests

### LINKED ISSUE #827: alignment checks added to lz4hc.c don't work even on linux
This check in lz4hc.c (e.g. in LZ4_initStreamHC())
#ifndef _MSC_VER  /* for some reason, Visual fails the aligment test on 32-bit x86 :
                   * it reports an aligment of 8-bytes,
                   * while actually aligning LZ4_streamHC_t on 4 bytes. */
    assert(((size_t)state & (LZ4_streamHC_t_alignment() - 1)) == 0);  /* check alignment */
#endif

doesn't work even on non-MS Windows 32 bit builds.

The problem is that, while the alignment is checked, no measures are taken to ensure that the buffers passed are aligned. On 64 bit builds it's OK since (apparently) malloc() returns memory alligned on 8 bytes. But on 32 bit it returns memory alligned on 4 bytes and that is not enough for the structure checked here. 
E.g. for wrong allocation look at:
 937 LZ4_streamHC_t* LZ4_createStreamHC(void)
 938 {
 939     LZ4_streamHC_t* const LZ4_streamHCPtr = (LZ4_streamHC_t*)ALLOC(sizeof(LZ4_streamHC_t));
 940     if (LZ4_streamHCPtr==NULL) return NULL;
 941     LZ4_initStreamHC(LZ4_streamHCPtr, sizeof(*LZ4_streamHCPtr));  /* full initialization, malloc'ed buffer      can be full of garbage */
 942     return LZ4_streamHCPtr;
 943 }

Cyan4973:
Re-inforced alignment tests.
Refactor definition of internal states.
No longer depend on `long long` type for alignment, which is apparently problematic in some scenarios.
Rely instead on `void*` alignment, which is more in line with the content of internal states.
fix #827

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -178,18 +178,10 @@
 #define unlikely(expr)   expect((expr) != 0, 0)
 #endif
 
-/* for some reason, Visual Studio can fail the aligment test on 32-bit x86 :
- * it sometimes report an aligment of 8-bytes (at least in some configurations),
- * while only providing a `malloc()` memory area aligned on 4-bytes,
- * which is inconsistent with malloc() contract.
- * The source of the issue is still unclear.
- * Mitigation : made the alignment test optional */
+/* Should the alignment test prove unreliable, for some reason,
+ * it can be disabled by setting LZ4_ALIGN_TEST to 0 */
 #ifndef LZ4_ALIGN_TEST  /* can be externally provided */
-#  if (defined(_MSC_VER) && !defined(_M_X64))
-#    define LZ4_ALIGN_TEST 0  /* disable on win32+visual */
-#  else
-#    define LZ4_ALIGN_TEST 1
-#  endif
+# define LZ4_ALIGN_TEST 1
 #endif
 
 
@@ -476,7 +468,7 @@ LZ4_memcpy_using_offset(BYTE* dstPtr, const BYTE* srcPtr, BYTE* dstEnd, const si
 
     switch(offset) {
     case 1:
-        memset(v, *srcPtr, 8);
+        MEM_INIT(v, *srcPtr, 8);
         break;
     case 2:
         LZ4_memcpy(v, srcPtr, 2);
@@ -1441,7 +1433,7 @@ LZ4_stream_t* LZ4_initStream (void* buffer, size_t size)
     if (buffer == NULL) { return NULL; }
     if (size < sizeof(LZ4_stream_t)) { return NULL; }
     if (!LZ4_isAligned(buffer, LZ4_stream_t_alignment())) return NULL;
-    MEM_INIT(buffer, 0, sizeof(LZ4_stream_t));
+    MEM_INIT(buffer, 0, sizeof(LZ4_stream_t_internal));
     return (LZ4_stream_t*)buffer;
 }
 
@@ -1450,7 +1442,7 @@ LZ4_stream_t* LZ4_initStream (void* buffer, size_t size)
 void LZ4_resetStream (LZ4_stream_t* LZ4_stream)
 {
     DEBUGLOG(5, "LZ4_resetStream (ctx:%p)", LZ4_stream);
-    MEM_INIT(LZ4_stream, 0, sizeof(LZ4_stream_t));
+    MEM_INIT(LZ4_stream, 0, sizeof(LZ4_stream_t_internal));
 }
 
 void LZ4_resetStream_fast(LZ4_stream_t* ctx) {
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -178,6 +178,20 @@
 #define unlikely(expr)   expect((expr) != 0, 0)
 #endif
 
+/* for some reason, Visual Studio can fail the aligment test on 32-bit x86 :
+ * it sometimes report an aligment of 8-bytes (at least in some configurations),
+ * while only providing a `malloc()` memory area aligned on 4-bytes,
+ * which is inconsistent with malloc() contract.
+ * The source of the issue is still unclear.
+ * Mitigation : made the alignment test optional */
+#ifndef LZ4_ALIGN_TEST  /* can be externally provided */
+#  if (defined(_MSC_VER) && !defined(_M_X64))
+#    define LZ4_ALIGN_TEST 0  /* disable on win32+visual */
+#  else
+#    define LZ4_ALIGN_TEST 1
+#  endif
+#endif
+
 
 /*-************************************
 *  Memory routines
@@ -242,6 +256,11 @@ static const int LZ4_minLength = (MFLIMIT+1);
 #else
 #  define DEBUGLOG(l, ...) {}    /* disabled */
 #endif
+
+static int LZ4_isAligned(const void* ptr, size_t alignment)
+{
+    return ((size_t)ptr & (alignment -1)) == 0;
+}
 
 
 /*-************************************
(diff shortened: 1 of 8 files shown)


Cyan4973:
Well, the alignment test checks that the memory area is properly rejected if it's not correctly aligned,
and that's not going to work on Windows 32-bit, since the alignment check is disabled there.

Possible solutions : 
- disable the test which checks alignment rejection
- disable the test but only on Windows 32-bit
- enable alignment check on Windows 32-bit
    + but then, it's reported as failing in some cases ?
        * note : I have not been able to reproduce this issue.

remittor:
> note : I have not been able to reproduce this issue.

It's just luck!
`malloc` returned pointer aligned by 8 bytes: `multiple valid alignments : hc1(00F9B020)`
You need to test on all Windows.

Cyan4973:
> It's just luck!
> `malloc` returned pointer aligned by 8 bytes: `multiple valid alignments : hc1(00F9B020)`

Start addresses of the tested memory regions are _manually_ assigned.
So traces show that, on 32-bit systems, both 32-bit and 64-bit alignment are present and tested.

lib/lz4hc.c
```diff
@@ -53,27 +53,37 @@
...
 
+/* for some reason, Visual Studio fails the aligment test on 32-bit x86 :
+ * it reports an aligment of 8-bytes,
+ * while LZ4_streamHC_t only requires alignment of 4-bytes
+ * resulting in initialization error when allocating state with malloc() */
+#if (defined(_MSC_VER) && !defined(_M_X64))
+#  define LZ4_ALIGN_TEST 0
+#else
+#  define LZ4_ALIGN_TEST 1
+#endif
```
remittor, line 86: As it turned out, this problem is relevant not only for MSVC. 
The problem is not with compilers, but with operating systems where the memory manager returns pointers aligned to 4 bytes.
Therefore, design `LZ4_ALIGN_TEST` must always be switch on.

tests/fuzzer.c
```diff
@@ -1200,6 +1200,28 @@ static void FUZ_unitTests(int compressionLevel)
...
 
+    DISPLAYLEVEL(3, "LZ4_initStreamHC with multiple valid alignments : ");
+    {   typedef struct {
+            LZ4_streamHC_t hc1;
+            char           c1;
+            LZ4_streamHC_t hc2;
+            char           c2;
+            LZ4_streamHC_t hc3;
+        } shct;
+        shct* const shc = (shct*)malloc(sizeof(*shc));
```
remittor, line 1229: At this point, you need to simulate the worst case:
```
BYTE* mem = (BYTE*)malloc(sizeof(shct) + 256);
assert(mem != NULL);
mem &= ~(0x0F);
mem += 16 + sizeof(void*);
shct* const shc = (const shct*)mem;
```

remittor (changes requested):
Related issue: https://github.com/lz4/lz4/issues/827#issuecomment-719983334

tests/fuzzer.c
```diff
@@ -1200,6 +1200,28 @@ static void FUZ_unitTests(int compressionLevel)
...
+        assert(shc != NULL);
+        memset(shc, 0, sizeof(*shc));
+        DISPLAYLEVEL(3, "hc1(%p) hc2(%p) hc3(%p) size(0x%x): ",
+                    &(shc->hc1), &(shc->hc2), &(shc->hc3), (unsigned)sizeof(LZ4_streamHC_t));
+        FUZ_CHECKTEST( LZ4_initStreamHC(&(shc->hc1), sizeof(shc->hc1)) == NULL, "hc1 (%p) failed init", &(shc->hc1) );
+        FUZ_CHECKTEST( LZ4_initStreamHC(&(shc->hc2), sizeof(shc->hc2)) == NULL, "hc2 (%p) failed init", &(shc->hc2)  );
+        FUZ_CHECKTEST( LZ4_initStreamHC(&(shc->hc3), sizeof(shc->hc3)) == NULL, "hc3 (%p) failed init", &(shc->hc3)  );
+        FUZ_CHECKTEST( LZ4_initStreamHC((char*)&(shc->hc1) + 1, sizeof(shc->hc1)) != NULL,
+                        "hc1+1 (%p) init must fail, due to bad alignment", (char*)&(shc->hc1) + 1 );
```
remittor, line 1220: This line of code is unnecessary.

remittor (changes requested):
Related issue: https://github.com/lz4/lz4/issues/827#issuecomment-719983334

lib/lz4.c
```diff
@@ -178,6 +178,12 @@
 #define unlikely(expr)   expect((expr) != 0, 0)
 #endif
 
+/* Should the alignment test prove unreliable, for some reason,
+ * it can be disabled by setting LZ4_ALIGN_TEST to 0 */
+#ifndef LZ4_ALIGN_TEST  /* can be externally provided */
+# define LZ4_ALIGN_TEST 1
+#endif
```
remittor, line 185: This is a very, very controversial decision! It will be quite difficult to get people using machines and compilers from the first list https://github.com/lz4/lz4/issues/827#issuecomment-719983334 to specify this definition at compilation.

lib/lz4.h
```diff
@@ -573,68 +573,60 @@ LZ4LIB_STATIC_API void LZ4_attach_dictionary(LZ4_stream_t* workingStream, const
...
  *  The structure definition can be convenient for static allocation
  *  (on stack, or as part of larger structure).
  *  Init this structure with LZ4_initStream() before first use.
  *  note : only use this definition in association with static linking !
- *    this definition is not API/ABI safe, and may change in a future version.
+ *  this definition is not API/ABI safe, and may change in future versions.
  */
-#define LZ4_STREAMSIZE_U64 ((1 << (LZ4_MEMORY_USAGE-3)) + 4 + ((sizeof(void*)==16) ? 4 : 0) /*AS-400*/ )
-#define LZ4_STREAMSIZE     (LZ4_STREAMSIZE_U64 * sizeof(unsigned long long))
+#define LZ4_STREAMSIZE       16416  /* static size, for inter-version compatibility */
```
remittor, line 623: Probably better to replace with this:
`#define LZ4_STREAMSIZE       ( (sizeof(LZ4_stream_t_internal) + sizeof(void*) - 1) & ~(sizeof(void*)-1) )`

lib/lz4hc.h
```diff
@@ -198,57 +198,32 @@ LZ4LIB_API int LZ4_saveDictHC (LZ4_streamHC_t* streamHCPtr, char* safeBuffer, in
...
 
-#endif
-
 
 /* Do not use these definitions directly !
  * Declare or allocate an LZ4_streamHC_t instead.
  */
-#define LZ4_STREAMHCSIZE       (4*LZ4HC_HASHTABLESIZE + 2*LZ4HC_MAXD + 56 + ((sizeof(void*)==16) ? 56 : 0) /* AS400*/ ) /* 262200 or 262256*/
-#define LZ4_STREAMHCSIZE_SIZET (LZ4_STREAMHCSIZE / sizeof(size_t))
+#define LZ4_STREAMHCSIZE       262200  /* static size, for inter-version compatibility */
```
remittor, line 223: Probably better to replace with this:
`#define LZ4_STREAMHCSIZE       ( (sizeof(LZ4HC_CCtx_internal) + sizeof(void*) - 1) & ~(sizeof(void*)-1) )`

lib/lz4.h
Cyan4973: The point here is to ensure that `LZ4_stream_t` has a certain minimum size
so that if `LZ4_stream_t_internal` evolves and becomes smaller in the future,
it would still allocate enough to remain compatible with older versions of the library.

PR #941, opened 2020-11-07
Revert "Replace "static" to "LZ4_FORCE_INLINE" for small functions"

Cyan4973:
This reverts commit 0e3933edd435c54cc2e21e38f5d4ba7bf644a24e ,
which resulted in important performance regression.

Reverting the commit restores performance (`gcc v9.3.0`, i7-9700k noturbo) :

Silesia C.Speed : 513 MB/s ==> 532 MB/s  (+3.7%)
Calgary C.Speed : 426 MB/s ==> 443 MB/s  (+4.0%)

Silesia D.Speed : 3263 MB/s ==> 3386 MB/s  (+3.8%)
Calgary D.Speed : 3190 MB/s ==> 3326 MB/s  (+4.3%)

There might be a case for more selective inlining, but this would need to be more thoroughly analyzed.
On short term, just revert the commit, which was part of PR #930 (note: PR itself wasn't cancelled, other parts of this PR are fine).

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -310,7 +310,7 @@ typedef enum {
 #define LZ4_memcpy(dst, src, size) memcpy(dst, src, size)
 #endif
 
-LZ4_FORCE_INLINE unsigned LZ4_isLittleEndian(void)
+static unsigned LZ4_isLittleEndian(void)
 {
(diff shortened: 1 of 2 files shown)


Cyan4973:
Follow up : performance difference is likely a side-effect of instruction alignment (which is more or less random).

PR #960, opened 2020-12-01
Makefile correctly export CFLAGS

### LINKED ISSUE #958: CFLAGS env var ignored by makefile
**Describe the bug**

We set some debug flags (`-O3 -fdebug-prefix-map=... -g`) during our build as env var.
Only `$CXXFLAGS` works.
`$CXXFLAGS` seems to be ignored.

image

Work-around is to use `make all CFLAGS="$CFLAGS"` but I don't know why.

image

And with the work-around, those `-W` from `$(DEBUGFLAGS)` disappeared.

 - OS: [e.g. Mac] CentOS 7
 - Version [e.g. 22] 1.9.3
 - Compiler [e.g. gcc] gcc-9 (devtoolset-9)
 - Build System [e.g. Makefile] gnu make 3.82

Cyan4973:
It seems that forwarding of `CFLAGS` environment variable to sub-Makefiles was lost at some point,
and it seems related to a `make` bug, which accidentally `unexport` global variables
with the same name as a Target-specific Variable which is conditionally assigned: http://savannah.gnu.org/bugs/?func=detailitem&item_id=59230

fix #958

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ jobs:
     #   This would typically be a build job when using workflows, possibly combined with build
     # This is based on your 1.0 configuration file or project settings
     - run: CFLAGS= make clangtest && make clean
-    - run: g++ -v; make gpptest     && make clean
+    - run: g++ -v; make cxxtest   && make clean
     - run: gcc -v; g++ -v; make ctocpptest && make clean
     - run: gcc-5 -v; CC=gcc-5 CFLAGS="-O2 -Werror" make check && make clean
     - run: gcc-5 -v; CC=gcc-5 CFLAGS="-O2 -m32 -Werror" CPPFLAGS=-I/usr/include/x86_64-linux-gnu make check && make clean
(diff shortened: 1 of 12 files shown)


xkszltl:
Did you find any bug report in upstream?

xkszltl:
Are those removal `$(Q)` and `@` intentional?

Cyan4973:
Yes (http://savannah.gnu.org/bugs/?func=detailitem&item_id=59230) and yes

Cyan4973:
cc @terrelln:
I think I will need some help to understand and maybe fix this issue : 
Compile OSS-Fuzz targets : https://travis-ci.org/github/lz4/lz4/jobs/747016288

This part of the code base wasn't modified by this PR,
so something else might have changed on the OSS-Fuzz side.
I'm not used to these traces, but I guess the failing command might be : 
`docker build --pull --no-cache -t gcr.io/oss-fuzz/lz4 projects/lz4`

terrelln:
Seems like an oss-fuzz issue. We can give it a day or two and see if it goes away. If it persists we can open an issue on oss-fuzz.

PR #973, opened 2021-03-01
Avoiding null pointer dereference

klebertarcisio:
This pull request aims to fix null pointer dereference.

[Null pointer dereference](https://cwe.mitre.org/data/definitions/401.html)


Cyan4973:
Yes, that's a good point.

Issue here is `C90` compliance (which is the standard used by `lz4`) : one can't add a statement before a declaration within the same block.

This is a minor issue, there are multiple ways to get this fixed.

klebertarcisio:
Hi @Cyan4973, thanks for the feedback. I reversed the commit to send another solution considering C90.

Cyan4973 (approved):

Cyan4973:
Thanks @klebertarcisio !

PR #978, opened 2021-05-14
disable allocs if HEAPMODE=0

ITotalJustice:
this PR #ifdefs out all alloc functions if HEAPMODE=0.

this PR also fixes a bug with lz4hc `#ifdef LZ4HC_HEAPMODE` which should instead be
`#if (LZ4HC_HEAPMODE)`. due to LZ4HC_HEAPMODE always being defined, even if set to 0, the heap routines will always be selected.

I am not sure if this PR is something you would accept, if not, i can create a seperate PR for the bug i mentioned above instead. Thanks!

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -187,6 +187,7 @@
 /*-************************************
 *  Memory routines
 **************************************/
+#if (LZ4_HEAPMODE)
 #ifdef LZ4_USER_MEMORY_FUNCTIONS
 /* memory management functions can be customized by user project.
  * Below functions must exist somewhere in the Project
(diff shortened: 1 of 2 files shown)


Cyan4973:
This PR intention is good,
but as can be seen from CI failures,
the execution is not correct.

I would suggest to separate both topics mentioned in the PR,
with the issue regarding `LZ4HC_HEAPMODE` being higher priority,
since it fails to achieve the intended effect.

On the other hand, removing alloc functions is a more complex topic, 
which should be handled with care, as it has deep ramifications.

ITotalJustice:
Thanks for the reply. I’ll open a new PR shortly to address the HEAPMODE macro bug

t-mat:
Hi, @ITotalJustice

Since recently we've removed many tests from `.travis.yml` and introduce GitHub Actions, this PR (and your `disable-allocs` branch) may fail CI tests even if you'll fix `LZ4HC_HEAPMODE` issue properly.

I'm really sorry about this situation.  But I highly recommend you to git-merge/rebase or cherry-pick from recent `dev` branch to ease PR merging process.

I'm just guessing you're familiar with `git` and GitHub.  But if you don't know about Git commands, feel free to ask us how properly to do it.

Issue #979, opened 2021-05-20
CMake 3.19.0 deprecation warning


sandyharvie:
See https://cmake.org/cmake/help/latest/release/3.19.html#deprecated-and-removed-features:

> Compatibility with versions of CMake older than 2.8.12 is now deprecated and will be removed from a future version. Calls to cmake_minimum_required() or cmake_policy() that set the policy version to an older value now issue a deprecation diagnostic.

Solution is to bump CMake minimum required version (specified [here](https://github.com/lz4/lz4/blob/dev/build/cmake/CMakeLists.txt#L37)) to 2.8.12.

t-mat:
Hi, @sandyharvie

It was unfortunate, but [CI test for this PR has failed](https://travis-ci.org/github/lz4/lz4/jobs/771831046#L18) since we're using old version of Ubuntu at travis-ci.org for compatibility.

```
Description:    Ubuntu 12.04.5 LTS
```

The test had failed [due to old version of CMake](https://travis-ci.org/github/lz4/lz4/jobs/771831046#L153).

```
CMake Error at CMakeLists.txt:37 (cmake_minimum_required):
  CMake 2.8.12 or higher is required.  You are running version 2.8.7
```

Now, we're introducing GitHub Actions with the recent version of Ubuntu.  So, I think we'll be able to merge your PR soon.

t-mat:
Since this #980 has been merged successfully, I'd like to close this issue.

PR #980, opened 2021-05-20
Fix issue #979

sandyharvie:
See #979 for more details.

### DIFF
--- a/build/cmake/CMakeLists.txt
+++ b/build/cmake/CMakeLists.txt
@@ -34,7 +34,7 @@ else()
     LANGUAGES C)
 endif()
 
-cmake_minimum_required (VERSION 2.8.6)
+cmake_minimum_required (VERSION 2.8.12)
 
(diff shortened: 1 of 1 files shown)


Cyan4973:
Since @t-mat updated the CI tests,
I presume this change should no longer fail CI.
Merging this change also presumes we do no longer intend to support `cmake` on Ubuntu Precise (and older).

t-mat:
For just in case, I'd like to point out about [failure of the test for this PR](https://travis-ci.org/github/lz4/lz4/builds/771831040).
This failure is normalcy.  Because this PR contains [old `.travis.yml`](https://github.com/sandyharvie/lz4/blob/7da3ab4bec3e38651ce53d660746d2b9d144d6de/.travis.yml).

I'll post to recommend other contributors to git-rebase/merge/cherry-pick from recent `dev` branch for their PRs.

PR #983, opened 2021-05-27
Fix usan runtime errors

t-mat:
This is a follow up PR of #982.
This PR fixes `make usan` runtime errors.

Majority of usan runtime errors are two types.

(1) Applying pointer arithmetic to NULL pointer.  (undefined behaviour)

```C
- const BYTE* p = x + 1;            // usan runtime error when `x` is NULL.
+ const BYTE* p = (const BYTE*) ((uptrval) x + 1);
```

(2) Result of pointer arithmetic is overflowed.  (undefined behaviour)

```C
- const BYTE* p = x - 0x4000;      // usan runtime error when `x` is less than 0x4000.
+ const BYTE* p = (const BYTE*) ((uptrval) x - 0x4000);
```

Patch for `LZ4_resetStreamHC_fast()` is just a bit cumbersome.  For this line

```C
LZ4_streamHCPtr->internal_donotuse.end -= (uptrval) LZ4_streamHCPtr->internal_donotuse.base;
```

`usan` claims

> applying non-zero offset to non-null pointer 0xXXXXXXXXXXXXXXXX produced null pointer

So, we need to cast both of pointers to `uptrval` and cast to `const LZ4_byte*`.

```C
LZ4_streamHCPtr->internal_donotuse.end =
    (const LZ4_byte*) (  (uptrval) LZ4_streamHCPtr->internal_donotuse.end
                       - (uptrval) LZ4_streamHCPtr->internal_donotuse.base  );
```

PR #985, opened 2021-05-27
Add GitHub Actions script ci.yml

t-mat:
This PR works well with PR #982, #983, #984.

This PR adds GH-Actions CI script which includes a majority of the `.travis.yml` tests except the following:

- May be added later
  - Compile OSS-Fuzz targets
  - tag build
- Doesn't include due to lack of `apt` package for Ubuntu 20.04.
  - (Trusty) gcc-4.4 compilation
  - (Xenial) gcc-5 compilation
  - (Trusty) clang-3.8 compilation
- Doesn't include due to lack of native hardware support of GitHub Actions.  But I've added QEMU for them.
  - aarch64 real-hw tests
  - PPC64LE real-hw tests
  - IBM s390x real-hw tests

Please review [an actual result](https://github.com/t-mat/lz4/actions/runs/882090375).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,462 @@
+# Known issues:
+# - This test script ignores exit code of cppcheck which can see under
+#   Job:Linux x64 scan-build + cppcheck in the GitHub Actions report.
+#   Because this project doesn't 100% follow their recommendation.
+#   Also sometimes it reports false positives.
+#
+# Notes:
+# - You can investigate various information at the right pane of GitHub
+#   Actions report page.
+#
+#   | Item                      | Section in the right pane             |
+#   | ------------------------- | ------------------------------------- |
+#   | OS, VM                    | Set up job                            |
+#   | git repo, commit hash     | Run actions/checkout@v2               |
+#   | gcc, tools                | Environment info                      |
+#
+# - To fail earlier, order of tests in the same job are roughly sorted by
+#   elapsed time.
+#
+# Tests which we still leave at travis-ci
+# - name: Compile OSS-Fuzz targets
+# - name: tag build
+# - name: aarch64 real-hw tests
+# - name: PPC64LE real-hw tests
+# - name: IBM s390x real-hw tests
+
+
+# Name of the workflow is also displayed as a SVG badge
+name: lz4 CI
+
+on: [push, pull_request]
+
+jobs:
+  # CI Environment information
+  #
+  # - "ubuntu-latest" (Ubuntu 20.04) has the following software
+  #   https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md
+  basic-env-info:
+    name: Linux x64 Environment Info (ubuntu-latest)
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: gcc
+      run: echo && gcc --version
+
+    - name: clang
+      run: echo && clang --version
+
+    - name: make
+      run: echo && make -v
+
+    - name: gcc packages
+      run: apt-cache search gcc | grep "^gcc-[0-9\.]* " | sort
+
+    - name: clang packages
+      run: apt-cache search clang | grep "^clang-[0-9\.]* " | sort
+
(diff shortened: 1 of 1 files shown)


Cyan4973:
> Doesn't include due to lack of apt package (...) gcc-4.4 compilation

We have an informal policy to support compilers up to 10 years.
`gcc-4.4` is crossing this limit.
It doesn't mean that we have to actively remove this compiler,
but if it's troublesome to keep checking it,
then it's good ground to let it go from test scope.

Cyan4973 (approved):

Cyan4973:
Thanks for this CI setup improvement @t-mat .
I presume it should be merged after #982, #983 and #984 ?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,462 @@
...
+
+
+  ubuntu-usan:
+    name: Linux x64 usan
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2 # https://github.com/actions/checkout
+    - name: usan
+      run: |
+        make clean usan MOREFLAGS='-Wcomma -Werror'
```
t-mat: We can change this line with

```C
make clean usan MOREFLAGS='-Wcomma -Werror' || echo Ignore these errors for now.  For details, see https://github.com/lz4/lz4/pull/983
```

t-mat:
> I presume it should be merged after #982, #983 and #984 ?

Right.  Without these patches, all GH-Actions tests will be failed.
But if you prefer to operate GH-Actions CI earlier, and `usan` issue may take long ime, we temporary ignore some results.
Since travis-ci closure is approaching, this "better than nothing" approach may be realistic for now.

Personally, I'm also worrying that these PRs are burden for you.


> We have an informal policy to support compilers up to 10 years.
> gcc-4.4 is crossing this limit.

Sure.  But for curious, I'd like to know a bit more about details.

(1) Do you mean '10 years' from the last release date of `gcc-major.minor.patch`, or the first release date of `gcc-major` ?

(2) Should we ideally test all `gcc-major.minor` version, or `gcc-major`?

Cyan4973:
> if you prefer to operate GH-Actions CI earlier, and usan issue may take long ime, we temporary ignore some results.

I agree.
Also by "ignore", I mean using the right set of flags to avoid having this test becoming red.

The problem with red tests is that they considerably reduce the strength of CI results, as users get accustomed to have some red nonetheless. Overtime, it leads to more and more pbs remaining unaddressed.

> Do you mean '10 years' from the last release date of gcc-major.minor.patch, or the first release date of gcc-major ?

It's more 10 years from first release,
but it's also important to remind that this is just ballpark here.
The point is not to act at 10 years + 1 month,
but just to acknowledge that every support and test case is a burden,
and that it's acceptable after a reasonable amount of time to let go some of them.

t-mat:
I changed `ci.yml` to ignore the exit code of `make usan` temporary.  It may be useful to double checking `make usan` related commits/PRs.

Since #982 has merged, now this PR is reay to work.

Cyan4973 (approved):

t-mat:
I'm sorry for spamming many PRs.  But my intention is reducing jobs in `.travis.yml`.  It may help us when we decide to migrate to travis-ci.com for some unknown reason, as a last resort.

FYI, here's a rough estimation:

- On average, lz4/lz4 has 33 commits/month (`ceil(2700.0/((2021-2014)*12))`).
- It takes [120 machine minutes/commit](https://travis-ci.org/github/lz4/lz4/builds/769024841).
- It consumes [10 credits/minute](https://docs.travis-ci.com/user/billing-overview/#usage---credits) for Linux test runner.
- It may need 39,600 (`33*120*10`) credits/month.
- It may exceed their [10,000 credits/month for "Free Plan"](https://travis-ci.com/plans).

Cyan4973:
Yes, we have to considerably reduce the load on travis,
only keep the few tests that can't be done elsewhere (on github actions)
and offer some unique value difficult to reproduce elsewhere.

t-mat:
So far, I'm checking build log of travis-ci and GH-Actions.
To make reverting process easier, I'll keep posting small PRs in revert-able manner.

PR #988, opened 2021-05-28
Remove CMake test from .travis.yml

### LINKED ISSUE #979: CMake 3.19.0 deprecation warning

t-mat:
Since we've confirmed CMake test has [successfully passed at GitHub Actions](https://github.com/lz4/lz4/runs/2692005339?check_suite_focus=true), we can safely remove it from .travis.yml.

It also helps to merge #980.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,13 +41,12 @@ matrix:
         - make clean
         - CC="c++ -Wno-deprecated" make -C tests fullbench-wmalloc  # stricter function signature check
 
-    - name: (Precise) g++ and clang CMake test
+    - name: (Precise) g++ and clang test
       dist: precise
       script:
         - make cxxtest
         - make clean
         - make examples
-        - make clean cmake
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Cyan4973:
Any reason you preferred to close this PR rather than merge it ?

t-mat:
oops.  It's my mistake.  Sorry 🙏

PR #989, opened 2021-05-28
Fix NULL ptr arithmetic in lz4.c

Cyan4973:
This PR is meant to fix all `NULL` ptr arithmetic within `lz4.c`.

That's it's sole scope.
It doesn't fix ptr arithmetic overflow.
It also doesn't try to fix issues in other units, such as `lz4hc.c` or `lz4frame.c`.

Waiting for test results, to ensure it doesn't introduce new bugs of its own.
_edit_ : well, all seems fine.


tamird:
We're seeing this trip in Fuchsia in `LZ4_compress_fast_continue` (see red in https://fuchsia-review.googlesource.com/c/fuchsia/+/574484) -- is there expected to be a release containing this fix in the near future?

Cyan4973:
There is no new release planned in the short term.
If you absolutely need this fix now, I would suggest either cherry-picking this PR, or use the `dev` branch in its current state.

PR #999, opened 2021-06-02
add -lrt on Solaris 10, required for nanosleep

pekdon:
To get access to nanosleep on Solaris 10, link against rt is required.

### DIFF
--- a/programs/Makefile
+++ b/programs/Makefile
@@ -52,13 +52,19 @@ DEBUGFLAGS= -Wall -Wextra -Wundef -Wcast-qual -Wcast-align -Wshadow \
             -Wswitch-enum -Wdeclaration-after-statement -Wstrict-prototypes \
             -Wpointer-arith -Wstrict-aliasing=1
 CFLAGS   += $(DEBUGFLAGS) $(MOREFLAGS)
+
+include ../Makefile.inc
+
+ifeq ($(TARGET_OS)$(shell uname -r),SunOS5.10)
+LDFLAGS  += -lrt
+endif
+
 FLAGS     = $(CFLAGS) $(CPPFLAGS) $(LDFLAGS)
 
 LZ4_VERSION=$(LIBVER)
 MD2ROFF   = ronn
 MD2ROFF_FLAGS = --roff --warnings --manual="User Commands" --organization="lz4 $(LZ4_VERSION)"
 
-include ../Makefile.inc
 
 default: lz4-release
 


programs/Makefile
```diff
@@ -52,13 +52,19 @@ DEBUGFLAGS= -Wall -Wextra -Wundef -Wcast-qual -Wcast-align -Wshadow \
             -Wswitch-enum -Wdeclaration-after-statement -Wstrict-prototypes \
             -Wpointer-arith -Wstrict-aliasing=1
 CFLAGS   += $(DEBUGFLAGS) $(MOREFLAGS)
+
+include ../Makefile.inc
+
+ifeq ($(TARGET_OS)$(shell uname -r),SunOS5.10)
```
Cyan4973, line 58: What's the purpose of listing both `TARGET_OS` and `$(shell uname -r)`

since `TARGET_OS` is already defined as `TARGET_OS ?= $(shell uname)` ?

_edit_: ok, so `TARGET_OS` is the name, while `-r` is for the version.
Hence, it makes this patch very specific, to `SunOS5.10` only.

Any reason why it has to be specific to this version only (as opposed to `SunOS` in general) ?

pekdon: lz4 builds fine on Solaris 11 without it so I thought it was unnecessary to add it to all Solaris versions. I don't have access to any < 10 machines for testing either.

Cyan4973:
I'm a bit torn on this one.
On one hand, it's a simple change, well localized.
On the other hand, it's so specific, targeting a single version of a single (non-mainstream) OS, that I'm wondering if it should be part of general upstream, or part of a downstream patch dedicated to this platform only.

servusdei2018:
> I'm a bit torn on this one.
> On one hand, it's a simple change, well localized.
> On the other hand, it's so specific, targeting a single version of a single (non-mainstream) OS, that I'm wondering if it should be part of general upstream, or part of a downstream patch dedicated to this platform only.

I propose a solution in the form of a `solaris.mk` file with Solaris-specific directives, and to include that in `Makefile`. Inside `solaris.mk` would be the business logic of all Solaris-specific configuration. This pattern would be a good architecture in the future, in the event that another OS needs some specific configuration or tweaking, and splits things into a neat hierarchical division.

Possibly, such OS-specific configurations could also be transferred to a new directory called `make` or something like that, in apprehension of such files cluttering the root directory.

Cyan4973:
It looks like a good pattern @servusdei2018 .

It will require a bit of time to setup properly. Nothing too difficult, but time is scarce.
In the meantime, I'll merge this PR, on the ground that the logic is well concentrated in a single location.

PR #1001, opened 2021-06-15
Cast ALLOC return value to satisfy C++

GabeNI:


Cyan4973 (approved):

GabeNI:
Why are these CI builds failing?  What should I do about them?

t-mat:
Hi @GabeNI,
I think your PR is fine.

Those errors are caused by the following reasons.  I'll fix these tests soon.  Sorry!

- As for `gcc-11`
  - [`g++-11` has been removed](https://github.com/actions/virtual-environments/issues/3467) from GitHub Actions default environment.

- As for `gcc-{4.7, 4.6, 4.4}` and `clang-{3.8, 3.7, 3.6, 3.5}`
  - `apt-get` [failed to retrieve](https://github.com/lz4/lz4/pull/1001/checks?check_run_id=2832986574#step:3:62) some package.
  - We should add `apt-get update` for all `apt-get`s.

Cyan4973:
Yes, as stated by @t-mat , this is unrelated, and fixed in a separate PR

PR #1005, opened 2021-06-25
fix LZ4HC_HEAPMODE macro guard

ITotalJustice:
Hi, this commit fixes the macro guard incorrectly using `#ifdef LZ4HC_HEAPMODE` around heap related functions, even though `LZ4HC_HEAPMODE` will always be defined at the top of `lz4hc.c`

I used `#if defined(LZ4HC_HEAPMODE) && LZ4HC_HEAPMODE==1` because that's what was used in `LZ4HC_compress_optimal()` and i assume it's just to workaround users `#undef` or deleting the macro rather than setting it to zero


Cyan4973 (approved):

Cyan4973:
Thanks @ITotalJustice !

PR #1007, opened 2021-06-29
Add LZ4_free size param

GabeNI:
Certain environments have allocators that don't track allocation size and thus have deallocators that accept a size parameter  (e.g., IOMalloc/IOFree in macOS kernel).  This change will allow for a user-replaceable free function to accept an additional size parameter.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -193,10 +193,18 @@
  * and be available at link time */
 void* LZ4_malloc(size_t s);
 void* LZ4_calloc(size_t n, size_t s);
+#ifdef LZ4_FREE_ACCEPTS_SIZE
 void  LZ4_free(void* p, size_t s);
+#else
+void  LZ4_free(void* p);
+#endif
 # define ALLOC(s)          LZ4_malloc(s)
 # define ALLOC_AND_ZERO(s) LZ4_calloc(1,s)
+#ifdef LZ4_FREE_ACCEPTS_SIZE
 # define FREEMEM(p)        LZ4_free(p,sizeof(*(p)))
+#else
+# define FREEMEM(p)        LZ4_free(p)
+#endif
 #else
 # include <stdlib.h>   /* malloc, calloc, free */
 # define ALLOC(s)          malloc(s)
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -201,15 +201,15 @@ void  LZ4_free(void* p);
 # define ALLOC(s)          LZ4_malloc(s)
 # define ALLOC_AND_ZERO(s) LZ4_calloc(1,s)
 #ifdef LZ4_FREE_ACCEPTS_SIZE
-# define FREEMEM(p)        LZ4_free(p,sizeof(*(p)))
+# define FREEMEM(p,s)      LZ4_free((p),(s))
 #else
-# define FREEMEM(p)        LZ4_free(p)
+# define FREEMEM(p,s)      (void)(s); LZ4_free((p))
 #endif
 #else
 # include <stdlib.h>   /* malloc, calloc, free */
 # define ALLOC(s)          malloc(s)
 # define ALLOC_AND_ZERO(s) calloc(1,s)
-# define FREEMEM(p)        free(p)
+# define FREEMEM(p,s)      (void)(s); free((p))
 #endif
 
 #include <string.h>   /* memset, memcpy */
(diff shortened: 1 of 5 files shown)


lib/lz4.c
```diff
@@ -193,10 +193,18 @@
...
 void* LZ4_calloc(size_t n, size_t s);
+#ifdef LZ4_FREE_ACCEPTS_SIZE
+void  LZ4_free(void* p, size_t s);
+#else
 void  LZ4_free(void* p);
+#endif
 # define ALLOC(s)          LZ4_malloc(s)
 # define ALLOC_AND_ZERO(s) LZ4_calloc(1,s)
+#ifdef LZ4_FREE_ACCEPTS_SIZE
+# define FREEMEM(p)        LZ4_free(p,sizeof(*(p)))
```
Cyan4973: For the "size" parameter to be always correct with just this simple trick,
it implies that `p`  has always a well defined type, e.g. not incomplete, not `void*`.
While this is generally the case, there are a few exceptions.

In `lz4hc.c`, there is an old  (deprecated but still supported) function `LZ4_freeHC()` that's still using  a `void*`.
It would need to be updated, to ensure it propagates the correct size.

More difficult, in case of library amalgamation, this new definition of `FREEMEM()` is propagated to `lz4frame.c`.
In this library, `FREEMEM()` is used a lot more, and there is a need to ensure that every time, the right "size" parameter is passed. I'm skeptical that `sizeof(*(p))` get it always right, as the use cases are more complex in this unit.

GabeNI: I guess "this one simple trick" only works for clickbait ads and not for code updates.  😁 I'll rework the solution.  Thank you for your feedback.

t-mat:
I wrote [basic allocation test for lz4cli](https://gist.github.com/t-mat/1129122a6c9932737867be778fc5c3fd) to support this PR.

It can be built by the following command

```sh
cd
git clone https://github.com/lz4/lz4.git lz4-pr-1007
cd lz4-pr-1007
mkdir pr-1007
cd pr-1007
curl -JLO https://gist.github.com/t-mat/1129122a6c9932737867be778fc5c3fd/raw/f18356934d0d79b1bc141a94568cf177143c1815/lz4cli-alloc-test.cpp
g++ -O3 -std=c++17 -I ../lib lz4cli-alloc-test.cpp && ./a.out -f lz4cli-alloc-test.cpp
```

So far it doesn't have 2 arguments `LZ4_free()`, but I think you can add trivially.

Since lz4cli doesn't cover all LZ4/HC/F functions, coverage of this test is not so high.  But I hope it helps you to improve your PR.

GabeNI:
> 
> 
> I wrote [basic allocation test for lz4cli](https://gist.github.com/t-mat/1129122a6c9932737867be778fc5c3fd) to support this PR.

Thank you!  I likely won't have time to look at this before vacation, but I'll find it quite useful when I return to this change.

Cyan4973:
Any follow up to this PR ?

GabeNI:
Nothing yet, but you read my mind.  I was just looking back at it the day before you pinged.  I'll try to dig into it again this week.

GabeNI:
> Any follow up to this PR ?

I've started working on some updates and augmenting @t-mat 's basic allocation test to exercise the new code.  So far my results are mixed, so I'll need to spend some time debugging.

Cyan4973:
@t-mat , should we consider adding your suggested [`lz4cli-alloc-test.cpp`](https://gist.github.com/t-mat/1129122a6c9932737867be778fc5c3fd#file-lz4cli-alloc-test-cpp) to LZ4's test suite ?

t-mat:
Personally, I'd like to avoid to use it as a general/independent test code.

I think it's good companion code for custom allocator bacause it's tiny and easy to tweak.  But it's too rough as a general test in CI.  If we use it in GH-Actions, at least I want to see proper stack trace.

So, instead of it, I'd like to consider to introduce [LeakSanitizer](https://clang.llvm.org/docs/LeakSanitizer.html).  Because we're already familiar with its nice and clean log.
And we can use lz4-cli and all our examples as a test scenario of it.  When we'll add example for new API (or increasing coverage), it also means adding new test case.

Cyan4973:
> instead of it, I'd like to consider to introduce [LeakSanitizer](https://clang.llvm.org/docs/LeakSanitizer.html).

Great idea @t-mat !

Cyan4973:
Closing, due to lack of update.
Feel free to open a new PR when you have an updated proposal.

GabeNI:
Our need for this passed, so I won't be pursuing it further.  Thank you.

Issue #974, opened 2021-03-27
Conflict between LZ4_STREAMSIZE and LZ4_MEMORY_USAGE


mcroomp:
**Describe the bug**
Current LZ4_STREAMSIZE is hard coded to 16416, so if LZ4_MEMORY_USAGE is increased, LZ4_sizeofState returns a buffer that is too small to hold the hash table.

**Expected behavior**
LZ4_sizeofState should return a buffer that fits LZ4_MEMORY_USAGE. For example: ((1 << LZ4_MEMORY_USAGE) + 32)

**To Reproduce**
Increase LZ4_MEMORY_USAGE to 20. LZ4_sizeofState still returns 16416.

**Additional context**
There is a static assert

LZ4_STATIC_ASSERT(LZ4_STREAMSIZE >= sizeof(LZ4_stream_t_internal));    /* A compilation error here means LZ4_STREAMSIZE is not large enough */

eloj:
This sounded like some nice low-hanging fruit, so I took a look. Unfortunately I ran into problems I don't currently fully understand. 

 Making the obvious change (as suggested above) results in the following:

Setting MEMORY_USAGE in the range 14 (default) to 18 passes the tests.
Setting MEMORY_USAGE to 12 hits a test for compression ratio in the tests, which makes sense.
The problem shows itself when I try to go to 19 or above.

At that point we hit memory corruption when running `frame-test`, specifically:

`LZ4F_decompress_usingDict on multiple linked blocks : malloc(): corrupted top size`
or sometimes
`LZ4F_decompress_usingDict on multiple linked blocks : frametest: malloc.c:2379: sysmalloc: Assertion '(old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)' failed.`

So I did what you do in a situation like this; you run the test under valgrind:

```
LZ4F_compressFrame_usingCDict, multiple linked blocks : ==42084== Invalid write of size 8
==42084==    at 0x484296F: memset (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==42084==    by 0x411445: LZ4_initStream (in /home/eddy/dev/EXT/lz4/tests/frametest)
==42084==    by 0x402529: LZ4F_compressBegin_usingCDict (in /home/eddy/dev/EXT/lz4/tests/frametest)
[...]

PR #1013, opened 2021-07-31
Define LZ4_STREAMSIZE in terms of LZ4_MEMORY_USAGE.

eloj:
This is required to correctly size a static member
to hold the hash table, whose size is derived from
LZ4_MEMORY_USAGE.

At the same time, document that going beyond an exponent
of 18 (for 256KiB) is not supported at this point in time.

Adds static assert to enforce this for now.

### DIFF
--- a/lib/lz4frame.c
+++ b/lib/lz4frame.c
@@ -588,6 +588,16 @@ static void LZ4F_initStream(void* ctx,
     }
 }
 
+static int ctxTypeID_to_size(int ctxTypeID) {
+    switch(ctxTypeID) {
+    case 1:
+        return LZ4_sizeofState();
+    case 2:
+        return LZ4_sizeofStateHC();
+    default:
+        return 0;
+    }
+}
 
 /*! LZ4F_compressBegin_usingCDict() :
  *  init streaming compression AND writes frame header into @dstBuffer.
@@ -610,7 +620,9 @@ size_t LZ4F_compressBegin_usingCDict(LZ4F_cctx* cctxPtr,
 
     /* cctx Management */
     {   U16 const ctxTypeID = (cctxPtr->prefs.compressionLevel < LZ4HC_CLEVEL_MIN) ? 1 : 2;
-        if (cctxPtr->lz4CtxAlloc < ctxTypeID) {
+        int requiredSize = ctxTypeID_to_size(ctxTypeID);
+        int allocatedSize = ctxTypeID_to_size(cctxPtr->lz4CtxAlloc);
+        if (allocatedSize < requiredSize) {
             /* not enough space allocated */
             FREEMEM(cctxPtr->lz4CtxPtr);
             if (cctxPtr->prefs.compressionLevel < LZ4HC_CLEVEL_MIN) {
--- a/lib/lz4.h
+++ b/lib/lz4.h
@@ -620,7 +620,7 @@ typedef struct {
  *  note : only use this definition in association with static linking !
  *  this definition is not API/ABI safe, and may change in future versions.
  */
-#define LZ4_STREAMSIZE       16416  /* static size, for inter-version compatibility */
+#define LZ4_STREAMSIZE       ((1UL << LZ4_MEMORY_USAGE) + 32)  /* static size, for inter-version compatibility */
 #define LZ4_STREAMSIZE_VOIDP (LZ4_STREAMSIZE / sizeof(void*))
 union LZ4_stream_u {
     void* table[LZ4_STREAMSIZE_VOIDP];


eloj:
The assert isn't the best[0], but it's what I was comfortable with, and it should stop anyone from building the library with memory corruption by mistake.

[0] The size of the exponent isn't _actually_ the issue; it just exposes a different issue, and you can argue its placing. There should probably be an assert inside `LZ4_createStreamHC`, similar to the one that exists in `LZ4_createStream`, to check that the static size is at least as large as required by the struct allocated there.

t-mat:
As @eloj stated, I believe this is a good first step to resolve #974 and related (still hidden) issues.

I also think this is a good chance to introduce basic test for `LZ4_MEMORY_USAGE` in `tests/Makefile`.  For example,

```
	.PHONY: test
-	test: test-lz4 test-lz4c test-frametest test-fullbench test-fuzzer test-install test-amalgamation listTest test-decompress-partial
+	test: test-lz4 test-lz4c test-frametest test-fullbench test-fuzzer test-install test-amalgamation listTest test-decompress-partial test-compile-with-lz4-memory-usage

+	test-compile-with-lz4-memory-usage:
+		CFLAGS="-O2 -D'LZ4_MEMORY_USAGE=12'" $(MAKE) clean all
+		CFLAGS="-O2 -D'LZ4_MEMORY_USAGE=13'" $(MAKE) clean all
+		CFLAGS="-O2 -D'LZ4_MEMORY_USAGE=14'" $(MAKE) clean all
+		! CFLAGS="-O2 -D'LZ4_MEMORY_USAGE=15'" $(MAKE) clean all	# fail (without PR #1013)
+		! CFLAGS="-O2 -D'LZ4_MEMORY_USAGE=16'" $(MAKE) clean all	# fail (without PR #1013)
+		! CFLAGS="-O2 -D'LZ4_MEMORY_USAGE=17'" $(MAKE) clean all	# fail (without PR #1013)
+		! CFLAGS="-O2 -D'LZ4_MEMORY_USAGE=18'" $(MAKE) clean all	# fail (without PR #1013)
+		! CFLAGS="-O2 -D'LZ4_MEMORY_USAGE=19'" $(MAKE) clean all	# fail
+		! CFLAGS="-O2 -D'LZ4_MEMORY_USAGE=20'" $(MAKE) clean all	# fail
```

We will be able to add further tests, but I believe simple tests are good enough for this PR.

Cyan4973:
The issue is actually in `lz4frame` : https://github.com/lz4/lz4/blob/dev/lib/lz4frame.c#L611

The `LZ4F_cctx` states encapsulates an LZ4 state. Depending on the selected compression level, this state is either a `LZ4_streamHC_t` or a `LZ4_stream_t`. Once allocated, the state is re-used in place, to minimize impact on the allocator.

Re-using the state is fine as long as it's the same type.
When moving from `LZ4_stream_t` to `LZ4_streamHC_t`, the internal state is re-allocated.
But when moving from `LZ4_streamHC_t` to `LZ4_stream_t`, as consequence of reducing the compression level, the existing state is presumed larger, so its space is re-used, to avoid another `free`/`malloc` round trip, and only its content is changed, through re-initialization.

So that's where the problem lies : if `LZ4_MEMORY_USAGE` is too large, `LZ4_stream_t` becomes larger than `LZ4_streamHC_t`, and now the assumption do not hold anymore.

So that's where I would put a static assert.
Note that if you are not comfortable doing it, I can do it in a separate PR.
As a beneficial consequence, it would make it possible to use `lz4.c` with any `LZ4_MEMORY_USAGE` value, even > 18. Only `lz4frame.c` is not compatible with large `LZ4_MEMORY_USAGE` values.

Another possibility, more involved, the `lz4frame` code could be changed to avoid this situation entirely. But given how "frequent" this situation happens (almost never?), is it worth it ?

eloj:
Definitely do the better version of the PR. This was only ever a stop-gap anyway.

The coupling of lz4frame.c to lz4.c via the assert was one of the things I didn't like with this change, since it was clear the the issue was not actually with `LZ4_MEMORY_USAGE` in lz4.c, even if I didn't feel comfortable taking it all the way.

Feel free to close this PR at your convenience, or I'll close it later.

t-mat:
> Another possibility, more involved, the lz4frame code could be changed to avoid this situation entirely. But given how "frequent" this situation happens (almost never?), is it worth it ?

I'm not sure about frequency of the situation, honestly.  But I think it's worth enough for clarity and safety.  We can do it with this PR and the following change

```C
/* lz4/lib/lz4frame.c */
...
+   static int ctxTypeID_to_size(int ctxTypeID) {
+       switch(ctxTypeID) {
+       case 1:     return LZ4_sizeofState();
+       case 2:     return LZ4_sizeofStateHC();
+       default:    return 0;
+       }
+   }
...
    size_t LZ4F_compressBegin_usingCDict(...)
    {
        /* cctx Management */
        {   U16 const ctxTypeID = (cctxPtr->prefs.compressionLevel < LZ4HC_CLEVEL_MIN) ? 1 : 2;
+           int const requiredSize  = ctxTypeID_to_size(ctxTypeID);
+           int const allocatedSize = ctxTypeID_to_size(cctxPtr->lz4CtxAlloc);
-           if (cctxPtr->lz4CtxAlloc < ctxTypeID) {
+           if (allocatedSize < requiredSize) {
                /* not enough space allocated */
```

Since everything becomes static/inline code, I think overhead of this change is negligible.

This change passes the following tests

```sh
cd lz4/tests
CFLAGS="-O3 -D'LZ4_MEMORY_USAGE=13'" make clean test-mem	# OK
CFLAGS="-O3 -D'LZ4_MEMORY_USAGE=14'" make clean test-mem	# OK
CFLAGS="-O3 -D'LZ4_MEMORY_USAGE=18'" make clean test-mem	# OK
CFLAGS="-O3 -D'LZ4_MEMORY_USAGE=20'" make clean test-mem	# OK
```

But fails the following test

```sh
cd lz4/tests
CFLAGS="-O3 -D'LZ4_MEMORY_USAGE=12'" make clean test-mem	# FAIL (lz4frame tester)
```

```
Starting lz4frame tester (64-bits, 1.9.3)
Seed = 4537
...
LZ4F_compressFrame_usingCDict, with NULL dict : 33821 bytes
LZ4F_compressFrame_usingCDict, with dict : compressed 64512 bytes into 38101 bytes
cSizeWithDict (38101) should have been more compact than cSizeNoDict(33821)
Error detected !
```

When `LZ4_MEMORY_USAGE` is small, it also means it's not able to cover `dictSize` (63KiB).

eloj:
I took your (@t-mat) changes (excluding the test-related ones) and force pushed them on this branch.

eloj:
So just so I understand this right, there is no reinit required if the states have the same type? i.e the `else if (cctxPtr->lz4CtxState != ctxTypeID)` should remain a conditional?

t-mat:
> So just so I understand this right, there is no reinit required if the states have the same type?

Yes, we don't need full (re)initialization.

This is just a bit tricky.  Actual (re)initialization of the context is the following part of the code.

```C
size_t LZ4F_compressBegin_usingCDict(...)
{
    ...
    /* context init */
    cctxPtr->cdict = cdict;
    if (cctxPtr->prefs.frameInfo.blockMode == LZ4F_blockLinked) {
        /* frame init only for blockLinked : blockIndependent will be init at each block */
        LZ4F_initStream(cctxPtr->lz4CtxPtr, cdict, cctxPtr->prefs.compressionLevel, LZ4F_blockLinked);
    }
    ...
}
```

>  i.e the else if (cctxPtr->lz4CtxState != ctxTypeID) should remain a conditional?

Since `LZ4F_initStream()` invokes internal fast/partial context initialization functions, avoiding full initialization with this condition may give us some gain.

lib/lz4frame.c
```diff
@@ -588,6 +588,16 @@ static void LZ4F_initStream(void* ctx,
     }
 }
 
+static int ctxTypeID_to_size(int ctxTypeID) {
```
Cyan4973, line 591: tip: prefer using an `enum` type for such a parameter.
If named properly, the `enum` is much preferable for readability, hence for the maintenance of the source code.

lib/lz4frame.c
```diff
@@ -610,7 +620,9 @@ size_t LZ4F_compressBegin_usingCDict(LZ4F_cctx* cctxPtr,
 
     /* cctx Management */
     {   U16 const ctxTypeID = (cctxPtr->prefs.compressionLevel < LZ4HC_CLEVEL_MIN) ? 1 : 2;
-        if (cctxPtr->lz4CtxAlloc < ctxTypeID) {
+        int requiredSize = ctxTypeID_to_size(ctxTypeID);
+        int allocatedSize = ctxTypeID_to_size(cctxPtr->lz4CtxAlloc);
+        if (allocatedSize < requiredSize) {
```
Cyan4973, line 625: 👍

Cyan4973 (approved):

Cyan4973:
I like this code change.
It's in line with the original intention of the code, and just generalizes it for situations where `LZ4_MEMORY_USAGE` could be modified by the user.

t-mat:
I'd like to talk about test for this PR and failure of lz4frame tester for `LZ4_MEMORY_USAGE=12`.

```sh
cd lz4/tests
CFLAGS="-O3 -D'LZ4_MEMORY_USAGE=12'" make clean test-mem	# FAIL (lz4frame tester)
```

Should I open new issue or continue it here?

Cyan4973:
Yes please

PR #1014, opened 2021-07-31
GitHub Ubuntu 16.04 deprecation

eloj:
Removes ubuntu-16.04 as a test platform.
    
The Ubuntu 16.04 environment [is being removed by github](https://github.com/actions/virtual-environments/issues/3287) on September 20, 2021. They will induce 'brownouts' starting from September 6 to get clients to move on.
    
As a result of this change, there is no testing of GCC versions prior to 4.8, nor clang versions prior to 3.9

Since after this change all remaining compilers support both C90 and C11, I removed those columns in separate commits, and made those always run. If you'd rather keep either one or both columns, just say so and I'll rebase those commits.

Fixes a github workflows warning:
```
lz4 CI : .github#L1
The ubuntu-16.04 environment is deprecated and will be removed on September 20, 2021. Migrate to ubuntu-latest instead. For more details see https://github.com/actions/virtual-environments/issues/3287
```


Cyan4973 (approved):

Cyan4973:
Thanks @eloj ,
removing Ubuntu 16.04 is fine,
what is important is to continue testing `C90`-only mode,
which, if I understand correctly, your modification preserves.

t-mat:
> what is important is to continue testing C90-only mode,
> which, if I understand correctly, your modification preserves.

FYI, I've confirmed [`make c_standards_c90` works well on GH-Actions](https://github.com/lz4/lz4/runs/3210327683#step:7:80).

PR #1029, opened 2021-10-13
meson: Fix symbol visibility with GNU compilers

xclaesse:


tristan957:
Why can you not use `gnu_symbol_visibility` instead of applying the `-fvisibility` flag manually?

xclaesse:
It does the same thing. Since I already had to test `cc.has_argument('-fvisibility=hidden')` I thought it would be fine to just add it there instead of letting meson add it for me.

xclaesse:
Argh, unit tests uses private symbols, I think that's why it fails CI.

Cyan4973:
What follow-up could this PR have ?

tristan957:
@Cyan4973 I would be happy to take this PR to completion. @xclaesse is a busy man. I need some various Meson build improvements anyway, so I will take this and add whatever else I deem as useful. Just give me some time.

tristan957:
One thing that would be beneficial is that when my PR is merged, lz4 gets a new release, assuming the release would be compatible with 1.9.3

Cyan4973:
> One thing that would be beneficial is that when my PR is merged, lz4 gets a new release, assuming the release would be compatible with 1.9.3

I'll have to discuss that goal with the team. Building a release is a significant effort, and I can't guarantee that I can save the time for it starting next week. Though I will try.

What you can be sure though is that your PR will be welcome, and will be merged once completed.

tristan957:
> What you can be sure though is that your PR will be welcome, and will be merged once completed.

Thanks!

> I'll have to discuss that goal with the team. Building a release is a significant effort, and I can't guarantee that I can save the time for it starting next week. Though I will try.

I really just want the release because my project is trying to publish on Meson's WrapDB, and it can't be accepted until we can use the WrapDB version of lz4. Unfortunately we can't use the upstream WrapDB lz4 because it misses some performance optimizations like https://github.com/hse-project/hse/blob/master/subprojects/packagefiles/lz4/lib/meson.build#L5, which is necessary to keep lz4 from hitting the PLT.

I have already started on trying to get the Meson build revamped, but I think starting from scratch will be better. Lots of cruft since its maintenance doesn't seem like a priority. No problem though.

After writing this out, I think I could work around this with the upstream build, but I will still go ahead and submit a PR!

PR #1030, opened 2021-10-19
Create CMake package when installing

kostasdizas:
Made a small change to the CMake build files to create a package so it can be easily included in other projects.

### DIFF
--- a/build/cmake/CMakeLists.txt
+++ b/build/cmake/CMakeLists.txt
@@ -103,6 +103,9 @@ option(LZ4_POSITION_INDEPENDENT_LIB "Use position independent code for static li
 set(LZ4_LIBRARIES_BUILT)
 if(BUILD_SHARED_LIBS)
   add_library(lz4_shared SHARED ${LZ4_SOURCES})
+  target_include_directories(lz4_shared
+    PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+    INTERFACE $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
   set_target_properties(lz4_shared PROPERTIES
     OUTPUT_NAME lz4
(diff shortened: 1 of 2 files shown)


Cyan4973 (approved):

Cyan4973:
Thanks @kostasdizas ! Looks good to me !

mxmlnkn:
Thank you so much! Exactly what I needed after finding out that the CMake integration of v1.9.3 wasn't working as easily as expected.

build/cmake/CMakeLists.txt
```diff
@@ -202,6 +209,30 @@ if(NOT LZ4_BUNDLED_MODE)
...
+    COMPATIBILITY SameMajorVersion)
+
+  set(LZ4_PKG_INSTALLDIR "${CMAKE_INSTALL_LIBDIR}/cmake/lz4")
+  configure_package_config_file(
+    "${CMAKE_CURRENT_LIST_DIR}/lz4Config.cmake.in"
+    "${CMAKE_CURRENT_BINARY_DIR}/lz4Config.cmake"
+    INSTALL_DESTINATION ${LZ4_PKG_INSTALLDIR})
+  export(EXPORT lz4Targets
+    FILE ${CMAKE_CURRENT_BINARY_DIR}/lz4Targets.cmake
+    NAMESPACE LZ4::)
```
Tachi107, line 225: Why is the namespace upper case? The config file is called `lz4`, and exporting a namespace named differently is confusing, as it breaks the common practice used in the wider CMake ecosystem.

This also creates issues with pretty much every project who was already using lz4 in CMake, as people (quite unsurprisingly) use the lowercase `lz4` namespace, and package managers like [vcpkg do as well](https://github.com/microsoft/vcpkg/blob/42cc830d874f8e24eb4eb05bd5bb28496e3d498c/ports/lz4/CMakeLists.txt).

PR #1040, opened 2021-11-22
Enable fast decoding on Apple/AArch64 builds (18-25% faster decompression)

zeux:
This makes decoding significantly faster on M1; measured on compressed source
code, decompressing 294 MB to 1301 MB takes 513 ms (2.53 GB/s) before, and
406 ms (3.2 GB/s) after this change on M1 Pro.

There's no way to check if the target architecture is M1 specifically but the
gains are likely to be similar on recent iterations of Apple processors, and
the original performance issue was probably more specific to Qualcomm.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -421,10 +421,12 @@ static const int      dec64table[8] = {0, 0, 0, -1, -4,  1, 2, 3};
 #ifndef LZ4_FAST_DEC_LOOP
 #  if defined __i386__ || defined _M_IX86 || defined __x86_64__ || defined _M_X64
 #    define LZ4_FAST_DEC_LOOP 1
+#  elif defined(__aarch64__) && defined(__APPLE__)
+#    define LZ4_FAST_DEC_LOOP 1
 #  elif defined(__aarch64__) && !defined(__clang__)
-     /* On aarch64, we disable this optimization for clang because on certain
-      * mobile chipsets, performance is reduced with clang. For information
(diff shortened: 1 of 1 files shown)


Cyan4973:
Well, hopefully, the condition `defined(__aarch64__) && defined(__APPLE__)` should make it more specific to Apple Silicon implementations of ARM64, and therefore avoid bringing in Qualcomm's (and possibly other vendors) issues.

Cyan4973 (approved):

terrelln:
I measured on my M1 MacBook Air, and see decompression speed go from 3.9 GB/s -> 4.6 GB/s.

> Well, hopefully, the condition defined(__aarch64__) && defined(__APPLE__) should make it more specific to Apple Silicon implementations of ARM64, and therefore avoid bringing in Qualcomm's (and possibly other vendors) issues.

This will also enable it for older iPhones, which ran on Qualcomm chips. I'm not sure how much we care about that. Or even if older iPhones had the mentioned performance issues. It may have been Android devices.

zeux:
I wouldn't be too worried about this - I don't think Apple ever used Qualcomm CPUs, but the last ARM Cortex CPU they used was Apple A5 (launched in 2011 and discontinued in 2016). It's hard to know for certain without measuring this but I wouldn't expect this to cause a regression on Apple's more recent hardware.

Cyan4973:
Thanks @zeux, this looks like a good trade off

PR #708, opened 2019-05-10
Add multiframe report to --list command

gabrielstedman:
```
 » ~/lz4/tests ## list * make listTest
python3 test-lz4-list.py
2019/05/10 16:52:59 - Generating /tmp/test_list_5M
2019/05/10 16:52:59 - > ./datagen -g5M > /tmp/test_list_5M
2019/05/10 16:52:59 - > /home/gabriel/lz4/tests/../lz4 --content-size /tmp/test_list_5M /tmp/test_list_5M-lz4f-1f--content-size.lz4
2019/05/10 16:52:59 - > /home/gabriel/lz4/tests/../lz4 -BI /tmp/test_list_5M /tmp/test_list_5M-lz4f-1f-BI.lz4
2019/05/10 16:52:59 - > /home/gabriel/lz4/tests/../lz4 -BD /tmp/test_list_5M /tmp/test_list_5M-lz4f-1f-BD.lz4
2019/05/10 16:52:59 - > /home/gabriel/lz4/tests/../lz4 -BX /tmp/test_list_5M /tmp/test_list_5M-lz4f-1f-BX.lz4
2019/05/10 16:52:59 - > /home/gabriel/lz4/tests/../lz4 --no-frame-crc /tmp/test_list_5M /tmp/test_list_5M-lz4f-1f--no-frame-crc.lz4
2019/05/10 16:52:59 - > /home/gabriel/lz4/tests/../lz4 -l /tmp/test_list_5M /tmp/test_list_5M-legc-1f.lz4
2019/05/10 16:53:00 - Generating /tmp/test_list_20M
2019/05/10 16:53:00 - > ./datagen -g20M > /tmp/test_list_20M
2019/05/10 16:53:00 - > /home/gabriel/lz4/tests/../lz4 --content-size /tmp/test_list_20M /tmp/test_list_20M-lz4f-1f--content-size.lz4
2019/05/10 16:53:00 - > /home/gabriel/lz4/tests/../lz4 -BI /tmp/test_list_20M /tmp/test_list_20M-lz4f-1f-BI.lz4
2019/05/10 16:53:00 - > /home/gabriel/lz4/tests/../lz4 -BD /tmp/test_list_20M /tmp/test_list_20M-lz4f-1f-BD.lz4
2019/05/10 16:53:01 - > /home/gabriel/lz4/tests/../lz4 -BX /tmp/test_list_20M /tmp/test_list_20M-lz4f-1f-BX.lz4
[...]

PR #1048, opened 2021-12-14
'--list' doesn't work stdin even it is a regular file

Low-power:
This is a following up of my comments on #708.

Listing information from a compressed file (via `--list`) requires seeking the input file, however the stdin could be a regular file and therefore seekable (ignoring seekable device files this time); for example:
```
$ lz4 --list < regular-file
```
This change fixes it:
```
# programs/lz4 --list < test.lz4
    Frames           Type Block  Compressed  Uncompressed     Ratio   Filename
         1       LZ4Frame   B6I     182.42K             -         -   stdin
```

### DIFF
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -712,11 +712,6 @@ int main(int argc, const char** argv)
     }
 
     if (mode == om_list){
-        /* Exit if trying to read from stdin as this isn't supported in this mode */
-        if(!strcmp(input_filename, stdinmark)){
-            DISPLAYLEVEL(1, "refusing to read from standard input in --list mode\n");
-            exit(1);
-        }
         if(!multiple_inputs){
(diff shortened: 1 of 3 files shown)


Cyan4973:
If we are going to support `stdin` as valid input for `--list`,
I wonder why it would matter to limit it to regular files only.
We might as well allow any content provided from `stdin` ?

_Edit_ : ah, I get it, that's because of `fseek()` ...

Cyan4973:
We should probably add a test, to ensure that this use case is and remains properly supported.
It seems appropriate to add this test right after this line : https://github.com/lz4/lz4/blob/dev/tests/Makefile#L362 .
The new test could be doing the same thing, just receiving the input file from `stdin`.

Cyan4973:
merging to feature branch for adaptations

PR #1049, opened 2021-12-14
meson: simplify version lookup and avoid deprecation warnings

eli-schwartz:
run_command() in development versions of meson will warn when the `check: ` kwarg is not specified. At the same time, lz4 has some gnarly code to manually check the return code and raise an error if it failed.

Kill two birds with one stone, by making run_command inherently raise a fatal error when erroring out, then proceeding in the knowledge that it must have succeeded.


eli-schwartz:
Personally, I would recommend storing the version number in a buildsystem agnostic text file, e.g. `VERSION`. meson directly supports this (since 0.57) via

```
project('lz4', 'c', version: files('VERSION'))
```

Cyan4973 (approved):

PR #1052, opened 2022-01-19
 CMake: Fix target includes pointing to the wrong directory

leonvictor:
This fixes the shared and static targets include directories, now pointing to `lz4/lib` rather than `lz4/build/cmake`

### DIFF
--- a/build/cmake/CMakeLists.txt
+++ b/build/cmake/CMakeLists.txt
@@ -104,7 +104,7 @@ set(LZ4_LIBRARIES_BUILT)
 if(BUILD_SHARED_LIBS)
   add_library(lz4_shared SHARED ${LZ4_SOURCES})
   target_include_directories(lz4_shared
-    PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+    PUBLIC $<BUILD_INTERFACE:${LZ4_LIB_SOURCE_DIR}>
     INTERFACE $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
   set_target_properties(lz4_shared PROPERTIES
     OUTPUT_NAME lz4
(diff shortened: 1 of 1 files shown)


Cyan4973:
I'm trying to understand what's the matter to understand what is being fixed here.
Also, should the new behavior be checked by a test ?

leonvictor:
`target_include_directories(target PUBLIC $<BUILD_INTERFACE: dir1 dir2>)` is used to specify directories to include when building a given target (so roughly translates to `-Idir1 -Idir2` args in gcc)

The previous behavior was to include `CMAKE_CURRENT_SOURCE_DIR` which points to the location of the current cmake file, in this case `lz4/build/cmake`. This PR changes that to use `LZ4_SOURCE_DIR` instead, which points to our source directory. 

Actually this was already the behavior [in the current release](https://github.com/lz4/lz4/blob/d44371841a2f1728a3f36839fd4b7e872d0927d3/build/cmake/CMakeLists.txt#L80). This was changed with #1030 and it's probably a mistake due to the PR focusing on installation rather than build targets. 

I hope this clarifies things ! 

As for testing, it looks like the `cmake` job compiles with make rather than `CMake --build`. I'm a bit clueless about the use-cases of other build methods, but we could add a new job actually buildind through CMake ?

Cyan4973:
Thanks for the clarifications @leonvictor !

> As for testing, it looks like the `cmake` job compiles with make rather than `CMake --build`. I'm a bit clueless about the use-cases of other build methods, but we could add a new job actually buildind through CMake ?

I believe you are correct, and the CI test should rather use `cmake --build`.
Do you want to bundle that change in this PR ?
Alternatively, I could do it in another commit.

leonvictor:
Hey, it looks like I'm not authorized to make change to the CI, even on my fork. I've never used github workflows so I might be missing something, but it's alright if you can do it yourself 😄

PR #1054, opened 2022-01-27
Disable _tzcnt_u64 for ARM64EC

mcfi:
The ARM64EC is a new Microsoft-designed ARM64 ABI that is compatible with AMD64 code. However, not all AMD64 intrinsic functions are supported. For, intrinsics that are lowered to AVX, AVX2 and AVX512 instructions are not supported, including the _tzcnt_u64. To make sure this file compiles for ARM64EC, the use of _tzcnt_u64 should be neutered.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -517,7 +517,7 @@ static unsigned LZ4_NbCommonBytes (reg_t val)
     assert(val != 0);
     if (LZ4_isLittleEndian()) {
         if (sizeof(val) == 8) {
-#       if defined(_MSC_VER) && (_MSC_VER >= 1800) && defined(_M_AMD64) && !defined(LZ4_FORCE_SW_BITCOUNT)
+#       if defined(_MSC_VER) && (_MSC_VER >= 1800) && (defined(_M_AMD64) && !defined(_M_ARM64EC)) && !defined(LZ4_FORCE_SW_BITCOUNT)
 #         if defined(__clang__) && (__clang_major__ < 10)
             /* Avoid undefined clang-cl intrinics issue.
              * See https://github.com/lz4/lz4/pull/1017 for details. */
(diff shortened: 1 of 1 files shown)


Cyan4973:
Interesting,
would there be anyway to test this architecture in CI ?
`qemu` tests for example ?

mcfi:
Thanks. I don't know how the current qemu tests work, but if you have an environment with Visual Studio 2022 17.1 preview installed with all the ARM64EC individual components, you may be able to build the lib + all the tests. The tests need to be run on ARM64 windows 11 though.

Cyan4973:
> Thanks. I don't know how the current qemu tests work, but if you have an environment with Visual Studio 2022 17.1 preview installed with all the ARM64EC individual components, you may be able to build the lib + all the tests. The tests need to be run on ARM64 windows 11 though.

That's unlikely to be compatible with CI unfortunately.

mcfi:
That's unfortunate, but totally understandable, since the ARM64EC is still in its infancy stage right now. We are building opensource projects as ARM64EC and porting the code if needed. Similar changes have been committed to other projects. See [here ](https://github.com/Tencent/rapidjson/commit/060c348ea09a4b11df12342a7cb5995a94059c4e#diff-93983663cb1c2a9de5a1de5f9aff95913c2f755353df3f34daaa23de828bf259)for another example. :)

lib/lz4.c
```diff
@@ -517,7 +517,7 @@ static unsigned LZ4_NbCommonBytes (reg_t val)
     assert(val != 0);
     if (LZ4_isLittleEndian()) {
         if (sizeof(val) == 8) {
-#       if defined(_MSC_VER) && (_MSC_VER >= 1800) && defined(_M_AMD64) && !defined(LZ4_FORCE_SW_BITCOUNT)
+#       if defined(_MSC_VER) && (_MSC_VER >= 1800) && (defined(_M_AMD64) && !defined(_M_ARM64EC)) && !defined(LZ4_FORCE_SW_BITCOUNT)
```
Cyan4973, line 520: Could you please add a one-line comment explaining why `_M_ARM64EC` must be exclude ?
Essentially a summary of what you already explain in the PR, just hosted directly in the code source.

mcfi: Thanks, I just added comments.

Cyan4973 (approved):
Thanks for your contribution @mcfi !

mcfi:
Thank you for your review!

PR #1059, opened 2022-01-29
Ensure list test exits with appropiate exit code on failure

gabrielstedman:


gabrielstedman:
The failure under macOS proves the exit code is now correct and validation fails.

I don't have a macOS environment to troubleshoot and I'm unsure how to gain shell access to that specific github runner.  
I'll try to figure if there's a way to gain local access to this `macos-latest` instance.

gabrielstedman:
I'll close this whilst Investigating a solution

Cyan4973:
Thanks @gabrielstedman !

If I understand properly, there was no issue in the library itself ?
Just the test script needed be updated ?

gabrielstedman:
That's exactly right.

PR #1064, opened 2022-02-04
Fixup meson build

tristan957:
The meson build had gotten a little out of hand. It needed to be cleaned
up and have its errors fixed. This should enable lz4 to switch to Meson
at any time should the need ever arise.


tristan957:
@Cyan4973 Here is what I got regarding fixing Meson. The only thing that I didn't try to port were the tests. I compiled the executables, but the tests are so entrenched in the Makefile that I just lost motivation. I see we don't even run the tests on the Meson build, so I don't think it will be an issue.

CC @eli-schwartz

tristan957:
Wondering if I should do something like this: https://github.com/hse-project/hse/blob/master/subprojects/packagefiles/lz4/lib/meson.build#L5. That is my custom lz4 build.

contrib/meson/meson.build
```diff
@@ -11,11 +11,16 @@
...
 # (together with meson_options.txt) and packaged for wrapdb.
 
-project('lz4', ['c'],
-  license: ['BSD', 'GPLv2'],
-  default_options : ['c_std=c99',
-    'buildtype=release'],
+project(
+  'lz4',
+  ['c'],
+  license: 'BSD AND GPLv2',
```
tristan957: @Cyan4973 could you pick out the spdx identifiers from this list? https://spdx.org/licenses/ that I can put here?

contrib/meson/meson/programs/meson.build
```diff
@@ -7,46 +7,54 @@
...
+  include_directories: include_directories(lz4_source_root / 'programs'),
+  dependencies: [liblz4_dep],
+  export_dynamic: get_option('debug') and host_machine.system() == 'windows',
+  install: true
+)
+
+install_man(lz4_source_root / 'programs/lz4.1')
+
+if meson.version().version_compare('>=0.61.0')
+  install_symlink(
```
tristan957: @eli-schwartz could you review this usage of install_symlink?

tristan957:
Wondering if we could forego an lz4 release and just create a new release in wrapdb with the new Meson build. That would solve my issues.

tristan957:
I also left the original copyright headers in place although the code is vastly different. Do you want me to add myself to the contributor headers?

Cyan4973:
> I also left the original copyright headers in place although the code is vastly different. Do you want me to add myself to the contributor headers?

sounds fair

contrib/meson/meson.build
Cyan4973: The library is licensed as BSD, so it would be `BSD-2-Clause-Patent`,
but the CLI (Command Line Interface) is licensed as GPLv2, which would be `GPL-2.0-or-later`.

tristan957:
> Wondering if I should do something like this: https://github.com/hse-project/hse/blob/master/subprojects/packagefiles/lz4/lib/meson.build#L5. That is my custom lz4 build.

I don't think is necessary. I ran a test to see if an executable would link without lz4 APIs going through the PLT with a static lz4 library, and it seemed good to go.

tristan957:
Anyways I feel really good about this PR and I am happy to mark it off my list of things to do.

tristan957:
https://github.com/mesonbuild/wrapdb/pull/179#issuecomment-942543236

hmmm...

contrib/meson/meson/contrib/gen_manual/meson.build
```diff
@@ -1,43 +1,40 @@
...
-    lz4_version,
-    join_paths(meson.current_source_dir(), lz4_root_dir, 'lib/lz4frame.h'),
-    '@OUTPUT@'],
-  install : false)
+foreach mp : manual_pages
+  custom_target(
+    '@0@_manual.html'.format(mp),
+    build_by_default: true,
+    input: lz4_source_root / 'lib/@0@.h'.format(mp),
+    output: '@0@_manual.html'.format(mp),
```
eli-schwartz, line 33: The use of `@INPUT@` and the input kwarg is a definite improvement.

contrib/meson/meson/examples/meson.build
```diff
@@ -1,49 +1,48 @@
...
 #
 # This source code is licensed under both the BSD-style license (found in the
 # LICENSE file in the root directory of this source tree) and the GPLv2 (found
 # in the COPYING file in the root directory of this source tree).
 # #############################################################################
 
-lz4_root_dir = '../../../..'
+examples = {
+  'printVersion': {
+    'sources': files(lz4_source_root / 'examples/printVersion.c'),
```
eli-schwartz: This seems to be optimized for a case that won't come up. I think you can make this a lot simpler.

```meson

examples = {
  'printVersion': 'printVersion.c',
}
```
And later on...

contrib/meson/meson/examples/meson.build
```diff
@@ -1,49 +1,48 @@
...
-  dependencies: liblz4_dep,
-  install: false)
-simpleBuffer = executable('simpleBuffer',
-  join_paths(lz4_root_dir, 'examples/simple_buffer.c'),
-  dependencies: liblz4_dep,
-  install: false)
+foreach e, attrs : examples
+  executable(
+    e,
+    attrs.get('sources'),
```
eli-schwartz: ... you can do this:


```meson

foreach e, src : examples
  executable(
    e,
    lz4_source_root / 'examples' / src,
```

contrib/meson/meson/meson.build
```diff
@@ -8,105 +9,60 @@
...
-default_library = get_option('default_library')
+fs = import('fs')
 
-host_machine_os = host_machine.system()
-os_windows = 'windows'
-os_linux = 'linux'
-os_darwin = 'darwin'
-os_freebsd = 'freebsd'
-os_sun = 'sunos'
+lz4_source_root = fs.parent(fs.parent(fs.parent(meson.current_source_dir())))
```
eli-schwartz: You cannot use fs.parent, because the fs module is not available until meson 0.53, but before Meson 0.61 we forgot to add a FeatureNew warning about it.

However it is not strictly needed. Instead of every file using `../../../` with different levels, you can define it once, here, but using `../` instead of `fs.parent()`

`meson.current_source_dir() / '../../../'`

contrib/meson/meson/meson.build
```diff
@@ -8,105 +9,60 @@
...
-    common_warning_flags += ['-pedantic', '-Wno-long-long', '-Wno-variadic-macros']
-  elif c_std == 'c99' or c_std == 'gnu99'
-    common_warning_flags += ['-pedantic']
-  endif
-  cc_compile_flags = cc.get_supported_arguments(common_warning_flags)
-  add_project_arguments(cc_compile_flags, language: 'c')
+add_project_arguments('-DXXH_NAMESPACE=LZ4_', language: 'c')
+
+if get_option('debug')
+  add_project_arguments(cc.get_supported_arguments([
```
eli-schwartz, line 27: You dropped `-pedantic` and it's not implied by the current project default value for the `warning_level` option. I see you consolidated the individual flags from each target individually, though.

contrib/meson/meson/lib/meson.build
```diff
@@ -1,57 +1,81 @@
...
+c_args = []
+
+compile_args = []
+
+if host_machine.system() == 'windows' and default_library != 'static'
+  c_args += '-DLZ4_DLL_EXPORT=1'
+endif
+
+if get_option('memory-usage') > 0
+  compile_args += '-DLZ4_MEMORY_USAGE=@0@'.format(get_option('memory-usage'))
```
eli-schwartz: You set this elsewhere using add_project_arguments.

contrib/meson/meson/programs/meson.build
```diff
@@ -1,52 +1,61 @@
...
+  lz4_source_root / 'programs/bench.c',
+  lz4_source_root / 'programs/datagen.c',
+  lz4_source_root / 'programs/lz4cli.c',
+  lz4_source_root / 'programs/lz4io.c',
+)
+
+lz4 = executable(
+  'lz4',
+  sources,
+  objects: liblz4.extract_all_objects(recursive: true),
```
eli-schwartz: Why do you need to extract all objects here? Guessing because the build never used gnu symbol visibility before, *and* the executable is relying on symbols that aren't exported with LZ4LIB_API?

contrib/meson/meson/programs/meson.build
eli-schwartz: `pointing_to` needs to be a string, not a meson `executable()` object.

So, since you want to create `lz4c` as a program which is a symlink to the `lz4` program, just use `pointing_to: 'lz4'`. (Meson then creates a relative symlink that tries to find the lz4 destination in the same directory as the symlink itself.)

eli-schwartz: I'd use a foreach loop. `foreach alias: ['lz4c', 'lz4cat', 'unlz4']`, create two symlinks, one in the bindir pointing to `'lz4'`, one in the mandir named `alias + '.1'` pointing to `'lz4.1'`

contrib/meson/meson/programs/meson.build
```diff
@@ -1,52 +1,61 @@
...
+  export_dynamic: get_option('debug') and host_machine.system() == 'windows',
+  install: true
+)
+
+install_man(lz4_source_root / 'programs/lz4.1')
+
+if meson.version().version_compare('>=0.61.0')
+  install_symlink(
+    'lz4c',
+    install_dir: get_option('prefix') / get_option('bindir'),
```
eli-schwartz: A meson `install_dir` never needs to be joined with get_option('prefix'), since internally it already uses the equivalent of join_paths or the `/` operator  to join any relative value with the prefix.

You only need to join these together if you need the absolute path of e.g. bindir as a string, for example to print in a message or to add to a configuration_data() and write into a configured header file.

contrib/meson/meson/programs/meson.build
tristan957: Exactly. @Cyan4973 any thoughts on this?

contrib/meson/meson/lib/meson.build
tristan957: `compile_args` is exported in `declare_dependency(compile_args:)`. Not used for c_args. Unless you had a different intention with your comment?

contrib/meson/meson/programs/meson.build
eli-schwartz: It seems a bit unusual to only create the aliases when the user has a new enough version of Meson. Though there is no denying it's a lot cleaner to use the Meson function.

contrib/meson/meson/lib/meson.build
eli-schwartz: No, I just mistook what its purpose was here. Sorry!

...

Does that mean this needs to be in the generated pkgconfig file too?

tristan957: @Cyan4973 would be the authority here. Maybe the behavior I wrote is not what should be exhibited.

contrib/meson/meson/programs/meson.build
tristan957: Since this is a community contributed port and the symlinks are deprecated as far as I can tell, I just wanted to say screw it and use the Meson functions.

eli-schwartz: Yeah, it's not super important for them to exist I guess so having it be less well supported is probably fine.

contrib/meson/meson/lib/meson.build
Cyan4973: Note that I'm not familiar with `meson`.
Consequently, I'm unaware of the consequences of employing `compile_args`.
From a `Makefile` perspective, something like `-DLZ4_MEMORY_USAGE=#` is supposed to belong to the `CPPFLAGS` variable, rather than the `CFLAGS` one. The difference can be substantial since `CFLAGS` are ignored at link stage, but `CPPFLAGS` are not. It's unclear if this level of differentiation exists at `meson` level.

Then, if, as mentioned by @eli-schwartz, the flags are then also present in `pkgconfig`, well this is unusual, but it seems it would allow the library to be compiled with a different hash table size than the default size (directly defined in the source code). This is pretty advanced stuff, that is rarely ever used. Furthermore, in most cases, this size is abstracted away and does not matter. So it takes a very advanced scenario for this case to matter. And such advanced scenario may more likely happen in the embedded world, where a `pkgconfig` is less likely to be a thing.
So, to summarize, yes, in some rare cases, it could prove useful. But as importantly, in all cases I can think of, and presuming that the `pkgconfig` remains in sync with the installed library, it should not be detrimental.

contrib/meson/meson/programs/meson.build
Cyan4973: I don't remember the details, but I know that the CLI executable is designed to be linked to the _static_ variant of the library, meaning it is not restricted to "public" symbols. One such example is the `xxhash` library, which is not exposed on the public side of `liblz4`, but is employed internally to calculate checksums. The same function is used either as a random number generator or as a checksum for different purposes within the CLI.

eli-schwartz: The most elegant solution here, IMHO, is to do a bit of project restucturing until you end up with:
- `libcommon.a`: a static library containing internal symbols shared between the library and the CLI
- `liblz4.so`/`liblz4.a`: a shared/static library containing public symbols of the library only. It links to libcommon
- `lz4`: a CLI executable that links to both libcommon and to liblz4

So, the CLI executable would be free to link to the shared variant of the library, and it ***is*** restricted to public symbols, but any time you decide you need a private symbol, you just move the private symbol to libcommon. This also has a neat side effect: you get to see which functions are broken out into this internal library, and which presumably represent weak points in the `liblz4.so` API that don't fully meet the needs of your own CLI. Who knows -- those might turn out to be things that merit polishing and stabilization.

The CLI executable then prefers to link to public symbols from liblz4, which depending on the vendor's choice may be static or shared, because some vendors have decided for some reason they like shared libraries ;)

But, for any non-public symbols, the CLI executable just falls back to statically linking them from the convenience library `libcommon`.

You get the best of both worlds. The downside is it requires more work.

contrib/meson/meson/lib/meson.build
tristan957: `compile_args` tells a project using this lz4 dep to compile with the listed `compile_args` in the compiler invocation. Based on what you said this isn't expected nor popular. Therefore I will remove it!

contrib/meson/meson/programs/meson.build
tristan957: I want to copy essentially what upstream is currently doing. So for me I see two options other than the re-org suggested by @eli-schwartz. Link against `liblz4_internal_dep` which is an internal static library or continue to do what I am doing currently which is just extracting objects from the static library.

contrib/meson/meson/meson.build
tristan957: I get the following error:

```
../meson/lib/meson.build:39:0: ERROR: Tried to form an absolute path to a source dir.
You should not do that but use relative paths instead.
```

contrib/meson/meson/meson.build
tristan957: Covered by moving to warning_level=3

contrib/meson/meson/meson.build
tristan957: For now I just bumped to minimum Meson version to 0.53 as mentioned.

eli-schwartz: (I didn't test it, sorry.)

Oh right... it should be an absolute path in both cases.

The difference is that, given rootdir: `/tmp/lz4/contrib/meson`:
- in the `fs.parent(rootdir)` case it is: `/tmp/lz4/lib`
- in the `currentdir / '../../../'` case it is: `/tmp/lz4/contrib/meson/meson/../../../lib`

When integrating lz4 into the wrapdb, we move the meson.build into the top level of the lz4 source tree so `/tmp/lz4` will be the rootdir and both methods will spawn an error.

Sorry, but you cannot simplify this. Please continue to hardcode a set number of `../../../` in the header of each file. It is necessary as long as the Meson files are in `contrib/meson/` instead of in `.` and there is no getting around it (well, other than putting them back where Meson expects them to be).

tristan957:
Alright hopefully that is the last push.

eli-schwartz (approved):
I think this answers all the feedback I had.

tristan957:
@Cyan4973 anything else you need from me on this PR?

Cyan4973:
It would be great if the `meson` CI test was successful (green) after this PR
https://github.com/lz4/lz4/runs/5128636505?check_suite_focus=true

tristan957:
Whoops. My apologies. Will look into it.

tristan957:
I just pushed a fix and have raised an internal issue with Meson, but that shouldn't affect lz4 in any way

tristan957:
Cool the CI is now all good to go as far as this PR is concerned

tristan957:
@vtorri isn't `pic` controlled by `b_staticpic` from Meson? Don't see anything that needs changes on this side of things. Don't see why that would need to be the default, but I am also not familiar with `pic` to begin with.

tristan957:
@Cyan4973 any more feedback for this PR? I got @eli-schwartz to approve since he is a maintainer of Meson and it's wrapdb, so he would be fairly authoritative on this matter.

Cyan4973 (approved):
This looks good to me

Cyan4973:
Thanks for the `meson` script's improvement @tristan957 !

tristan957:
You are welcome. I'll definitely be giving this a go as soon as we get a new release on wrapdb. I may come back with more feedback, but I think this is a much better base than what we had before.

tristan957:
So @eli-schwartz informed me that I can't see the results of my labor easily unless there is a 1.9.4 release of lz4. Would this be possible? Otherwise I can wait until the next release that fits within your schedule.

Cyan4973:
I don't have any availability to work on a new release of `lz4` before Q2, at best.

tristan957:
That's ok. No big deal. Look forward to the next release

tristan957:
Hey @Cyan4973. Wondering if there was anything on your roadmap for a release including this work any time soon. I know Q2 isn't over just yet though

Cyan4973:
There is no release planned _soon_.
I'll need a few days of focus to get to it.
Maybe during the coming Summer.

PR #1066, opened 2022-02-10
when building the library, add gnu visibility and pic flags to meson

vtorri:


eli-schwartz:
#1064 already does visibility, which requires more work to get the programs linking.

Issue #1065, opened 2022-02-08
LZ4 1.9.3 compilation failures on OS400 using ILE CRTCMOD


jonrumsey:
Compiling LZ4 in Native ILE on OS400 results in compilation errors on the static asserts regarding the size of LZ4_STREAMDECODESIZE, LZ4_STREAMSIZE and LZ4_STREAMHCSIZE being large enough to cater for 16 byte pointers and the compiler adding implicit padding for field alignment. 

Whilst solving the first issue seems relatively straightforward by reorganizing fields in a structure to avoid implicit padding, I am hesitant to suggest how to solve the remaining two issues as the calculation of the values does not seem immediately obvious and comments suggest that there is a potential version to version compatibility impact of changing those values.

**1) lz4/lz4-1.9.3/lib/lz4.c line 1426**
`LZ4_STATIC_ASSERT(LZ4_STREAMDECODESIZE >= sizeof(LZ4_streamDecode_t_internal));`

LZ4_STREAMDECODESIZE is set to the size of an array of unsigned long long (8 bytes), on OS400 this array has 6 elements (6 x 8 = 48 bytes)

LZ4_streamDecode_t_internal is defined as follows;

```
    const LZ4_byte* externalDict;
    size_t extDictSize;
    const LZ4_byte* prefixEnd;
    size_t prefixSize;
```
However to ensure the pointers start on a 16 byte boundary and to ensure the size of the structure is also aligned, implicit padding is added by the compiler after the size_t fields. So the padded size of this structure on OS400 is actually 64 bytes. Provided the order of the fields in the structure is unimportant, reordering the fields like this avoids implicit padding and keeps this structure to 48 bytes as calculated;

```
    const LZ4_byte* externalDict;
    const LZ4_byte* prefixEnd;
    size_t extDictSize;
    size_t prefixSize;
```
**2) lz4/lz4-1.9.3/lib/lz4.c line 2268**
A further compile failure on the following;

`LZ4_STATIC_ASSERT(LZ4_STREAMSIZE >= sizeof(LZ4_stream_t_internal));    /* A compilation error here means LZ4_STREAMSIZE is not large enough */`

On OS/400, LZ4_STREAMSIZE is (1 << 14 + 32) 16416 but sizeof(LZ4_stream_t_internal) is 16488 bytes, again it contains implicit padding to ensure pointer fields are aligned ;

```
|LZ4_stream_t_internal                                    
|=========================================================
|      Offset       |      Length       | Member Name     
[...]

PR #1070, opened 2022-03-12
Remove assumptions that pointers are no larger than 64-bit

jonrumsey:
Re-order fields in LZ4_stream_t_internal and LZ4_streamDecode_t_internal to minimize implicit padding on OS400 where pointers are aligned on 16 byte boundaries.

Adjust calculations of LZ4_STREAMSIZE and LZ4_STREAMHCSIZE to allow for pointers larger than 64-bit.

This addresses issue https://github.com/lz4/lz4/issues/1065

### DIFF
--- a/lib/lz4.h
+++ b/lib/lz4.h
@@ -608,15 +608,16 @@ struct LZ4_stream_t_internal {
     LZ4_u32 hashTable[LZ4_HASH_SIZE_U32];
     LZ4_u32 currentOffset;
     LZ4_u32 tableType;
+    LZ4_u32 dictSize;
+    /* Implicit padding to ensure pointer is aligned */
     const LZ4_byte* dictionary;
     const LZ4_stream_t_internal* dictCtx;
-    LZ4_u32 dictSize;
 };
 
 typedef struct {
     const LZ4_byte* externalDict;
-    size_t extDictSize;
     const LZ4_byte* prefixEnd;
+    size_t extDictSize;
     size_t prefixSize;
 } LZ4_streamDecode_t_internal;
 
@@ -631,7 +632,7 @@ typedef struct {
  *  note : only use this definition in association with static linking !
  *  this definition is not API/ABI safe, and may change in future versions.
  */
-#define LZ4_STREAMSIZE       ((1UL << LZ4_MEMORY_USAGE) + 32)  /* static size, for inter-version compatibility */
+#define LZ4_STREAMSIZE       ((1UL << LZ4_MEMORY_USAGE) + sizeof(void*) + sizeof(void*) + 16)  /* static size, for inter-version compatibility */
 #define LZ4_STREAMSIZE_VOIDP (LZ4_STREAMSIZE / sizeof(void*))
 union LZ4_stream_u {
     void* table[LZ4_STREAMSIZE_VOIDP];
--- a/lib/lz4hc.h
+++ b/lib/lz4hc.h
@@ -220,7 +220,7 @@ struct LZ4HC_CCtx_internal
 /* Do not use these definitions directly !
  * Declare or allocate an LZ4_streamHC_t instead.
  */
-#define LZ4_STREAMHCSIZE       262200  /* static size, for inter-version compatibility */
+#define LZ4_STREAMHCSIZE       (262176 + sizeof(void*) + sizeof(void*) + sizeof(void*))  /* static size, for inter-version compatibility */
 #define LZ4_STREAMHCSIZE_VOIDP (LZ4_STREAMHCSIZE / sizeof(void*))
 union LZ4_streamHC_u {
     void* table[LZ4_STREAMHCSIZE_VOIDP];


Cyan4973:
Thanks for the suggested improvement.

I've no doubt that it works as intended assuming correct version alignment between the user program and the library.

What I'm more concerned of is
what happens if there is a version misalignment ?

The most common issue is an older library version linked to a program expecting a newer one.
I would expect the program to plan for this in advance, but it's still a concern to properly underline and document.

Another possible scenario is a new library version linked to a program expecting an older library.
In which case, it's expected that it's up to the library to ensure backward compatibility.

Since this PR impacts `LZ4_stream_t`'s size, I have to wonder what happens in case of version mismatch scenarios.
I don't think these scenarios are tested in CI, so we have to assume that CI is (currently) blind to these topics.

t-mat:
Since I don't have enough free time to see this issue, I just put the following links to help contributors to think about the impact of this change.

<details>
<summary>History of LZ4_stream_t_internal</summary>

https://github.com/lz4/lz4/blame/874f3e095b1d64ab3545a3f2a3f7403a44ebb3bb/lib/lz4.h#L128-L135
```C
typedef struct {
    uint32_t hashTable[LZ4_HASH_SIZE_U32];
    uint32_t currentOffset;
    uint32_t initCheck;
    const uint8_t* dictionary;
    uint8_t* bufferStart;   /* obsolete, used for slideInputBuffer */
    uint32_t dictSize;
} LZ4_stream_t_internal;
```


https://github.com/lz4/lz4/blame/62cb52b3410bdfd696669910de010ab03666edc8/lib/lz4.h#L368-L376
```C
typedef struct {
    uint32_t hashTable[LZ4_HASH_SIZE_U32];
    uint32_t currentOffset;
    uint16_t initCheck;
    uint16_t tableType;
    const uint8_t* dictionary;
    uint8_t* bufferStart;   /* obsolete, used for slideInputBuffer */
    uint32_t dictSize;
} LZ4_stream_t_internal;
```


https://github.com/lz4/lz4/blame/6933f5ad9ca57caa8a76103f1382560fba0f9691/lib/lz4.h#L360-L367
```C
typedef struct {
    uint32_t hashTable[LZ4_HASH_SIZE_U32];
    uint32_t currentOffset;
    uint32_t initCheck;
    const uint8_t* dictionary;
    uint8_t* bufferStart;   /* obsolete, used for slideInputBuffer */
    uint32_t dictSize;
} LZ4_stream_t_internal;
```


https://github.com/lz4/lz4/blame/d7399232a4d547d7183c193997e17d534d000f52/lib/lz4.h#L564-L572
```C
struct LZ4_stream_t_internal {
    uint32_t hashTable[LZ4_HASH_SIZE_U32];
    uint32_t currentOffset;
    uint16_t dirty;
    uint16_t tableType;
    const uint8_t* dictionary;
    const LZ4_stream_t_internal* dictCtx;
    uint32_t dictSize;
};
```


https://github.com/lz4/lz4/blame/78433070abd9da36f225a5be1c302e42cdad67bd/lib/lz4.h#L586-L593
```C
struct LZ4_stream_t_internal {
    uint32_t hashTable[LZ4_HASH_SIZE_U32];
    uint32_t currentOffset;
    uint32_t tableType;
    const uint8_t* dictionary;
    const LZ4_stream_t_internal* dictCtx;
    uint32_t dictSize;
};
```


This PR
https://github.com/lz4/lz4/blame/845f75f649c2d398e92baa0ee0517b6e4653e879/lib/lz4.h#L607-L615
```C
struct LZ4_stream_t_internal {
    LZ4_u32 hashTable[LZ4_HASH_SIZE_U32];
    LZ4_u32 currentOffset;
    LZ4_u32 tableType;
    LZ4_u32 dictSize;
    /* Implicit padding to ensure pointer is aligned */
    const LZ4_byte* dictionary;
    const LZ4_stream_t_internal* dictCtx;
};
```
</details>

jonrumsey:
Thanks for your comments, I think I understand the point that you make that the order of the fields may be significant when relinking a program to a newer version of the library. I had proposed this change on the assumption that whilst the size of the structure is significant (hence the need for the LZ4_STATIC_ASSERTS) that the '_internal' suffix and this comment block;

```
  * Do not use these definitions directly.
  * They are only exposed to allow static allocation of 'LZ4_stream_t' and 'LZ4_streamDecode_t'.
  * Accessing members will expose user code to API and/or ABI break in future versions of the library.
```
  
seemed to imply that the layout/field offsets is not something that should be relied upon outside the library. 

In terms of a potential change in structure size, all compilers will add padding to the end of the structure as necessary to ensure the whole structure is aligned (as the struct is not declared as packed), I believe that by shuffling the fields that this would keep the size of the structure the same on platforms with 32 and 64-bit pointers, but also avoid the needless padding on OS400.

The structure before the pull request;

```
struct LZ4_stream_t_internal {
    uint32_t hashTable[LZ4_HASH_SIZE_U32];   // 16384 on all platforms
    uint32_t currentOffset;                  // 4 on all platforms
    uint32_t tableType;                      // 4 on all platforms
                                             // 0 on 32 & 64-bit, 8 on OS400 for pointer alignment
    const uint8_t* dictionary;               // 4 on 32-bit, 8 on 64-bit and 16 on OS400
    const LZ4_stream_t_internal* dictCtx;    // 4 on 32-bit, 8 on 64-bit and 16 on OS400
    uint32_t dictSize;                       // 4 on all platforms
                                             // 0 on 32-bit, 4 on 64-bit platforms, 12 on OS400 (structure padding)
};
```

32-bit structure is 16404 (16384 + 4 + 4 + 4 + 4 + 4), 64-bit structure is 16416 (16384 + 4 + 4 + 8 + 8 + 4 + 4) and OS400 structure is 16448 (16384 + 4 + 4 + 8 + 16 + 16 + 4 + 12). 

The structure proposed in the pull request;

```
struct LZ4_stream_t_internal {
    LZ4_u32 hashTable[LZ4_HASH_SIZE_U32];    // 16384 on all platforms
    LZ4_u32 currentOffset;                   // 4 on all platforms
    LZ4_u32 tableType;                       // 4 on all platforms
    LZ4_u32 dictSize;                        // 4 on all platforms
    /* Implicit padding */                   // 0 on 32-bit, 4 on both 64-bit and OS400
    const LZ4_byte* dictionary;              // 4 on 32-bit, 8 on 64-bit and 16 on OS400 
    const LZ4_stream_t_internal* dictCtx;    // 4 on 32-bit, 8 on 64-bit and 16 on OS400
                                             // 0 bytes on all platforms, final pointer ensures structure aligned
};
```

32-bit structure is 16404 (16384 + 4 + 4 + 4 + 4 + 4), 64-bit structure is 16416 (16384 + 4 + 4 + 4 + 4 + 8 + 8) and OS400 structure is 16432 (16384 + 4 + 4 + 4 + 4 + 16 + 16).

The only platform that the structure changes size is on OS400 (smaller by 16 bytes) and that was intentional to eliminate the unnecessary padding and to keep the calculation of LZ4_STREAMSIZE based on the platform pointer size to resolve the compilation errors.

Perhaps the safer change to make would be to leave structures as-is and for the calculation of LZ4_STREAMDECODESIZE, LZ4_STREAMSIZE and LZ4_STREAMHCSIZE to factor in the implicit padding added on platforms where pointers are larger than 64-bit. I don't think that would be as efficient, but maybe simpler to maintain in step with any future structure changes, so I'll close this pull request and rework on that basis.

Cyan4973:
I believe that only the size of the structure matters, not the order of the fields.
Or, said differently, the order of the fields may only matter through its potential to change the size of the `struct` (by introducing additional / less padding bytes).

You might well be right that your modification doesn't impact 32-bit and 64-bit systems.
And if it only impacts OS400, then it's fine with me, since it's the intention of this PR.

My original question was rather about "has it been tested / checked".

Cyan4973:
Any plan to submit a new PR on this topic ?

jonrumsey:
Sorry I have not had an opportunity to work on this (to leave structures unchanged and change the macros to take into account padding). I will try and make some time to look at doing this over the next couple of days.

PR #1079, opened 2022-05-01
Fetch trailing bytes in a branchless way during match length encoding

danlark1:
Disclaimer: I want maintainers to verify the speedup and generally open to comments if this direction even makes sense

The idea is to fetch last 5 bytes during LZ4_count given we are allowed to read 4 bytes beyond buffer. These 5 bytes can be used for testing next positions without any pointer load and thus improve the dependency chain.

The idea is taken from google/snappy. I was wondering if this gives a better codegen for lz4 as well and it turns out that yes. For all comments on all assembly, etc, see this https://github.com/google/snappy/blob/8dd58a519f79f0742d4c68fbccb2aed2ddb651e8/snappy-internal.h#L191

Preliminary benchmarks:

clang: 5-10%

Silesia.tar

```
BM_ZFlatLZ4/silesia.tar                  -0.0447         -0.0446     355984938     340079628     355931342     340051698
```

gcc: 0.5-1%

```
BM_ZFlatLZ4/silesia.tar                  -0.0049         -0.0048     327836186     326238017     327743015     326154035
```

Other benchmarks TBD but generally we see improvements in cases when there are lots of matches. For example, xml data set from silesia gives more improvements even for gcc

clang:
```
BM_ZFlatLZ4/xml_mean                -0.0466         -0.0466       6292872       5999359       6292167       5999121
```

gcc:

```
BM_ZFlatLZ4/xml_mean                -0.0325         -0.0321       5840298       5650403       5837447       5650165
```

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -939,8 +939,8 @@ LZ4_FORCE_INLINE int LZ4_compress_generic_validated(
         outputDirective == notLimited && /* Cannot be relaxed, ip changes between match and table update. */
         dictDirective == noDict &&       /* Cannot be relaxed, ip changes between match and table update. */
         dictIssue == noDictIssue &&      /* Cannot be relaxed, ip changes between match and table update. */
-        LZ4_isLittleEndian() &&          /* Can be relaxed. */
-        sizeof(reg_t) == 8               /* Can be relaxed. */
+        LZ4_isLittleEndian() &&          /* Can be relaxed for 64 bit BIG ENDIAN. */
+        sizeof(reg_t) == 8               /* Can be relaxed but not all hacks can be applied. */
         ;
 
 
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -933,15 +933,17 @@ LZ4_FORCE_INLINE int LZ4_compress_generic_validated(
                  const dictIssue_directive dictIssue,
                  const int acceleration)
 {
-    // During the match encoding, we can fetch in a branchless way last 5 bytes
-    // without any dependency chain on x86 and aarch which results in faster
-    // compression speeds for highly compressed data (with lots of matches).
+    /* During the match encoding, we can fetch in a branchless way last 5 bytes
+       without any dependency chain on x86 and aarch which results in faster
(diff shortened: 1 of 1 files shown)


Cyan4973:
If I understand correctly, 
the whole point of this complexity is to avoid reloading data from input pointer `ip` after a match
on the ground that this data has already been fetched during match counting.
Though this is not a "free" operation, it requires actively loading more data at end of match counting,
but it nonetheless results in better performance, presumably because loading unconditionally at position `p+4` reduces the dependency chain, compared to loading position `ip + ml`.

This sounds like a cool trick,
though I'm quite concerned by the complexity / performance trade off involved, from a maintenance perspective.

I guess for the time being you are mostly concerned by the first part of the equation, performance benefits.

danlark1:
> If I understand correctly, the whole point of this complexity is to avoid reloading data from input pointer `ip` after a match on the ground that this data has already been fetched during match counting. Though this is not a "free" operation, it requires actively loading more data at end of match counting, but it nonetheless results in better performance, presumably because loading unconditionally at position `p+4` reduces the dependency chain, compared to loading position `ip + ml`.

Exactly, yes. It works in combination with the fact it's safe to read 8 bytes from `ip+4` and that hash is computed by at most 5 bytes (`LZ4_hash5`).

> This sounds like a cool trick, though I'm quite concerned by the complexity / performance trade off involved, from a maintenance perspective.

It's fine, totally agree. If we are not able to get enough performance benefits out of it, we would just apply it internally. Other request is to think whether we can get even more out of that trick or simplify. Or maybe restructure the compression algorithm a little.

> I guess for the time being you are mostly concerned by the first part of the equation, performance benefits.

Yes

Cyan4973:
> whether we can get even more out of that trick or simplify. Or maybe restructure the compression algorithm a little.

Sounds like promising directions.
The underlying loading mechanism might prove more generalizable.
And of course, simplifying the code would play in favor of its acceptance in mainline.

danlark1:
I decided it will be hard to push to a consistent production state but there is definitely an opportunity

PR #1089, opened 2022-06-01
Add CMakeLists.txt

legleux:
`CMakeLists.txt` from 1.9.2 seems to work fine, include it so CMake can be used to build lz4.


Cyan4973:
What's the problem with `build/cmake` ?

legleux:
Hmm, the problem was, I always use `build` as the `CMAKE_BINARY_DIR` so when I checked out `1.9.2`, then `1.9.3` I didn't notice it.
Sorry!

Issue #1051, opened 2022-01-14
Support partial decompression with dictionary 


hsiangkao:
Hi,
(Sorry, sadly I didn't seek much time working on modifying lz4, so just try to record a feature request first...)

Our next step of EROFS on lz4 is to support compression with an external dictionary.
Test results show that it can have benefits on some files with small cluster sizes (which have small
I/O amplication) and specific patterns.

The compressor side looks good. However, it seems that a partial decompression API like
LZ4_decompress_safe_partial_usingDict() is missing...

It would be much helpful to get this so that we don't need to take more unnecessary page cache...

Thanks,
Gao Xiang

hsiangkao:
many thanks, it has been resolved! ;)

PR #1093, opened 2022-06-07
API: add `LZ4_decompress_safe_partial_usingDict` to support partial decompression with dict

yawqi:
Add `LZ4_decompress_safe_partial_usingDict` API to support partial decompression with dictionary.

This patch is still working in progress, I will appreciate it if you guys could give me some suggestions. Thanks for your review!

Feature request #1051.

### DIFF
--- a/lib/lz4.h
+++ b/lib/lz4.h
@@ -430,6 +430,7 @@ LZ4LIB_API int LZ4_decompress_safe_continue (LZ4_streamDecode_t* LZ4_streamDecod
  */
 LZ4LIB_API int LZ4_decompress_safe_usingDict (const char* src, char* dst, int srcSize, int dstCapcity, const char* dictStart, int dictSize);
 
+LZ4LIB_API int LZ4_decompress_safe_partial_usingDict(const char* source, char* dest, int compressedSize, int targetOutputSize, int maxOutputSize, const char* dictStart, int dictSize);
 #endif /* LZ4_H_2983827168210 */
 
 
--- a/ossfuzz/round_trip_fuzzer.c
+++ b/ossfuzz/round_trip_fuzzer.c
@@ -20,8 +20,13 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
 
     size_t const partialCapacity = FUZZ_getRange_from_uint32(partialCapacitySeed, 0, size);
     size_t const dstCapacity = LZ4_compressBound(size);
-
-    char* const dst = (char*)malloc(dstCapacity);
+    size_t const largeSize = 64 * 1024 - 1;
+    size_t const smallSize = 1024;
+    char* const dstPlusLargePrefix = (char*)malloc(dstCapacity + largeSize);
+    char* const dstPlusSmallPrefix = dstPlusLargePrefix + largeSize - smallSize;
+    char* const largeDict = (char*)malloc(largeSize);
+    char* const smallDict = largeDict + largeSize - smallSize;
+    char* const dst = dstPlusLargePrefix + largeSize;
     char* const rt = (char*)malloc(size);
 
     FUZZ_ASSERT(dst);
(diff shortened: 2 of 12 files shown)


lib/lz4.h
```diff
@@ -777,6 +778,8 @@ LZ4LIB_API int LZ4_decompress_fast_usingDict (const char* src, char* dst, int or
 LZ4LIB_API void LZ4_resetStream (LZ4_stream_t* streamPtr);
 
 
+/*! Obsolete partial decompress with dict functions */
+LZ4_DEPRECATED("use LZ4_decompress_safe_partial_usingDict() instead") LZ4LIB_API int LZ4_decompress_safe_partial_withPrefix64k(const char* source, char* dest, int compressedSize, int targetOutputSize, int dstCapacity);
```
Cyan4973: It's unusual for a newly introduced function to be immediately classified "deprecated".
Is this entry added for the sake of consistency with `LZ4_decompress_safe_withPrefix64k()` ?

ossfuzz/round_trip_fuzzer.c
```diff
@@ -47,7 +52,61 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
         FUZZ_ASSERT_MSG(!memcmp(data, partial, partialSize), "Corruption!");
         free(partial);
     }
-
+    /* Partial decompression using dict with no dict. */
```
Cyan4973, line 55: That's pretty nice,
good testing is a must for a new feature.

Cyan4973:
The implementation looks fine to me.
I believe it "just" re-employs internal functions within a scope which is "more complete".
So that limits amount of changes, hence of risks, in the engine.

Moreover, it comes with a great battery of tests, which is a welcome relief for reliability concerns.

I'm a little concerned by the length of the selected name `LZ4_decompress_safe_partial_usingDict()`, which sounds like a mouthful, but have no better counter-proposal to suggest. Plus, it respects the existing naming convention, so it doesn't feel "out of place".

lib/lz4.h
yawqi: Thanks for your review! Yes, should I remove it?

Cyan4973:
OK, I'll go ahead and say that it's almost completely fine, on first attempt. So that's great job. Even the name is green-lighted, because it's a direct reference to existing `LZ4_decompress_safe_partial`.

My only comment is about `LZ4_decompress_safe_partial_withPrefix64k()`.
I don't think it's worthwhile to publish this entry and make it immediately "deprecated".
The reason we have deprecated entries in `lz4.h` is because they used to be part of the official API, therefore some existing programs might still be depending on them, and we don't want to break their linkage just for a version update.
However, for this specific instance, there is no prior usage of this entry point since it never existed.
So no need to publish it at all.
This is one less entry point to maintain, and also one less potential vulnerability vector to worry about in the future.

lib/lz4.h
Cyan4973: Yes please

yawqi:
> OK, I'll go ahead and say that it's almost completely fine, on first attempt. So that's great job. Even the name is green-lighted, because it's a direct reference to existing `LZ4_decompress_safe_partial`.
> 
> My only comment is about `LZ4_decompress_safe_partial_withPrefix64k()`. I don't think it's worthwhile to publish this entry and make it immediately "deprecated". The reason we have deprecated entries in `lz4.h` is because they used to be part of the official API, therefore some existing programs might still be depending on them, and we don't want to break their linkage just for a version update. However, for this specific instance, there is no prior usage of this entry point since it never existed. So no need to publish it at all. This is one less entry point to maintain, and also one less potential vulnerability vector to worry about in the future.

Thanks for your review! I will remove the `LZ4_decompress_safe_partial_withPrefix64k` entry.

lib/lz4.h
yawqi: Thanks! Done.

Cyan4973 (approved):

hsiangkao:
Hi Cyan, could we merge it so I can develop more based on this? Many thanks!

hsiangkao:
Thanks!

Issue #1086, opened 2022-05-26
lz4 1.9.3 cli tool is not able to compress block devices


Acu43:
lz4 cli tool is not able to compress block devices after this [commit](https://github.com/lz4/lz4/commit/729eef61a12d381f436da180329dba74b9892aa4).
Block devices are not treated as regular files so lz4 exits with an error code as "is not a regular file:

$ ls -ld /dev/dm-2 
brw-rw---- 1 root disk 254, 2 May 26 12:48 /dev/dm-2
$ lz4 /dev/dm-2 
/dev/dm-2: is not a regular file 
$
$ lz4 --version
*** LZ4 command line interface 64-bits v1.9.3, by Yann Collet ***

Correct behaviour for 1.8.3:
$ lz4 /dev/dm-2 
Compressed filename will be : /dev/dm-2.lz4 
Compressed 104857600 bytes into 433460 bytes ==> 0.41%                         
$ lz4 --version
*** LZ4 command line interface 64-bits v1.8.3, by Yann Collet ***

1. Create an lvm volume or a physical partition.
2. lz4 the-path-of-the-lvm-volume
3. Observer that it fails

**System (please complete the following information):**
 - OS: Debian bullseye
 - Version: 1.9.3

Cyan4973:
If I understand #704 correctly, which is the patch causing this issue,
the only intention was for `lz4 -t non-existing-file` to fail, as it should.

But the solution proposed is overly specific, only works for single-file tests, 
and implement the check at an incorrect place in the code base (it should be `lz4io.c`).

Removing the code of #704,
I have the pleasure to notice that the original test case `lz4 -t non-existing-file` still correctly fails, as it should.
So there is no need for this extra code, which is the one producing the side effect on block devices reported in this issue.

I presume the underlying condition, which was a lack of reporting the named file as non-present, has been fixed since.

Cyan4973:
Fixed in latest `dev` branch update

PR #1095, opened 2022-06-21
enable block device compression #1086

lixiaofei38:


Cyan4973:
What kind of test could be added to ensure the PR works as intended (and the property remain respected in the future) ?

Cyan4973:
I'll merge this PR into a feature branch which implements a Block device test,
so that we can observe this PR fixing the test,
and can continue checking this condition in the future.

PR #1098, opened 2022-07-01
Block device compression

### LINKED ISSUE #1086: lz4 1.9.3 cli tool is not able to compress block devices

Cyan4973:
Fix #1086 
by removing code from #704 
which seems no longer necessary to fix the problem it initially targeted (`-t not-existing-filename`)
but introduced the side effect reported in #1086 .

Note : an earlier version of this fix was based on #1095 ,
it was rolled back after discovering that it doesn't work.

Also : used this opportunity to also fix `-tm` (test mode on multiple lz4-compressed files)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -250,7 +250,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2 # https://github.com/actions/checkout
-
     - name: custom LZ4_DISTANCE_MAX; test LZ4_USER_MEMORY_FUNCTIONS
       run: |
         MOREFLAGS='-DLZ4_DISTANCE_MAX=8000' make V=1 check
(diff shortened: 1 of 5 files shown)


Cyan4973:
cc @lixiaofei38 : the proposed fix doesn't seem to work ( https://github.com/lz4/lz4/runs/7142938640?check_suite_focus=true )

Cyan4973:
Changed the fix for #1086, which is no longer based on #1095.

PR #1100, opened 2022-07-09
Fix recent fuzzers issues and false positives

yoniko:
- Fixed incorrect free in `round_trip_fuzzer.c` (https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=48884)
- Fixed `round_trip_frame_uncompressed_fuzzer.c` to not use uninitialized memory (https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=48910)
(note file was reformatted, please ignore whitespaces when reviewing)

### DIFF
--- a/ossfuzz/round_trip_frame_uncompressed_fuzzer.c
+++ b/ossfuzz/round_trip_frame_uncompressed_fuzzer.c
@@ -16,123 +16,119 @@
 #include "lz4frame_static.h"
 
 static void decompress(LZ4F_dctx *dctx, void *src, void *dst,
-                           size_t dstCapacity, size_t readSize) {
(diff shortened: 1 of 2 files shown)


Cyan4973 (approved):

Cyan4973:
Thanks for the fixes @yoniko ! 
They look good to me.

I'm especially surprised by the incorrect `free()` fixed in `round_trip_fuzzer.c`. This is a blatant issue, it should have been caught when produced. 

It looks like CI tests have some serious shortcomings when it comes to validating oss-fuzz related code.

yoniko:
Edited commit messages & descriptions which have been wrong

PR #1120, opened 2022-07-17
`make test` can run in `-j` parallel mode

### LINKED ISSUE #957: Parallel test does not work
**Describe the bug**

`make check -j` fails:
image

Likely to be a race condition for filename.

**Expected behavior**

Support parallelism as other target.

 - OS: [e.g. Mac] CentOS 7
 - Version [e.g. 22] v1.9.3
 - Compiler [e.g. gcc] gcc-9 (devtoolset-9)
 - Build System [e.g. Makefile] makefile
 - Other hardware specs [e.g Core 2 duo...] Xeon Gold 6140

Cyan4973:
fix #957

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -179,7 +179,7 @@ list:
 check: test-lz4-essentials
 
 .PHONY: test
-test: test-lz4 test-lz4c test-frametest test-fullbench test-fuzzer test-install test-amalgamation listTest test-decompress-partial
+test: test-lz4 test-lz4c test-frametest test-fullbench test-fuzzer test-amalgamation listTest test-decompress-partial
 
 .PHONY: test32
 test32: CFLAGS+=-m32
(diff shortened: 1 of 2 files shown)


tests/Makefile
```diff
@@ -201,279 +201,300 @@ test-compile-with-lz4-memory-usage:
 	$(MAKE) clean; CFLAGS=-O0 CPPFLAGS=-D'LZ4_MEMORY_USAGE=LZ4_MEMORY_USAGE_MAX' $(MAKE) all
 
 .PHONY: test-lz4-sparse
+# Rules regarding Temporary test files :
+# Each test must use its own unique set of names during execution.
+# Each temporary test file must begin by an FPREFIX.
+# Each FPREFIX must be unique for each test.
```
t-mat, line 207: This is a confirmation for minor/corner case.
As for uniqueness, we also do not accept common prefix of `FPREFIX`.

For example

```
# NG
test-1: FPREFIX = tmp-test-1
	...
	@$(RM) $(FPREFIX)*  # This command (`rm tmp-test-1*`) also removes `tmp-test-1-1*`
...
test-1-1: FPREFIX = tmp-test-1-1
	...
	@$(RM) $(FPREFIX)*
...
```

t-mat:
I've compared durations of the following tests.

- #1120 (This PR) : https://github.com/lz4/lz4/runs/7362381775?check_suite_focus=true
- #1119 (Previous PR) : https://github.com/lz4/lz4/runs/7379894072?check_suite_focus=true

And glad to see a superb improvement for all `make -C tests` variants.  Especially for `-mx32`!

xkszltl:
Good to see the large speed up in `make test`, together with the `-j` support.
Previously we can only afford `make check` in most builds.

Issue #1071, opened 2022-03-21
Implement LZ4_FREESTANDING


Cyan4973:
Suggested by @t-mat  in  : 
https://github.com/lz4/lz4/pull/1068#issuecomment-1060814983

t-mat:
I've written simple sample of `-nostdlib` program for Linux.
https://gist.github.com/t-mat/aa78ff79f31ee79c4a709d4def3ee1d7

I think our next steps are

(1) Think about other platforms.  Do we need similar `-nostdlib` program for macos/win32/BSDs/etc?  How about CI tests?

(2) Which standard C function(s) will be needed by the `LZ4_FREESTANDING` test program?
Perhaps we should write our test program first.  And after that we can think about our needs for standard functions easily.

Cyan4973:
Freestanding, i.e. no dependency on any standard C library function, is a real use case, but it's difficult to clearly articulate, because it's more a collection of niche use cases.

I generally don't see the point of Freestanding in a situation where a fully-fledged OS is available, hence wouldn't consider  `macos`, `win32`, `linux`, `BSD`, etc. as valid targets.
It's more for very limited systems with a layer barely worthy of being called an OS, so it's more the territory of "internet of things", from which I already exclude things like smart watches for example, which are already too complex.

I think `liblz4` has a very narrow surface when it comes to C standard lib dependency. Essentially, `malloc`, `free`, and `memcpy`. Probably `memset` and `calloc`. Maybe `realloc`.
*edit* : nope, `realloc` is only for the CLI, not the library.

So that's the API surface that any freestanding environment must provide for a freestanding `liblz4` to work.

Ah, and I should add: in the context of freestanding environment, I would only expect static linking.

Cyan4973:
Question :
should we try to make this topic an item for `v1.9.4` release ?

t-mat:
I think we can simply skip this `LZ4_FREESTANDING` for new release.  Because we don't see any urgent demand for it.
I believe all project maintainers who may like this feature already managed their situation properly.

But we may need to some remark for `LZ4F_*` and its requirement of `<stdio.h>` though.

t-mat:
I wrote new -nostdlib harness for Linux platforms.
https://gist.github.com/t-mat/dbdc4f71ae52f4eca83fd67f683e7452

Now it #includes the following file and can be compiled with `-nostdlib`.

- lz4.c
- lz4hc.c
- lz4frame.c
- xxhash.c

Compilation:

```
cd
[...]

PR #1124, opened 2022-07-31
Introduce LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION

t-mat:
Part of #1071.

This changeset introduces new compile time switch macro `LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION`
which removes the following functions when it's defined.

```
// lz4.c
LZ4_createStream
LZ4_freeStream
LZ4_createStreamDecode
LZ4_freeStreamDecode
LZ4_create              // legacy

// lz4hc.c
LZ4_createStreamHC(void)
LZ4_freeStreamHC
LZ4_createHC            // legacy
LZ4_freeHC              // legacy
```

These functions uses dynamic memory allocation functions such as malloc() and free().
It'll be useful for freestanding environment which doesn't have these allocation functions.

Since this change breaks API, this macro is only valid with lz4 as a static linked object.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -189,12 +189,9 @@
 *  Memory routines
 **************************************/
 #if defined(LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION)
-#  undef LZ4_ALLOC
-#  undef LZ4_ALLOC_AND_ZERO
-#  undef LZ4_FREEMEM
-#  define LZ4_ALLOC(x)			lz4_error_memory_allocation_is_disabled
-#  define LZ4_ALLOC_AND_ZERO(x)	lz4_error_memory_allocation_is_disabled
-#  define LZ4_FREEMEM(x)		lz4_error_memory_allocation_is_disabled
+#  define ALLOC(s)          lz4_error_memory_allocation_is_disabled
+#  define ALLOC_AND_ZERO(s) lz4_error_memory_allocation_is_disabled
+#  define FREEMEM(p)        lz4_error_memory_allocation_is_disabled
 #elif defined(LZ4_USER_MEMORY_FUNCTIONS)
 /* memory management functions can be customized by user project.
  * Below functions must exist somewhere in the Project
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -188,7 +188,14 @@
 /*-************************************
 *  Memory routines
 **************************************/
-#ifdef LZ4_USER_MEMORY_FUNCTIONS
+#if defined(LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION)
+#  undef LZ4_ALLOC
+#  undef LZ4_ALLOC_AND_ZERO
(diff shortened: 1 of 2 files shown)


lib/lz4.c
```diff
@@ -188,7 +188,14 @@
 /*-************************************
 *  Memory routines
 **************************************/
-#ifdef LZ4_USER_MEMORY_FUNCTIONS
+#if defined(LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION)
+#  undef LZ4_ALLOC
+#  undef LZ4_ALLOC_AND_ZERO
+#  undef LZ4_FREEMEM
+#  define LZ4_ALLOC(x)			lz4_error_memory_allocation_is_disabled
```
Cyan4973: If I understand properly, these macros `LZ4_ALLOC(x)` and next are designed to fail at compile time, as a way to detect if they are employed anywhere.

However, these macros are also not used anywhere.

Did you mean to replace `ALLOC(s)` ?

t-mat: oops.  You're right.  They must be `ALLOC(s)`, `ALLOC_AND_ZERO(s)` and `FREEMEM(p)`.

lib/lz4.c
```diff
@@ -188,7 +188,11 @@
 /*-************************************
 *  Memory routines
 **************************************/
-#ifdef LZ4_USER_MEMORY_FUNCTIONS
+#if defined(LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION)
```
Cyan4973, line 191: How will users be made aware of this new capability ?
Maybe it should be documented ?

t-mat: Yes!  It needs document.  Draft version:

```C
/*!
 * LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION :
 * Disable relatively high-level LZ4/HC functions which uses dynamic memory
 * allocation (malloc(), calloc(), free()).
 *
 * Note that this is a compile-time switch.  And since it disables public/stable
 * LZ4 v1 API functions, we don't recommend to use this symbol to generate
 * library for distribution.
 *
 * The following functions are removed when this symbol is defined.
 *
 * // lz4.c
 * LZ4_createStream
 * LZ4_freeStream
 * LZ4_createStreamDecode
 * LZ4_freeStreamDecode
 * LZ4_create              // legacy
 * 
 * // lz4hc.c
 * LZ4_createStreamHC(void)
 * LZ4_freeStreamHC
 * LZ4_createHC            // legacy
 * LZ4_freeHC              // legacy
 */
```

I also realize I should add `#ifndef` to declaration of these functions in header.

Cyan4973: Do you expect to provide this documentation as part of this PR ?
Or do you want to make it a separate PR ?

t-mat: I'd like to send it as a separate PR.  Sorry!

PR #1126, opened 2022-08-01
refactor read_variable_length()

Cyan4973:
- improved documentation
- more `assert()`
- better distinction between "end of scanning area" and "end of buffer"
- minor performance optimization, by separating first branch from the rest of the loop.

The last item is more controversial, because it's difficult to measure the benefit. While it does result in measurable performance improvement on my bench platform, the main lesson is that the LZ4 decoding loop works so close to the limits of cpu front end that performance is primarily driven by the  DSB / MITE repartition of instructions, which is itself depending on complex instruction alignments consequences, which are essentially random from a C code perspective (it's up to the compiler to decide that, and unfortunately, it's not doing a consistent job). So minor code changes can result in vast performance differences, in either direction, for any minor detail (exact compiler version, set of flags, surrounding code), making it hard to determine if a given change is globally positive or not.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1732,19 +1732,19 @@ typedef enum { decode_full_block = 0, partial_decode = 1 } earlyEnd_directive;
  * @initial_check - check ip >= ipmax before start of loop.  Returns initial_error if so.
  * @error (output) - error code.  Must be set to 0 before call.
  */
+typedef unsigned Rvl_t;  /* note : using size_t seems to result in slowdown */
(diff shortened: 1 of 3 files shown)


Cyan4973:
Closing this.
I will have to settle for something less ambitious.

I realize it has been a long time since I last touched the decoder, and it has evolved a lot. There are some long distances relations between different parts of the code that I don't like much. A larger refactor would probably help to ensure all parts of the decoder are fully under control.

PR #1130, opened 2022-08-07
Add LZ4_FREESTANDING

t-mat:
#1071 

This PR adds `LZ4_FREESTANDING` and its tiny document and test code.

Notes about the test code:

- It tests basic lz4.c and lz4hc.c de/compression functions.
- It verifies that these functions are able to be compiled and executed without standard library (freestanding environment).
- It doesn't do benchmark.

I'll send another PR which adds `test-freestanding` to GitHub Actions.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -56,7 +56,7 @@ NB_LOOPS     ?= -i1
 .PHONY: default
 default: all
 
-all: fullbench fuzzer frametest roundTripTest datagen checkFrame decompress-partial freestanding
+all: fullbench fuzzer frametest roundTripTest datagen checkFrame decompress-partial
 
 all32: CFLAGS+=-m32
 all32: all
@@ -182,7 +182,7 @@ list:
 check: test-lz4-essentials
 
 .PHONY: test
-test: test-lz4 test-lz4c test-frametest test-fullbench test-fuzzer test-amalgamation listTest test-decompress-partial test-freestanding
+test: test-lz4 test-lz4c test-frametest test-fullbench test-fuzzer test-amalgamation listTest test-decompress-partial
 
 .PHONY: test32
 test32: CFLAGS+=-m32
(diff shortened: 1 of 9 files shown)


t-mat:
oops.  I have no idea about [macos' restriction](https://github.com/lz4/lz4/runs/7712002588?check_suite_focus=true#step:5:20) 😞   I'll try to resolve it later though.

t-mat:
I've managed macos restriction with *cheat* 😈 

Since `test-freestanding` requires specific environment, I've removed it from "LZ4 standard" test.  So, the following invocation of LZ4 standard tests doesn't test anything about freestanding.

```
make -C tests all
make -C tests test
```

Instead of them, we need to invoke `test-freestanding` explicitly with the following form.

```
make test-freestanding

# or

make -C tests test-freestanding
```

Cyan4973:
> Since `test-freestanding` requires specific environment, I've removed it from "LZ4 standard" test.

I believe that's the correct choice.

Cyan4973 (approved):

Cyan4973:
Could we get some documentation about this new `LZ4_FREESTANDING` capability ?

t-mat:
Yes, I'll add it to this PR soon.

Cyan4973 (approved):

tests/Makefile
```diff
@@ -606,4 +609,10 @@ test-decompress-partial : decompress-partial decompress-partial-usingDict
 	@echo "\n ---- test decompress-partial-usingDict ----"
 	./decompress-partial-usingDict$(EXT)
 
+test-freestanding: freestanding
+	@echo "\n ---- test freestanding ----"
+	./freestanding$(EXT)
+	-strace ./freestanding$(EXT)
```
Cyan4973, line 615: Quick question about this test :
What is the purpose of invoking `strace` and `ltrace` ?
Is it informational only ?
The presence of `-` in front of these lines makes it seem so.

Was there an initial intention to make these tests able to fail, if some condition of the executable was not met ?
For example, if `ltrace` returns anything?

cc @t-mat

t-mat: It's just for information.

Since `strace` and `ltrace` are not POSIX compatible command (right?), I put `-` for them.
I belive Linux and some BSD-like OSes have them, but not sure about wider compatibility.

PR #1136, opened 2022-08-12
Support MSVC 2022 (version 17.3)

t-mat:
This PR adds MSVC 2022 project file and minor fixes for false positive warnings.

### DIFF
--- a/lib/lz4.h
+++ b/lib/lz4.h
@@ -311,10 +311,12 @@ LZ4LIB_API int LZ4_decompress_safe_partial (const char* src, char* dst, int srcS
 ***********************************************/
 typedef union LZ4_stream_u LZ4_stream_t;  /* incomplete type (defined later) */
 
+#if !defined(RC_INVOKED) /* https://docs.microsoft.com/en-us/windows/win32/menurc/predefined-macros */
 #if !defined(LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION)
 LZ4LIB_API LZ4_stream_t* LZ4_createStream(void);
 LZ4LIB_API int           LZ4_freeStream (LZ4_stream_t* streamPtr);
 #endif /* !defined(LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION) */
+#endif
 
 /*! LZ4_resetStream_fast() : v1.9.0+
  *  Use this to prepare an LZ4_stream_t for a new chain of dependent blocks
@@ -398,10 +400,12 @@ typedef union LZ4_streamDecode_u LZ4_streamDecode_t;   /* tracking context */
  *  creation / destruction of streaming decompression tracking context.
  *  A tracking context can be re-used multiple times.
  */
+#if !defined(RC_INVOKED) /* https://docs.microsoft.com/en-us/windows/win32/menurc/predefined-macros */
 #if !defined(LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION)
 LZ4LIB_API LZ4_streamDecode_t* LZ4_createStreamDecode(void);
 LZ4LIB_API int                 LZ4_freeStreamDecode (LZ4_streamDecode_t* LZ4_stream);
 #endif /* !defined(LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION) */
+#endif
 
 /*! LZ4_setStreamDecode() :
  *  An LZ4_streamDecode_t context can be allocated once and re-used multiple times.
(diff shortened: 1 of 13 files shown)


t-mat:
Ugh, I don't understand why AppVeyor reports the following error

https://ci.appveyor.com/project/YannCollet/lz4-1lndh/builds/44453956

```
Pull request #1136 - Support MSVC 2022 (version 17.3)
Fix: remove unused value
This patch fixes the following error from "make staticAnalyze"

datagencli.c:106:21: warning: Value stored to 'size' is never read
                    size=0;
                    ^    ~
```

```
an hour ago by Takayuki Matsuoka
dev 8a5b3b5a ← msvc-17.3 f7e95939
Failed 10 minutes ago in 11 min 35 sec
```

But f7e95939 removes this line 😕

Cyan4973:
This topic raises a larger question about maintenance of Visual Studio projects.

One issue with having multiple supported projects for multiple versions of Visual Studio is maintenance workload.
Every time a file is added or changes its name, it's a nuisance to go through all the build systems to fix that.
This is especially true for Visual Studio which seems to only support explicit names, instead of glob expressions.
This maintenance effort acts as a deterrence, meaning that contributors will unconsciously avoid this situation, and therefore avoid any contribution that could mess the build system, sometimes specifically because of Visual Studio projects support.

As a consequence, I would prefer a reduction in support maintenance of build systems.
For Visual Studio, this *could* be achieved by leveraging `cmake`, and using `cmake` to automatically generate all corresponding project files for all compatible versions of Visual Studio.
This process could be automated as part of `make all`, or `make visual` for example,
so that end users don't have to run `cmake` on their side.

I'm not yet sure if this is a sensible plan.
There might be a need for additional resources that aren't automatically generated by `cmake`.
And of course, the actual source code may need some adjustments to support new compiler versions or new flags, so not _everything_ is automated when adding support for a new version.

But the more important objective is : it should be possible to add a source file into the `lz4` project, without needing to manually edit all the Visual Studio project files

t-mat:
I'm not sure about older version of MSVC, but recent versions of VS/MSVC is supporting `cmake`.

https://docs.microsoft.com/en-us/cpp/build/cmake-projects-in-visual-studio

So it seems good time to introduce `cmake` as a default build system for MSVC.

Since lz4/build/cmake/CMakeLists.txt is not identical to lz4/build/VS*, I think we should create new CMakeLists.txt for MSVC firstly and maintain it for while.
And will merge it to lz4/build/cmake/CMakeLists.txt step by step.

I'd like to avoid including new cmake build script for MSVC in the next release because it may require some time to stabilize.

Cyan4973:
Agreed

lib/lz4.h
```diff
@@ -311,10 +311,12 @@ LZ4LIB_API int LZ4_decompress_safe_partial (const char* src, char* dst, int srcS
 ***********************************************/
 typedef union LZ4_stream_u LZ4_stream_t;  /* incomplete type (defined later) */
 
+#if !defined(RC_INVOKED) /* https://docs.microsoft.com/en-us/windows/win32/menurc/predefined-macros */
```
Cyan4973, line 327: I don't understand the relation between `RC_INVOKED` and following declarations.
Could you explain ?

t-mat: (1) Truncation causes the warning "RC4011: identifier truncated"

`rc.exe` reports the following warnings

```
lz4\build\VS2022\..\..\lib\lz4.h(314): warning RC4011: identifier truncated to 'LZ4_STATIC_LINKING_ONLY_DISABLE'
lz4\build\VS2022\..\..\lib\lz4.h(401): warning RC4011: identifier truncated to 'LZ4_STATIC_LINKING_ONLY_DISABLE'
```

The reason of the warning is trunction.  Since `rc` is a legacy resource compiler, it truncates long symbols.


(2) `rc.exe` predefines `RC_INVOKED`

https://docs.microsoft.com/en-us/windows/win32/menurc/predefined-macros

`rc.exe` predefines `RC_INVOKED`.  And we can use it to determine that header is being processed by `rc.exe` or not.


(3) Use `RC_INVOKED` to avoid truncation warnings.

The following `#if !defined()` block means "skip this block when `rc.exe` is trying to read this block."

```
#if !defined(RC_INVOKED)
...
#endif
```

I surround `LZ4_STATIC_LINKING_ONLY_DISABLE` with this block to avoid the warning.

----

Other possible solution may be surrounding entire header except version number with `#if !defined(RC_INVOKED)`.

Cyan4973: It's a good explanation.
I believe it does not transpire from the provided link only.
Could you add a word about the reason of this presence in code documentation ?

Cyan4973 (approved):

PR #1147, opened 2022-08-19
Add dependabot

tristan957:
Dependabot is configured to update GitHub actions. It will check weekly.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,7 @@
+version: 2
+
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly


Cyan4973:
What's the purpose of dependabot ?

felixhandte:
Looks like dependabot is a GitHub-maintained bot that checks whether you have out-of-date dependencies and will file an issue/PR if it detects that you do. The utility is questionable, since LZ4 doesn't have any dependencies...

I dunno, maybe it's useful for other users of dependabot and LZ4 to see that they don't transitively have out-of-date deps via LZ4?

tristan957:
LZ4 has GitHub Action dependencies which is what I have configured the bot to update.

Here is an example PR: https://github.com/hse-project/hse-java/pull/18.

Cyan4973:
It's still not completely clear to me what are the benefits, 
though it seems this is something we can try with limited risks on the library itself.
So OK, let's try it out.
We'll discover the impact on CI behavior.

t-mat:
@Cyan4973, actual source code and build script of lz4 doesn't have any external dependency,
but GitHub Actions contains the following lines which depends external code.

https://github.com/lz4/lz4/blob/v1.9.4/.github/workflows/ci.yml#L78

```
    - uses: actions/checkout@v2 # https://github.com/actions/checkout
```


https://github.com/lz4/lz4/blob/v1.9.4/.github/workflows/ci.yml#L500

```
      uses: actions/upload-artifact@v1
```

Since there're "supply chain attack" nowadays, we must check actual update of the external code as usual.  But it may help us to notice updates.

tristan957:
The main benefits are that when there are updates to dependencies of the CI, you will automatically be notified of it.

Since actions are built with JavaScript, it is important to stay up to date on your dependencies.

dearblue:
Hello,
For your information, a PullRequest may also be sent to the fork repository.
To disable this, it seems that the fork repository needs to be deleted once.

https://github.com/dependabot/dependabot-core/issues/2804#issuecomment-737781797

eli-schwartz:
Isn't the usual recommendation to solve the forks issue, to use Renovate instead of Dependabot?

Issue #1157, opened 2022-09-08
[performance](lz4f) why LZ4F_compressEnd set maxBufferSize=0 


yixiutt:
I use a global LZ4F_cctx  and i found that LZ4F_compressEnd set maxBufferSize=0 so in LZ4F_compressBegin, tmpBuffer buffer will be free and calloc a new memory。
Why tmpBuffer cannot be reused in LZ4F_cctx?

Cyan4973:
Indeed,

this change was made in https://github.com/lz4/lz4/pull/236,
in order to fix https://github.com/lz4/lz4/issues/232,
an unspecified bug in the fuzzer.

I must admit that I don't quite follow the relation.
The original intention was to re-use the buffer,
but somehow, forcing `maxBufferSize=0` seems to have fixed _something_,
though I have difficulties to link the problem and the solution.

Anyway, worth an investigation...

yixiutt:
OK, I will delete `maxBufferSize=0` and test in my case. I'll let you know if I make some progress

Cyan4973:
I also don't quite understand why `calloc()` is used for `tmpBuffer`,
given the use case for this buffer, I would have expected `malloc()` to be sufficient.

Conjecture : it may have been a case of silencing `msan`. 
In which case, I don't think it was the right fix ...

yixiutt:
I agree with u

yixiutt:
tmpBuffer may not need to load into physical memory cause tmp buffer always bigger than actual use, so calloc here would cause a lot page fault than expected

Cyan4973:
Fixed in #1158

PR #1158, opened 2022-09-08
allocation optimization for lz4frame compression

### LINKED ISSUE #232: A bug in frametest
`d:\GitHub\lz4\programs>frametest -s9469 -t111080`
Starting lz4frame tester (64-bits, r132)
Seed = 9469
111080   Error at pos 8/334721 : 95 != AF
Error => Decompression failed (error -18:ERROR_contentChecksum_invalid) (seed 9469, test nb 111086)

Tested with `dev` branch.

Cyan4973:
as noted by @yixiutt in #1157, the temporary buffer managed by lz4frame compression context is being invalidated at end of compression, forcing it to be re-allocated at next compression job.

This shouldn't be necessary.
This behavior was introduced in #236, as a way to fix #232, but neither the issue is explained, nor why the patch fixes it.

This patch reverts to previous behavior,
where temporary buffer is reused between compression calls. 
This results in a net reduction of allocation workload.

Additionally, the temporary buffer should only need `malloc()`, not `calloc()`, thus saving some potential 0-initialization cost.

This diff implements both changes. It's expected to improve compression speed when repetitively compressing small data.

Performance impact on M1 laptop : 
| filename | size | cSpeed before | cSpeed after | delta |
| --- | --- | --- | --- | --- |
| enwik7 | 1000000 | 485 MB/s | 485 MB/s | +0% |
| enwik6 | 100000 | 505 MB/s | 508 MB/s  | +1% |
| enwik5 | 10000 | 862 MB/s | 1037 MB/s |  +20% |
| enwik4 | 1000 | 359 MB/s | 940 MB/s | +160% |

As expected, performance difference is only perceptible for small data, but it can matter a lot in this case.

Once this diff is merged, long fuzzer tests will be run to ensure that no sanitizer warning gets triggered.

Additionally :
- fixed a minor ubsan warning in `LZ4F_decompress()`
- added an `LZ4F_compressUpdate()` test to `fullbench`

### DIFF
--- a/lib/lz4frame.c
+++ b/lib/lz4frame.c
@@ -264,7 +264,7 @@ typedef struct LZ4F_cctx_s
     LZ4F_CustomMem cmem;
     LZ4F_preferences_t prefs;
     U32    version;
-    U32    cStage;
+    U32    cStage;     /* 0 : compression uninitialized ; 1 : initialized, can compress */
     const LZ4F_CDict* cdict;
     size_t maxBlockSize;
     size_t maxBufferSize;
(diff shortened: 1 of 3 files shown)


yixiutt:
LGTM

PR #1159, opened 2022-09-08
attempt to enable ubsan tests in CI

Cyan4973:
let's see if these tests pass now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -396,25 +396,18 @@ jobs:
   lz4-ubsan-x64:
     name: Linux x64 ubsan
     runs-on: ubuntu-latest
-    env:                        # Set environment variables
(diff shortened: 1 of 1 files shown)


Cyan4973:
OK, it seems to work fine.
That's partly because `ubsan` tests do not automatically terminate process with an error when they detect a UB warning.

In this case, there are still a nb of pointer overflow situations which get detected.
So, if the `ubsan` test was designed to fail on detecting any UB, the CI test would fail.
In this case though, we specifically opt out of pointer-overflow ubsan warnings,
using flags : `-fno-sanitize-recover=undefined -fsanitize-recover=pointer-overflow`

It's okay.
It's still a pretty good coverage of `ubsan` warnings.
And for the remaining ones, I can spend some time to look at them.
After the recent set of fixes, there are not that many remaining now.

PR #1171, opened 2022-09-21
Update snapcraft.yaml to reflect build of v1.9.4

dosaboy:


Cyan4973 (approved):

Cyan4973:
To be fair, while this update fixes the problem at hand,
I'm slightly concerned that this `snapcraft.yaml` file might always be incorrect at release time.

When next release comes in, it will be, say, `v1.9.5`,
and if this file is once again forgotten,
it will still mention `v1.9.4`.

One could say "just make this file update part of the release process",
but with barely 1 release per year, this is easy to forget,
and it makes the release process heavier (how many others "you just have to"?)

What works better for this case
is to automate the update,
typically by generating the file, or updating the file,
from `Makefile`, and typically make it part of `make all`.

`make all` will be run at least once before each release,
so anything it builds and updates will be merged before tagging the release.

Looks to me that only the line `version:` needs to be updated.
Other deeper changes might be desirable at times, but these ones can't be automated.

dosaboy:
@Cyan4973 yes i totally agree and yes it could be templated/autogenerated

PR #1177, opened 2022-10-07
Basic configuration for lz4 cpp swift package

Archdoog:
- Creates Package.swift
- Adds module map for swift package public headers.


richardtop:
Why is his closed?

Archdoog:
I ended up just forking and using https://github.com/Archdoog/lz4-cpp-SwiftPM. At some point later I ended up just having to use cmake built static libraries, so I'm no longer using lz4 as a swift package. 

That said, feel free to take this and run with it. I'd be happy to review if you tag me.

PR #1185, opened 2022-11-05
Add arm64 assembly decoder

Siguza:
I wrote a tiny LZ4 block decoder in arm64 assembly for use in freestanding environments.   
Figured it would be nice to link it here, since you have a 32-bit ARM one already. :)


Cyan4973:
Great work @Siguza !

PR #1187, opened 2022-11-09
Add environment check for freestanding test : resolves #1186

### LINKED ISSUE #1186: freestanding test fails on FreeBSD 13.1-STABLE
**Describe the bug**
freestanding test fails with either SIGBUS (Clang) or SIGSEGV (GCC)

**Expected behavior**
Test to succeed

**To Reproduce**
Execute freestanding binary

 - OS: FreeBSD (amd64)
 - Version 13.1-STABLE
 - Compiler: Clang 14.0.5 / GCC 11.3.0
 - Build System: Makefile or Meson
 - Other hardware specs: amd64, Intel Tigerlake CPU

**Additional context**
gdb output (GCC binary):

```
Program received signal SIGSEGV, Segmentation fault.
Address not mapped to object.
0x0000000000000000 in ?? ()
(gdb) bt full
#0  0x0000000000000000 in ?? ()
No symbol table info available.
#1  0x0000000000000001 in ?? ()
No symbol table info available.
#2  0x00007fffffffece0 in ?? ()
No symbol table info available.
#3  0x0000000000000000 in ?? ()
No symbol table info available.
```

This with using both 1.9.4 (Meson backported) and latest master (Makefile/Meson)
Might as well add that this occurs without any optimization enabled (-O0)

Also confirmed on 13.1-RELEASE
Reference: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=266063

t-mat:
Freestanding test is intended to run on only Linux x86_64 platform. Because it uses Linux x86_64 syscall directly to avoid any dependency to the standard library.

This changeset adds platform checking code to tests/Makefile to avoid unintended error in non-Linux x86_64 platforms.

This changeset resolves issue #1186

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -115,9 +115,6 @@ decompress-partial: lz4.o decompress-partial.c
 decompress-partial-usingDict: lz4.o decompress-partial-usingDict.c
 	$(CC) $(FLAGS) $^ -o $@$(EXT)
 
-freestanding: freestanding.c
-	$(CC) -ffreestanding -nostdlib $^ -o $@$(EXT)
-
 .PHONY: clean
 clean:
 	@$(MAKE) -C $(LZ4DIR) $@ > $(VOID)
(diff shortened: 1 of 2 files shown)


Cyan4973 (approved):

tristan957:
See my PR for Meson fix.

t-mat:
Note about plan of consequent PR:
As an adaptation of #1184, I'll add `test-freestanding` to the `all` of `tests/Makefile`.
Since we have `macos` platform, it'll be safeguard of same issue.

PR #1190, opened 2022-11-22
Fixed const-ness of src data pointer in lz4file and install lz4file.h

vsolontsov-volant:


Cyan4973 (approved):
Good point !
Indeed, this ptr parameter should have been a `const` all along : this function is writing _into a file_ from a read buffer.

lib/Makefile
```diff
@@ -207,6 +207,7 @@ install: lib liblz4.pc
 	$(INSTALL_DATA) lz4.h $(DESTDIR)$(includedir)/lz4.h
 	$(INSTALL_DATA) lz4hc.h $(DESTDIR)$(includedir)/lz4hc.h
 	$(INSTALL_DATA) lz4frame.h $(DESTDIR)$(includedir)/lz4frame.h
+	$(INSTALL_DATA) lz4file.h $(DESTDIR)$(includedir)/lz4file.h
```
Cyan4973, line 210: My understanding is that you wanted to provide this additional header file in a separate PR ?

vsolontsov-volant: That was an request from @t-mat. Anyway this change is wrong -- `lz4file.h` should be installed together with `lz4frame_static.h`while here I put it with all the base API.

PR #1198, opened 2022-12-03
Add `freestanding` to `all` and `test` of `tests/Makefile`

t-mat:
Follow up of #1184 and #1187

Now `tests/Makefile` contains

- `freestanding` as a member of `all`.
- `test-freestanding` as a member of `test`.

Note that as for `all32` (`-m32`) builds, `test-freestanding` skips its actual test.

----
```
cd /path/to/lz4
cd tests

make all
# cc -ffreestanding -nostdlib freestanding.c -o freestanding

make test
#  ---- test freestanding ----
# ./freestanding
# strace ./freestanding
# ...
# +++ exited with 0 +++
# ltrace ./freestanding
# +++ exited (status 0) +++
```

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -56,7 +56,7 @@ NB_LOOPS     ?= -i1
 .PHONY: default
 default: all
 
-all: fullbench fuzzer frametest roundTripTest datagen checkFrame decompress-partial
+all: fullbench fuzzer frametest roundTripTest datagen checkFrame decompress-partial freestanding
 
 all32: CFLAGS+=-m32
 all32: all
@@ -180,7 +180,7 @@ list:
 check: test-lz4-essentials
 
 .PHONY: test
-test: test-lz4 test-lz4c test-frametest test-fullbench test-fuzzer test-amalgamation listTest test-decompress-partial
+test: test-lz4 test-lz4c test-frametest test-fullbench test-fuzzer test-amalgamation listTest test-decompress-partial test-freestanding
 
 .PHONY: test32
 test32: CFLAGS+=-m32
@@ -373,6 +373,10 @@ ifneq ($(UNAME_S), Linux)
 endif
 
 ifneq ($(UNAME_P), x86_64)
+  FREESTANDING_CFLAGS :=
+endif
+
+ifneq (,$(findstring -m32,$(CFLAGS)))
   FREESTANDING_CFLAGS :=
 endif
 


t-mat:
Sorry, I forgot about `ubsan32` case.  I'll resolve this error later.

Cyan4973:
> Note that as for `all32` (`-m32`) builds, `test-freestanding` skips its actual test.

Could you tell me why `freestanding` (or just its tests?) is not compatible with 32-bit ?

t-mat:
`-m32` generates `i386` code but `os_syscall1()` uses `rax` and 64-bits registers (and its own stack frame).

https://github.com/lz4/lz4/blob/1f3adea1c68ea61f41903d959b4f98f45aa527a7/tests/freestanding.c#L156-L161

Also, due to lack of my experience with freestanding Linux environment, I just avoid to #include almost everything.  As a result, the following line causes incompatibility:

https://github.com/lz4/lz4/blob/1f3adea1c68ea61f41903d959b4f98f45aa527a7/tests/freestanding.c#L154

Because each platform has its own numbers for syscall

x64 (exit = `60`):
https://github.com/torvalds/linux/blob/v4.17/arch/x86/entry/syscalls/syscall_64.tbl#L71

x86 (exit = `1`):
https://github.com/torvalds/linux/blob/v4.17/arch/x86/entry/syscalls/syscall_32.tbl#L15

We may need to include `<sys/syscall.h>` for portability.

Cyan4973:
Thanks for the clear explanation @t-mat !

> We may need to include `<sys/syscall.h>` for portability.

Isn't the idea of including _anything_ going against the very concept of freestanding ?

_edit_ : 
OK, sorry, now I understand that these specificities happen in the test program `tests/freestanding.c`,
not in the `lz4` freestanding library.

In which case, of course, the test itself can be specific to certain platforms.

Maybe this could be called out explicitly in the unit name ?

t-mat:
> these specificities happen in the test program tests/freestanding.c, not in the lz4 freestanding library.

Right.  If we can successfully run `lz4` freestanding library without standard library (`libc`, etc) but only depends platform's own syscall, we can say it conforms the concept of the freestanding (bare bone C environment).

> Maybe this could be called out explicitly in the unit name ?

Yes.  It'd be nice to have explicit / specific name (and maybe test) for them.

Cyan4973:
Hi @t-mat ,

it seems I completely lost track of this PR.

If I understand correctly, all it does is enable `freestanding` test automatically as part of `make all` and `make test`,
as well as ensuring it's setup properly for 32-bit `x86` target.

Is it still valid as is ?
Could it rebased ?
(I see there was a `ubsan` test failure, presumably due to some external flaky condition)

t-mat:
Yeah, I also suspect this PR is outdated.  I've tried to implement wider support of freestanding test.  But it seems it's been failed.

Please close this PR for now.
When I write better test for other platforms/architectures, I'll send another PR.

PR #1199, opened 2022-12-17
Improve LZ4F_decompress() docs

embg:
zstd requires users to pass the frame header along with the frame content, while lz4 separates the header and the content. This PR adds an explicit warning for users coming from zstd who are used to passing in the header and content together.

### DIFF
--- a/lib/lz4frame.h
+++ b/lib/lz4frame.h
@@ -473,6 +473,11 @@ LZ4F_getFrameInfo(LZ4F_dctx* dctx,
  * `dstBuffer` can freely change between each consecutive function invocation.
  * `dstBuffer` content will be overwritten.
  *
+ *  Note: if `LZ4F_getFrameInfo()` is called before `LZ4F_decompress()`, srcBuffer must be updated to reflect
+ *  the number of bytes consumed after reading the frame header. Failure to update srcBuffer before calling
+ *  `LZ4F_decompress()` will cause decompression failure or, even worse, successful but incorrect decompression.
+ *  See the `LZ4F_getFrameInfo()` docs for details.
+ *
  * @return : an hint of how many `srcSize` bytes LZ4F_decompress() expects for next call.
  *  Schematically, it's the size of the current (or remaining) compressed block + header of next block.
  *  Respecting the hint provides some small speed benefit, because it skips intermediate buffers.


lib/lz4frame.h
```diff
@@ -487,6 +487,10 @@ LZ4F_getFrameInfo(LZ4F_dctx* dctx,
  *  Use LZ4F_resetDecompressionContext() to return to clean state.
  *
  *  After a frame is fully decoded, dctx can be used again to decompress another frame.
+ *
+ *  Note: the frame header should not be passed into this function. Decompression of a frame
+ *  should start with the first byte following the frame header. Use LZ4F_headerSize() to determine
+ *  where the frame content starts.
```
Cyan4973: Not exactly.
It's possible to invoke `LZ4F_decompress()` directly, at the beginning of a frame.
Or at least, it should be, and if that's not the case, then there is a bug.

I suspect the issue is rather on `LZ4F_getFrameInfo()` side.
As mentioned in the documentation of this method: 
```
Its usage is optional: user can also invoke LZ4F_decompress() directly.
```

So that's the condition: `LZ4F_decompress()` can be called at the beginning of a frame,
but in this case, the optional method `LZ4F_getFrameInfo()` must not have been used.

The reason is : `LZ4F_getFrameInfo()` is effectively _starting_ the decompression process, it's a part of it. And as a consequence, it's _consuming_ the header : 
```
The number of bytes consumed from srcBuffer will be updated in *srcSizePtr (necessarily <= original value).
LZ4F_getFrameInfo() only consumes bytes when decoding has not yet started,
and when decoding the header has been successful.
```

This is in contrast with `zstd`, where the equivalent `ZSTD_getFrameHeader()` only _reads_ the header, but does not start the decompression process.

embg: Ya, the bug we got was from missing this bit of documentation: https://github.com/lz4/lz4/blob/073ad37f1fe1f628b1b3171410088a50f3ac165f/lib/lz4frame.h#L448
We called `LZ4F_getFrameHeader()` and then tried to decompress from the beginning of the frame. 

Do you think it would be useful to add a note about this in the `LZ4F_decompress()` docs as well? To help out poor souls like myself who failed to check the `LZ4F_getFrameHeader()` docs properly :)

Cyan4973: Sure !

Cyan4973:
Any update planned for this PR ?

Cyan4973 (approved):

PR #1206, opened 2023-01-10
Clean up generation of internal static library

tristan957:
No sense in relinking if the default_library indicates that a static library was already generated.


Cyan4973 (approved):

tristan957:
@Cyan4973 do you think this could be merged? Seems like @eli-schwartz has a PR which would cause a conflict.

Cyan4973:
Sure!

eli-schwartz:
Rebasing my PR on top of this change was super simple, done. :)

PR #1207, opened 2023-01-16
build: move meson files from contrib, to go alongside other build systems

eli-schwartz:
This is not really "contrib" stuff, there's a standard directory for all the included project configurations.


eli-schwartz:
After the rebase, test lz4hc-hugefile times out in 360 seconds, which seems like it's probably a flake and will work if the job is restarted?

Cyan4973 (approved):

tristan957 (approved):

tristan957:
@Cyan4973 anything holding this back?

PR #1209, opened 2023-02-13
Add 64-bit detection for LoongArch

zhaixiaojuan:
The LoongArch architecture (LoongArch) is an Instruction Set Architecture (ISA) that has a RISC style.

Documentations:
ISA:
https://loongson.github.io/LoongArch-Documentation/LoongArch-Vol1-EN.html
ABI:
https://loongson.github.io/LoongArch-Documentation/LoongArch-ELF-ABI-EN.html
More docs can be found at:

### DIFF
--- a/programs/platform.h
+++ b/programs/platform.h
@@ -48,6 +48,7 @@ extern "C" {
   || defined __x86_64__s || defined _M_X64                                                                          /* x86 64-bit */    \
   || defined __arm64__ || defined __aarch64__ || defined __ARM64_ARCH_8__                                           /* ARM 64-bit */    \
   || (defined __mips  && (__mips == 64 || __mips == 4 || __mips == 3))                                              /* MIPS 64-bit */   \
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Cyan4973:
As mentioned in the other PR, it's better if we can have a CI test associated (ensuring the library actually works on target cpu).
If not, or if too complex to execute, well it's not the end of the world, we would just have to accept that maybe some future change might impact the target architecture, and this would go unnoticed, due to lack of test. Not ideal, but sometimes, that's the only option left while keeping complexity under control.

t-mat:
@Cyan4973 I have no idea/experience with it, but it seems QEMU 7 has been starting to support LoongArch64.  I hope we'll be able to test it in our CI.

Cyan4973:
Yeah, indeed, I've seen this announcement too,
but `qemu` 7.1 seems to be _very_ recent,
so I'm not too hopeful that CI can support it yet...

PR #1212, opened 2023-02-26
refuse to compress directories

### LINKED ISSUE #1211: Regression when input file is a directory
**Describe the bug**

As of 1.9.3, `mkdir d; lz4 d` reports `d: is not a regular file`.

But in 1.9.4, it says `d: Permission denied` on Win and `Error 40 : Error reading d` on WSL. And it creates a 0KB *d.lz4* on WSL.

**System:**
 - OS: Windows 11 and WSL Debian 12
 - Version 1.9.4

Cyan4973:
fix #1211, reported by @imba-tjd

### DIFF
--- a/programs/util.h
+++ b/programs/util.h
@@ -195,10 +195,7 @@ UTIL_STATIC void* UTIL_realloc(void* ptr, size_t size)
 /*-****************************************
 *  String functions
 ******************************************/
-/*
- * A modified version of realloc().
- * If UTIL_realloc() fails the original block is freed.
-*/
+/* supports a==NULL or b==NULL */
 UTIL_STATIC int UTIL_sameString(const char* a, const char* b)
 {
     assert(a!=NULL && b!=NULL);  /* unsupported scenario */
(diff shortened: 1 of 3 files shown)


imba-tjd:
Why not use existing `UTIL_isDirectory` 🤔

programs/util.h
```diff
@@ -195,10 +195,7 @@ UTIL_STATIC void* UTIL_realloc(void* ptr, size_t size)
...
 *  String functions
 ******************************************/
-/*
- * A modified version of realloc().
- * If UTIL_realloc() fails the original block is freed.
-*/
+/* supports a==NULL or b==NULL */
 UTIL_STATIC int UTIL_sameString(const char* a, const char* b)
 {
     assert(a!=NULL && b!=NULL);  /* unsupported scenario */
```
t-mat: Currently, if `a == NULL` *or* `b == NULL`, this line rises assert.
But above comment and the following `if()` clause seems to support these cases.

If so, this assert should refuse only `a == NULL` *and* `b == NULL` case.
Therefore, expected condition may be `assert(! (a == NULL && b == NULL));`.
In short, `assert(a != NULL || b != NULL);`.

Cyan4973: Good point @t-mat !
Thanks, let's fix that as you suggest.

On a related tangent, since this `assert()` was never triggered so far, it's questionable if this is a use case worth supporting, since it seems unused.

But anyway, this function was not the point of this PR, 
I just wanted to fix the code comment which was nonsense.

Cyan4973:
> Why not use existing `UTIL_isDirectory` 🤔

Good point @imba-tjd , I totally missed its existence ...

PR #1214, opened 2023-03-01
Move assert() in LZ4_decompress_fast_continue() on separate line

benpicco:
This makes the code easier to read, but also helps with `assert()` implementations such as [this](https://godbolt.org/z/4v35rrTvn).

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -2593,9 +2593,11 @@ LZ4_FORCE_O2 int
 LZ4_decompress_fast_continue (LZ4_streamDecode_t* LZ4_streamDecode,
                         const char* source, char* dest, int originalSize)
 {
-    LZ4_streamDecode_t_internal* const lz4sd =
-        (assert(LZ4_streamDecode!=NULL), &LZ4_streamDecode->internal_donotuse);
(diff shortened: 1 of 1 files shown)


benpicco:
I found a solution for my assert issue, so this might as well be closed.

lib/lz4.c
```diff
@@ -2593,9 +2593,11 @@ LZ4_FORCE_O2 int
 LZ4_decompress_fast_continue (LZ4_streamDecode_t* LZ4_streamDecode,
                         const char* source, char* dest, int originalSize)
 {
-    LZ4_streamDecode_t_internal* const lz4sd =
-        (assert(LZ4_streamDecode!=NULL), &LZ4_streamDecode->internal_donotuse);
     int result;
+    LZ4_streamDecode_t_internal* const lz4sd = &LZ4_streamDecode->internal_donotuse;
+
+    assert(LZ4_streamDecode != NULL);
```
Cyan4973, line 2599: It's too late,
the test `LZ4_streamDecode != NULL` must happen before dereferencing `LZ4_streamDecode`.

PR #1220, opened 2023-04-08
lib/Makefile: Support building on legacy OS X

sevan:
Toolchain on legacy versions require -dynamiclib to be stated for a shared object to be generated. Tested on OS X 10.4 (GCC) and 10.15 (clang)

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -65,7 +65,7 @@ ifeq ($(TARGET_OS), Darwin)
 	SHARED_EXT = dylib
 	SHARED_EXT_MAJOR = $(LIBVER_MAJOR).$(SHARED_EXT)
 	SHARED_EXT_VER = $(LIBVER).$(SHARED_EXT)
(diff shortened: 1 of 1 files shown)


Cyan4973:
This PR seems to add this flag to *all* versions of `osx`, including recent ones, which do not seem to need it.

Is there a sort of guarantee that this flag cannot be detrimental to newer versions ?

sevan:
It shouldn't be an issue as `-dynamiclib` is still supported by [clang](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-dynamiclib) though the option was [originally from GCC](https://gcc.gnu.org/onlinedocs/gcc/Darwin-Options.html).
[The macOS build via github action built it successfully](https://github.com/lz4/lz4/actions/runs/4643262734/jobs/8217787232) and that's one major release behind (macOS 12 vs 13).

Cyan4973 (approved):

PR #1221, opened 2023-04-08
Improve decompression speed

Nicoshev:
Hello, I hope you are doing well,

I wanted to propose three changes which yield decompression speed improvements.

1. Optimizing LZ4_memcpy_using_offset

When the offset is a power of 2 lower or equal than sizeof(reg_t), instead of loading sizeof(reg_t) bytes onto the stack and doing memcpy, a variable of type reg_t is filled using only 1 multiplication. Then, the variable is written onto the output on a memset style.

The third parameter BYTE* dstEnd was changed to size_t length. This has two advantages: In copying loops stop conditions, the source pointer is compared against an immediate numbers instead of another variable; On LZ4_decompress_generic, at the end of the unsafe decode loop (lines ~2130 through ~2135), it is not needed to set and maintain the variable cpy while doing the copy onto the output.

2. Optimizing reading length variables

MINMATCH is added at the time the copy operation's length token is read. This reduces the size of the following code branches, which should help CPU's branch predictors.

rvl_error value has been changed to LZ4_MAX_INPUT_SIZE + 1. This one is controversial but yielded noticeable performance improvements. Now read_variable_length returns LZ4_MAX_INPUT_SIZE + 1 on error, and its result is directly added to the length variable, without needing to maintain the additional addl variable. Then, if length is greater than LZ4_MAX_INPUT_SIZE, the decompression routine follows the error path.

The deletion of the additional addl variable reduced the need to use the stack, which yielded the noticed performance improvements. Maybe rvl_error can be refactored in a more descriptive way to preserve code readability while keeping the performance gains.

3. Removing a redundant code branch

Within LZ4_decompress_generic, on line ~2061, the following error condition is checked:
(checkOffset) && (unlikely(match + dictSize < lowPrefix))
Any of the two code branches that follow also check such condition, so it should be safe to remove it.

With these three changes, I was able to get about 4% decompression speed improvement when processing the silesia corpus. On other workloads, the gains went up to about 10%.

I hope you find these changes useful,
Nicolas

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -2122,7 +2122,6 @@ LZ4_decompress_generic(
 
             assert((op <= oend) && (oend-op >= 32));
             if (unlikely(offset<16)) {
-                assert(op + length >= dstPtr + MINMATCH);
                 LZ4_memcpy_using_offset(op, match, length, offset);
             } else {
                 LZ4_wildCopy32(op, match, op + length);
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -375,6 +375,7 @@ static reg_t LZ4_read_ARCH(const void* memPtr) { return *(const reg_t*) memPtr; 
 
 static void LZ4_write16(void* memPtr, U16 value) { *(U16*)memPtr = value; }
 static void LZ4_write32(void* memPtr, U32 value) { *(U32*)memPtr = value; }
+static void LZ4_write_ARCH(void* memPtr, reg_t value) { *(reg_t*)memPtr = value; }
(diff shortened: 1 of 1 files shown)


Cyan4973:
These look like good improvement ideas @Nicoshev .

Does it seem possible to separate them into 3 separate PR ?

It would make it easier to evaluate their benefits and complexity, and to isolate drawbacks if there are any.

Nicoshev:
Thanks @Cyan4973, for sure, I'll prepare three separate PRs

PR #1224, opened 2023-04-08
Remove redundant error check

Nicoshev:
Within LZ4_decompress_generic, on line 2061, the following error condition is checked:

(checkOffset) && (unlikely(match + dictSize < lowPrefix))

Any of the two code branches that follow also check such condition, so it should be safe to remove it.


Cyan4973 (approved):
Yes, that sounds reasonable

PR #1225, opened 2023-04-09
Add packing support for msc

Nicoshev:
Hello, I hope you are doing well

I wanted to create this additional PR, which adds support for packed structures when using the Microsoft compiler

Regards,
Nicolas


Cyan4973:
Does that bring some benefit ? like better performance for example ?

Nicoshev:
Yes, I saw about 2% to 3% improvement on decompression speed

Nicoshev:
Yes, it brought performance improvements when we tested it last year alongside other changes. I believe the impact of this change was a gain of around 2% to 3%, although I do not remember if it was on compression or decompression or both. The tests we did were not on the Silesia corpus though, so I do not know the impact when operating on that set of files. I can do the test though.

Cyan4973 (approved):

PR #1226, opened 2023-04-10
Reduce usage of variable cpy on decompression

Nicoshev:
Hello, I hope you are doing well

I wanted to propose an additional code optimization for the decompression function:

The variable cpy is now not used from the code paths that copy literals within the fast decompression loop.
This optimizes the usage of the CPU registers, as well as reduces the size of certain code branches.
On average, I got slightly over 1% improvement in decompression speed when decompressing the Silesia corpus using an X64 CPU.

This optimization works very well with the one proposed in the PR "Optimize LZ4_memcpy_using_offset".
When both optimizations are applied together, the usage of the variable cpy is completely removed from the fast decompression loop.

Regards,
Nicolas

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -2025,19 +2025,18 @@ LZ4_decompress_generic(
                 if (unlikely((uptrval)(ip)+length<(uptrval)(ip))) { goto _output_error; } /* overflow detection */
 
                 /* copy literals */
-                cpy = op+length;
                 LZ4_STATIC_ASSERT(MFLIMIT >= WILDCOPYLENGTH);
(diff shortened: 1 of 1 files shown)


Cyan4973:
I've been able to observe very small (<1%) gains on some platforms (M1 + `clang`, 9700k + `gcc-11`).
On `x64` "skylake-era" cpus, performance fluctuations are a mess, sometimes increasing (`clang`) or decreasing (`gcc-9`) by a lot, depending on compiler version. This effect seems unrelated to this PR, and is more a symptom of instruction alignment, an uncontrollable side-effect which is particularly sensible on this cpu generation.

In the end, I mostly like that this PR tends to make the code more readable, and it's a good enough reason.

Cyan4973 (approved):

PR #1228, opened 2023-04-13
Set CMake minimum requirement to 3.5

haampie:
`cmake_minimum_required(VERSION 2.x)` has the effect that any recent CMake runs with default settings that were considered sensible a decade ago.

Instead, use `cmake_minimum_required(VERSION 2.x...3.y)` so that newer CMake uses modern defaults for policies, without dropping CMake 2.x support.

One such policy that I need for example is CMP0042, which makes CMake output `@rpath/liblz4.dylib` as install name on macOS, so that packages linking to it register `@rpath/liblz4.dylib` as a load command, so that rpaths will be considered to locate it. Without it, dependents typically fail at runtime and cannot locate the library if it's in a non-system dir.

### DIFF
--- a/build/cmake/CMakeLists.txt
+++ b/build/cmake/CMakeLists.txt
@@ -10,7 +10,7 @@
 # LZ4's CMake support is maintained by Evan Nemerson; when filing
 # bugs please mention @nemequ to make sure I see it.
 
-cmake_minimum_required(VERSION 2.8.12...3.26)
+cmake_minimum_required(VERSION 3.5)
 
 set(LZ4_TOP_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../..")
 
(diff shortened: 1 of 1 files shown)


haampie:
ping @nemequ

Cyan4973:
Note :
the `2.8.12` minimum version requirement is old policy, dated from the time when `CMakeLists.txt` was created, and multiple active distributions used to rely on this version of `cmake`.

By now, things have moved on, and we could update the minimum version, up to a maximum of `3.5.1` if need be.

It's just that there is no point in reducing compatibility (by increasing the minimum version requirement) if there is _no_ benefit in exchange, and so far, no one has been able to articulate what is gained by updating the minimum version to something more recent.
But this can be debated.

haampie:
Feel free to update the PR accordingly. This is more of an issue promoted to a actionable change :p

Cyan4973 (approved):

Cyan4973:
I think `cmake` has started rolling some warning messages on very recent versions of `cmake`,
which states, in substance, that some future version of `cmake` will stop supporting minimum versions < 3.5.

Consequently, I think this is the minimum version this PR should jump to.
Updating the minimum version to `v3.5` would be enough to fix CMP0042 support.

I prefer a single minimum version number, because otherwise, the `CMakeLists.txt` script will behave differently depending on local versions, increasing support matrix complexity.

Cyan4973 (changes requested):
change minimum to 3.5

haampie:
Done

build/cmake/CMakeLists.txt
```diff
@@ -24,14 +24,7 @@ string(REGEX REPLACE "^#define LZ4_VERSION_RELEASE +([0-9]+) +.*$" "\\1" LZ4_VER
 set(LZ4_VERSION_STRING "${LZ4_VERSION_MAJOR}.${LZ4_VERSION_MINOR}.${LZ4_VERSION_RELEASE}")
 mark_as_advanced(LZ4_VERSION_STRING LZ4_VERSION_MAJOR LZ4_VERSION_MINOR LZ4_VERSION_RELEASE)
 
-if("${CMAKE_VERSION}" VERSION_LESS "3.0")
-  project(LZ4 C)
-else()
-  cmake_policy (SET CMP0048 NEW)
```
Cyan4973, line 30: Question : is it now useless to specify this line ?
`cmake_policy (SET CMP0048 NEW)`

haampie: Yeah, the policy is from 3.0 and with minimum required >= 3.0 it's automatically NEW.

The policies are always such that new cmake behaves the same as the "minimum required" cmake version.

Cyan4973 (approved):

Cyan4973:
Thanks @haampie

PR #1234, opened 2023-05-23
Pedantic suggestions

DimitriPapadopoulos:
Pedantic suggestions by [`checkpatch.pl`](https://github.com/torvalds/linux/blob/master/scripts/checkpatch.pl) from the Linux kernel.

### DIFF
--- a/examples/fileCompress.c
+++ b/examples/fileCompress.c
@@ -1,7 +1,7 @@
 /* LZ4file API example : compress a file
  * Modified from an example code by anjiahao
  *
- * This example will demonstrate how 
+ * This example will demonstrate how
  * to manipulate lz4 compressed files like
  * normal files */
 
--- a/lib/lz4frame.c
+++ b/lib/lz4frame.c
@@ -314,9 +314,9 @@ static LZ4F_errorCode_t LZ4F_returnErrorCode(LZ4F_errorCodes code)
 
 #define RETURN_ERROR(e) return LZ4F_returnErrorCode(LZ4F_ERROR_ ## e)
 
-#define RETURN_ERROR_IF(c,e) if (c) RETURN_ERROR(e)
+#define RETURN_ERROR_IF(c,e) do { if (c) RETURN_ERROR(e); } while (0)
 
-#define FORWARD_IF_ERROR(r)  if (LZ4F_isError(r)) return (r)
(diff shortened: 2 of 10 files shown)


Cyan4973:
Thanks for the suggestions @DimitriPapadopoulos ,

one of the things I would like to keep unchanged
is the white space at the end of printed strings, 
typically just before the `\n`.

There are scenarios in which the printed lines overwrite an existing blob of characters,
in which cases, the additional space character is helpful to distinguish the message from the noisy text background.
This is a property I would like to preserve.

DimitriPapadopoulos:
I'll remove the relevant commit. However, I wonder whether all current occurrences of trailing whitespaces have to do with these scenarios, or whether new occurrences of trailing whitespaces should be added to fully cover such scenarios. I mean, are these occurrences fully under control? Don't get me wrong, there's no wrong answer here, I am asking for my own education.

DimitriPapadopoulos:
Fixed by removing that commit.

Cyan4973:
OK,
I would like to come back to this PR, and see if it can be merged.

The main issue here is that it doesn't pass CI tests.
More specifically, it doesn't pass Visual Studio tests in debug mode, which also trigger some additional flags and analyzer pass.

Since the issue is labelled `conditional expression is constant`, it probably doesn't like the `do { ... } while (0)` and other equivalent changes.

I understand the benefits of `do { ... } while (0)` compared to just `{ ... }`,
but this is minor and this code base doesn't expect them.
Meanwhile, Visual Studio compatibility is an important objective in itself.

So I guess passing CI tests (which is mandatory) can be fixed by removing the relevant macro definition changes.

t-mat:
Perhaps, we can introduce inline `__pragma` for msvc.  For example, https://stackoverflow.com/a/4816375

DimitriPapadopoulos:
~~I agree `{ ... }` instead of `do { ... } while (0)` is good enough, especially outside public APIs.~~

DimitriPapadopoulos:
Actually, C4127 has been disabled in most source files:

https://github.com/lz4/lz4/blob/34d9127010f5c0265aad8c13baf718e27d4dd1f1/lib/lz4frame.c#L47-L49

I presume it has been forgotten in a couple files. I'll look into it.

DimitriPapadopoulos:
C4127 is now disabled in all files, and CI tests pass.

Cyan4973 (approved):

PR #1236, opened 2023-06-05
Add external user defined functions for memcpy and memmove

thomasverstringe:
To support reading dictionary through filestream, add external user defined functions for memcpy and memmove where a pointer to dictionary is used.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -359,6 +359,14 @@ typedef enum {
 #  endif
 #endif
 
+#if defined(LZ4_USER_DICT_FUNCTION)
+#   define LZ4_memcpy_dict(dst, src, size) LZ4_user_dict_memcpy(dst, src, size)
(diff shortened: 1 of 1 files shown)


thomasverstringe:
I wanted to add this feature for when decompressing on an embedded device with limited memory, where it is not possible to load a 100KB dictionary file into RAM.

If you open the dictionary as a filestream:
```
fs_open(&dict)
```
then define a mock pointer (pmock) that's just a random pointer to a char (cast easily to BYTE) and pass this to dctx->dict,
and have the user dictionary memcpy defined as:

```
void LZ4_user_dict_memcpy(void *dst, void *src, size_t size)
{
    int offset = (int)((void *)(pmock + dictSize) - src);

    fs_seek(&dict, 0 - offset, FS_SEEK_END);
    fs_read(&dict, dst, size);
}
```
This works well although it will be slower.


To elaborate, because of the limited memory on the embedded device, compression (pc side) is done by defining block sizes of 1024 or 2048 bytes so that no more than this needs to be allocated at a certain time.

PR #1241, opened 2023-06-14
contrib/gen_manual: Allow caller to set CXX_FOR_BUILD for gen_manual.

### LINKED ISSUE #1240: lz4 fails to cross-compile due to gen_manual rule
**Describe the bug**
lz4 uses $(CXX) to compile gen_manual, and then executes it during the build.

In a cross-build, say using a cross-compiler on an x86 system to build an lz4 that runs on aarch64, $(CXX) is a cross-compiler, so using it here yields an aarch64 executable for gen_manual, which can't be executed on x86.

**Expected behavior**
lz4 uses a native compiler for gen_manual when cross-compiling everything else.

For example, perhaps lz4 could sprout a parameter CXX_FOR_BUILD, which defaults to $(CXX), and which is used to compile gen_manual before executing it.

**To Reproduce**
Cross-compile lz4.

**Additional context**
Many other software packages that do something similar expose CC_FOR_BUILD or CXX_FOR_BUILD parameters, so there is plenty of precedent for this approach.

riastradh:
This way, in a cross-build of lz4, you can set CXX_FOR_BUILD to a native compiler that can produce a gen_manual executable to run at build-time.

fix https://github.com/lz4/lz4/issues/1240

### DIFF
--- a/contrib/gen_manual/Makefile
+++ b/contrib/gen_manual/Makefile
@@ -56,8 +56,9 @@ endif
 .PHONY: default
 default: gen_manual
 
+CXX_FOR_BUILD = $(CXX)
 gen_manual: gen_manual.cpp
-	$(CXX) $(FLAGS) $^ -o $@$(EXT)
+	$(CXX_FOR_BUILD) $(FLAGS) $^ -o $@$(EXT)
 
 $(LZ4MANUAL) : gen_manual $(LZ4API)
 	echo "Update lz4 manual in /doc"


contrib/gen_manual/Makefile
```diff
@@ -56,8 +56,9 @@ endif
 .PHONY: default
 default: gen_manual
 
+CXX_FOR_BUILD = $(CXX)
```
Cyan4973, line 59: I suspect what you meant here is : 
```
CXX_FOR_BUILD ?= $(CXX)
```

This way, `CXX_FOR_BUILD` won't be redefined if it's already set via an environment variable.

Cyan4973:
This sounds like a nice feature.
Now, a few questions : 
- Why `CXX_FOR_BUILD` ? as opposed to, for example, `CXX_FOR_MANUAL` ? or `CXX_FOR_HOST` ?
- If the variable is not published, it will not be known, hence it will not be used. This is fine for semi-private variables that have no intention to be useful outside of a very specific private scenario, but otherwise, it should be documented somewhere. `https://github.com/lz4/lz4/blob/dev/contrib/gen_manual/README.md` feels like the correct place for that.

Cyan4973:
I believe the better solution to this issue is to build the manual separately.

PR #1242, opened 2023-06-16
Don't clobber default WINDRES in MinGW environments

uckelman-sf:
`WINDRES` is already set when building with MinGW; setting it unconditionally to `windres` prevents the MinGW version of `windres` from being called.

This patch skips setting `WINDRES` for MinGW builds.

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -65,11 +65,7 @@ endif
 
 ifeq ($(WINBASED),yes)
 EXT        = .exe
-  ifneq (,$(filter MINGW%,$(TARGET_OS)))
-# WINDRES should already be set correctly for MinGW
-  else
-WINDRES    = windres
-  endif
+WINDRES ?= windres
 endif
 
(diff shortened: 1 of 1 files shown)


Makefile.inc
```diff
@@ -65,7 +65,11 @@ endif
 
 ifeq ($(WINBASED),yes)
 EXT        = .exe
+  ifneq (,$(filter MINGW%,$(TARGET_OS)))
+# WINDRES should already be set correctly for MinGW
+  else
 WINDRES    = windres
```
Cyan4973: If the goal is to not redefine `WINDRES` when it's already defined by environment,
what about : 
``` 
WINDRES ?= windres
```

uckelman-sf: Thanks for pointing this out. I was too narrowly focused on solving the problem at hand.

Cyan4973 (approved):

PR #1243, opened 2023-06-16
Add $(EXT) to Makefile targets for executables

uckelman-sf:
This is the current recipe for `lz4` in `lib/Makefile` used when `$(WINBASED)` is `yes`:
```
lz4: $(OBJFILES) lz4-exe.o
       $(CC) $(FLAGS) $^ -o $@$(EXT)
```
The command in the recipe writes a file named `$@$(EXT)`, which will be `lz4.exe`. The target of the recipe is for `lz4`, not `lz4.exe`. As `lz4` is never written when compiling for Windows, `make` never sees that it has a current `lz4.exe` and will always rebuild it. In particular, running `make` and then `make install` causes `lz4.exe` to be linked _twice_, not once like it ought to be. 

This PR appends `$(EXT)` to all of the all of the executable targets so that they match the files produced by their recipes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ lib lib-release liblz4.a:
 lz4 : liblz4.a
 lz4-release : lib-release
 lz4 lz4-release :
-	$(MAKE) -C $(PRGDIR) lz4$(EXT) lz4-release
+	$(MAKE) -C $(PRGDIR) lz4-nonrelease lz4-release
 	cp $(PRGDIR)/lz4$(EXT) .
 
 .PHONY: examples
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ lib lib-release liblz4.a:
 lz4 : liblz4.a
 lz4-release : lib-release
 lz4 lz4-release :
-	$(MAKE) -C $(PRGDIR) $@
+	$(MAKE) -C $(PRGDIR) lz4$(EXT) lz4-release
 	cp $(PRGDIR)/lz4$(EXT) .
 
 .PHONY: examples
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -47,12 +47,12 @@ build_script:
       make -v &&
       echo ----- &&
       if not [%PLATFORM%]==[clang] (
-        make -C programs lz4 &&
+        make -C programs lz4-nonrelease &&
         make -C tests fullbench &&
         make -C tests fuzzer &&
         make -C lib lib V=1
       ) ELSE (
-        make -C programs lz4 CC=clang MOREFLAGS="--target=x86_64-w64-mingw32 -Werror -Wconversion -Wno-sign-conversion" &&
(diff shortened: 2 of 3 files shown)


Cyan4973:
Could `lz4$(EXT)` be defined as a dependency of `lz4` target ?
This way, we wouldn't need to "lose" the `lz4` target, so it could still be invoked as usual, notably in CI,
and other recipes which depend on it wouldn't need to be updated, 
as `lz4$(EXT)` as a named dependency feels a little bit uglier than plain `lz4`
(it feels like an exposed implementation detail).

uckelman-sf:
> Could lz4$(EXT) be defined as a dependency of lz4 target ?

No, it can't. There's a specific reason for that: On platforms which aren't Windows, `$(EXT)` is the empty string, which means that `lz4` and `lz4$(EXT)` are same target.

Cyan4973:
> This is the current recipe for `lz4` in `lib/Makefile` used when `$(WINBASED)` is `yes`:
> 
> ```
> lz4: $(OBJFILES) lz4-exe.o
>        $(CC) $(FLAGS) $^ -o $@$(EXT)
> ```
> 
> The command in the recipe writes a file named `$@$(EXT)`, which will be `lz4.exe`. The target of the recipe is for `lz4`, not `lz4.exe`. As `lz4` is never written when compiling for Windows, `make` never sees that it has a current `lz4.exe` and will always rebuild it. In particular, running `make` and then `make install` causes `lz4.exe` to be linked _twice_, not once like it ought to be.
> 

Looking back on this topic, I'm trying to reproduce the same behavior on a local Windows system.
So far, I have failed.

For example, I have been testing the command `make lz4` under `mingw64`, `msys2`, and `cygwin64`. I could not observe the mentioned double-rebuild when calling again `make lz4` or `make` or even `make install`.

> `make` never sees that it has a current `lz4.exe` and will always rebuild it

Would you mind sharing the specifics of the environment where the reported effect can be observed ?

Furthermore, in all tested environments, I also observed that `$(EXT)` seems redundant nowadays, because the `Makefile` still properly generates `.exe` binaries such as `lz4.exe` when using the regular recipe for `make lz4` without any `$(EXT)`.

It could be that this topic has evolved since this `Makefile` was first written, and now there is no need to manually add these `$(EXT)` extension for Windows, the `.exe` suffix seems added automatically.
And if that is the case, then I think I would prefer this solution : just remove the Windows-specific `$(EXT)`, making the script easier to read and maintain.

Cyan4973:
At this point, I will prefer to remove the special handling of Windows executable suffix via `$(EXT)` variable,
on the ground that newer compilers now deal with it automatically,
thus reducing the associated cruft in `Makefile` with its complexity and side effects (such as the one described in this PR).

The exact timing for the removal of this variable is less obvious,
as it seems one needs a relatively recent compiler to benefit from this automatic `.exe` extension (> 2017),
so there might still be old active compilers around that wouldn't be able to handle `.exe` extension by themselves.

So this is a general direction, but now there's also a matter of right timing.

PR #1244, opened 2023-06-16
Don't conflate the shared library name with the shared library filename

uckelman-sf:
Currently, when building with MinGW, `$(LIBLZ4)` is `liblz4`. This is the correct _name_ for the library, but not the correct _filename_ for the DLL, which is `liblz4.dll`. This causes a problem with the following Makefile target:
```
 $(LIBLZ4): $(SRCFILES) liblz4-dll.o
 ifeq ($(BUILD_SHARED),yes)
        @echo compiling dynamic library $(LIBVER)
       $(CC) $(FLAGS) -DLZ4_DLL_EXPORT=1 -shared $^ -o dll/$@.dll -Wl,--out-implib,dll/$(LIBLZ4_EXP)
 endif
```
The recipe writes `dll/liblz4.dll` (and also `dll/liblz4.dll.a`), but does not write `liblz4`---which is the target. As a result, make cannot see that after running `make` that it has a fresh `dll/liblz4.dll` and thus will relink the library on runing `make` again (or `make install`).

Additionally, this target ends up being circular, due to `$(LIBLZ4)` being `liblz4`:
```
liblz4: $(LIBLZ4)
```

The cause of these problems is that current usage treats `LIBLZ4` as both the name of the shared library and the path of the shared library file, which for Windows builds are not the same thing. This PR fixes the problem by separating the name of the shared library out to `LIBLZ4_NAME`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,7 @@ ld.exe*
 # test artifacts
 *.lz4
 tmp*
+
+# generated Windows resource files
+lib/*.rc
+programs/*.rc
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -38,26 +38,26 @@ ifeq ($(TARGET_OS),)
 endif
 
 ifneq (,$(filter Windows%,$(TARGET_OS)))
-LIBLZ4 = liblz4-$(LIBVER_MAJOR)
-LIBLZ4_EXP = liblz4.lib
-WINBASED   = yes
+LIBLZ4_NAME = liblz4-$(LIBVER_MAJOR)
+LIBLZ4_EXP  = liblz4.lib
+WINBASED    = yes
 else
-LIBLZ4_EXP = liblz4.dll.a
+LIBLZ4_EXP  = liblz4.dll.a
   ifneq (,$(filter MINGW%,$(TARGET_OS)))
-LIBLZ4     = liblz4
-WINBASED   = yes
+LIBLZ4_NAME = liblz4
+WINBASED    = yes
   else
     ifneq (,$(filter MSYS%,$(TARGET_OS)))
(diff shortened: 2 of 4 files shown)


uckelman-sf:
This project has 1400+ lines of Makefiles, the bulk of which handle platform-specific things such as shared library and executable extensions. Bugs like the ones addressed by #1242, #1243, and this PR are what I see every time someone tries to handle platform differences manually in Makefiles.

Generating your Makefiles using autotools would eliminate bugs of this sort and leave you with dramatically less build code to maintain. I'd be happy to give you a PR for using autotools so you could see what difference it would make, if you'd like.

Cyan4973 (approved):

Cyan4973:
> This project has 1400+ lines of Makefiles, the bulk of which handle platform-specific things such as shared library and executable extensions. Bugs like the ones addressed by #1242, #1243, and this PR are what I see every time someone tries to handle platform differences manually in Makefiles.
> 
> Generating your Makefiles using autotools would eliminate bugs of this sort and leave you with dramatically less build code to maintain. I'd be happy to give you a PR for using autotools so you could see what difference it would make, if you'd like.

That sounds like a reasonable proposal.
Note though that there is a general lack of proficiency at using `autotools` on the maintainer side. At a minimum, there would be a ramp up period, and it's not clear at this stage how much effort must be spent into this direction in order to reach a "minimum threshold" that would make maintenance of autotools-created makefiles viable. And I suspect it's difficult to even just guess that without trying ...
In the meantime, it's important to not "lose" capabilities (i.e. not lose currently exposed targets). It's not clear how complex a first `autotool`-based recipe has to be in order to reach this target.

uckelman-sf:
I will make you a PR so you can see just how it will look.

PR #1253, opened 2023-07-19
Make hashes identical between LE and BE platforms

ltrk2:
My changes aim to resolve a difference in coding between little-endian and big-endian platforms in certain cases.

When invoked on the same data but on platforms of different endianness the `LZ4_hash4` function would produce different results owing to the multiplication performed in its body. I aim to resolve this by reading the contents of the binary as if they were represented in little-endian - which of course doesn't constrain the input data in any way; it's just for the sake of consistency with the ultimate goal of the aforementioned arithmetic operation producing the same numeric result and thus the same hash in the end.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ lz4
 
 # IDE / editors files
 .clang_complete
+.vscode
 _codelite/
 _codelite_lz4/
 bin/
--- a/build/meson/meson/meson.build
+++ b/build/meson/meson/meson.build
@@ -81,6 +81,15 @@ if get_option('memory-usage') > 0
   compile_args += '-DLZ4_MEMORY_USAGE=@0@'.format(get_option('memory-usage'))
 endif
 
+if get_option('endianness-independent-output')
+  if get_option('default_library') != 'static'
+    error('Endianness-independent output can only be enabled in static builds')
+  endif
+
+  add_project_arguments('-DLZ4_STATIC_LINKING_ONLY_ENDIANNESS_INDEPENDENT_OUTPUT')
+  compile_args += '-DLZ4_STATIC_LINKING_ONLY_ENDIANNESS_INDEPENDENT_OUTPUT'
+endif
+
 if get_option('unstable')
   add_project_arguments('-DLZ4_STATIC_LINKING_ONLY', language: 'c')
   compile_args += '-DLZ4_STATIC_LINKING_ONLY'
--- a/build/meson/meson_options.txt
+++ b/build/meson/meson_options.txt
@@ -18,6 +18,8 @@ option('disable-memory-allocation', type: 'boolean', value: false,
   description: 'See LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION. Static builds only')
 option('distance-max', type: 'integer', min: 0, max: 65535, value: 65535,
   description: 'See LZ4_DISTANCE_MAX')
+option('endianness-independent-output', type: 'boolean', value: false,
+  description: 'See LZ4_STATIC_LINKING_ONLY_ENDIANNESS_INDEPENDENT_OUTPUT. Static builds only')
 option('examples', type: 'boolean', value: false,
   description: 'Enable examples')
 option('fast-dec-loop', type: 'feature', value: 'auto',
--- a/lib/README.md
+++ b/lib/README.md
@@ -111,7 +111,8 @@ The following build macro can be selected to adjust source code behavior at comp
 - `LZ4_STATIC_LINKING_ONLY_ENDIANNESS_INDEPENDENT_OUTPUT` : experimental feature aimed at producing the same
   compressed output on platforms of different endianness (i.e. little-endian and big-endian).
   Output on little-endian platforms shall remain unchanged, while big-endian platforms will start producing
-  the same output as little-endian. This isn't expected to impact backward- and forward-compatibility in any way.
+  the same output as little-endian ones. This isn't expected to impact backward- and forward-compatibility
+  in any way.
 
 - `LZ4_FREESTANDING` : by setting this build macro to 1,
   LZ4/HC removes dependencies on the C standard library,
--- a/lib/README.md
+++ b/lib/README.md
@@ -108,6 +108,11 @@ The following build macro can be selected to adjust source code behavior at comp
   Remove support of dynamic memory allocation.
   For more details, see description of this macro in `lib/lz4.c`.
 
+- `LZ4_STATIC_LINKING_ONLY_ENDIANNESS_INDEPENDENT_OUTPUT` : experimental feature aimed at producing the same
+  compressed output on platforms of different endianness (i.e. little-endian and big-endian).
+  Output on little-endian platforms shall remain unchanged, while big-endian platforms will start producing
+  the same output as little-endian. This isn't expected to impact backward- and forward-compatibility in any way.
+
 - `LZ4_FREESTANDING` : by setting this build macro to 1,
   LZ4/HC removes dependencies on the C standard library,
   including allocation functions and `memmove()`, `memcpy()`, and `memset()`.
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -431,6 +431,16 @@ static U16 LZ4_readLE16(const void* memPtr)
     }
 }
 
+static U32 LZ4_readLE32(const void* memPtr)
+{
+    if (LZ4_isLittleEndian()) {
+        return LZ4_read32(memPtr);
+    } else {
+        const BYTE* p = (const BYTE*)memPtr;
+        return (U32)p[0] + (p[1]<<8) + (p[2]<<16) + (p[3]<<24);
+    }
+}
+
 static void LZ4_writeLE16(void* memPtr, U16 value)
 {
     if (LZ4_isLittleEndian()) {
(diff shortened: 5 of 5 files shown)


ltrk2:
@Cyan4973 do you think this could make it into the `dev` branch? I do see a failure but I believe it needs only be rerun because it seems to have gotten stuck during some environment setup phase.

t-mat:
Hi @ltrk2 ,

As you know, `liblz4` guarantees forward/backward litte-/big-endian compatibility.

But, even between with little endian platforms, `liblz4` doesn't give any guarantee for "identical" output.

Please check it with the following commands:

```
cd
mkdir lz4-pr-1253
cd    lz4-pr-1253
git clone --depth 1 --branch r129 https://github.com/lz4/lz4.git lz4-r129
cd lz4-r129

make -C programs lz4
make -C programs lz4c32

./programs/lz4    -1 README.md README.md.lz4
./programs/lz4c32 -1 README.md README.md.lz4c32

ls -l README.md.lz4*

./programs/lz4c32 -d README.md.lz4 output-from-lz4c32
./programs/lz4    -d README.md.lz4c32 output-from-lz4

diff output-from-lz4c32 output-from-lz4
```

Could you explain your aim just a bit?  For example, do you want to prevent possible future optimization for this identical output feature?

ltrk2:
@t-mat thanks for reaching out! Let me play with this a little and get back to you.

ltrk2:
@t-mat I see what you're trying to show me here: even when compiled on the same machine, 64-bit and 32-bit versions of the application produce different outputs, and even though those outputs are different they can be successfully decompressed with other versions.

I'll seek to explain below the use-case I'm faced with so that you can gain some insight into where the motivation for my proposal is coming from.

#### Context
I work on the [ClickHouse](https://github.com/ClickHouse/ClickHouse) open-source project, a column-oriented database management system. My current focus is porting it to [IBM Z](https://en.wikipedia.org/wiki/IBM_Z), a 64-bit big-endian processor architecture.

ClickHouse supports [compressing columns using LZ4](https://clickhouse.com/docs/en/sql-reference/statements/create/table#general-purpose-codecs) using `liblz4`.

Example for [uncompressed](https://fiddle.clickhouse.com/d48ff9e8-bde1-43b3-9bd4-83b7f72a0cef), and [LZ4 compressed](https://fiddle.clickhouse.com/93a37bd6-92ee-437d-aa6b-8ebb5cc77b65) columns. Notice the difference in size and that only the hash of files being compressed is different.

#### Problem
ClickHouse uses checksums - similar to the hashes seen in the example queries - to verify the integrity of stored artifacts. Since compression using `liblz4` produces a different output on IBM Z than it does on x86-64, the checksums don't match and thus the resulting artifacts are not transferable between the two platforms.

#### Proposal
I propose that some means to achieve equivalent output on different platforms be provided on a best-effort basis. If deemed necessary for the sake of backward- and forward-compatibility, locking this potentially experimental feature behind a feature flag would be completely acceptable.

#### Current Patch
I introduced an `LZ4_readLE32` function to perform reads in an endianness-independent manner in certain cases, which follows a precedent set by `LZ4_readLE16`. This allows for equivalent compressed binaries to be created under x86-64 (64-bit, little-endian) and IBM Z (64-bit, big-endian) so that interoperability in terms of artifacts produced on different platforms becomes feasible in my use-case.

Since as a result of these changes IBM Z ends up producing the same output as x86-64 would, I imagine that this makes it backward- and forward-compatible by default according to the example you showed me. Because of this, I believe that it could be introduced as is since it wouldn't break compatibility. I'll make sure to do some testing around this and get back to you.

#### Having gained some insight into the use-case I'm after, could you please share your thoughts on the matter?

Cyan4973:
It would certainly be possible to add a feature flag to support this.

The next question will probably be which mode should be the default one on big-endian platforms,
but I believe this question is of lesser importance for you.
As long as there is a way to support the behavior you want,
you can decide to employ it in your specific setup.

t-mat:
@ltrk2, thank you for the explanation.

In general, I think it's doable and nice to have a feature which is suitable for your situation.


> locking this potentially experimental feature behind a feature flag would be completely acceptable.

If your application allows you to use `liblz4` as a static library, it's nice to surround your feature with new `LZ4_STATIC_LINKING_ONLY_*` build macro.

We already have a simple macro [`LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION`](https://github.com/lz4/lz4/blob/v1.9.4/lib/lz4.h#L328-L331).  you can refer it as a template.


> I'll make sure to do some testing around this and get back to you.

I think this feature will be fragile without test.  Because majority of contributors can not care about it.
So, if you can share your test method with us, it may help us a lot.

But please note that I don't say that CI test code must be a part of this PR.

At least it's marked as experimental and `LZ4_STATIC_LINKING_ONLY_*` feature, we can test it by hand for a while.

I suppose that we'll use [QEMU](https://github.com/lz4/lz4/blob/v1.9.4/.github/workflows/ci.yml#L515) and gcc cross compiler for it.  But please let us know if you have a better idea/method for it.

t-mat:
> The next question will probably be which mode should be the default one on big-endian platforms,

To decide the default, we need actual benchmarks on different big-endian platforms from embedded to server class ones.

But so far, since we don't have much data and idea about the default, I think we can leave it as is, disabled.

ltrk2:
@t-mat @Cyan4973 thank you for the insight. I shall seek to implement this with `LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION` in mind.

In the meantime, please find below a test scenario similar to an earlier example. Please note that it employs cross-compilation to produce S390X (IBM Z) binaries and relies on QEMU for emulation on AMD64 hardware. I tested it on Ubuntu 22.04.

```
# this should cover the list of dependencies required to cross-compile for s390x and perform emulation using QEMU
apt-get install binutils-s390x-linux-gnu libc6-dev-s390x-cross gcc-s390x-linux-gnu binfmt-support qemu-user-static

git clone https://github.com/lz4/lz4.git
cd lz4

cp -r programs programs-s390x-dev
cp -r programs programs-amd64-dev

CC=gcc make -C programs-amd64-dev lz4
make clean

CC=s390x-linux-gnu-gcc make -C programs-s390x-dev lz4
make clean

file programs-amd64-dev/lz4
file programs-s390x-dev/lz4

programs-amd64-dev/lz4 -1 README.md README.md.lz4-amd64-dev
qemu-s390x-static -L /usr/s390x-linux-gnu programs-s390x-dev/lz4 -1 README.md README.md.lz4-s390x-dev

# note that the output sizes are different
ls -l | grep README.md.

programs-amd64-dev/lz4 -d README.md.lz4-s390x-dev out-1
qemu-s390x-static -L /usr/s390x-linux-gnu programs-s390x-dev/lz4 -d README.md.lz4-amd64-dev out-2

# because of the compatibility that was mentioned we still arrive at the same decompressed result
cmp out-1 out-2

# we're going to check out my branch
git remote add ltrk2 https://github.com/ltrk2/lz4.git

cp -r programs programs-s390x-ltrk2
cp -r programs programs-amd64-ltrk2

CC=gcc make -C programs-amd64-ltrk2 lz4
make clean

CC=s390x-linux-gnu-gcc make -C programs-s390x-ltrk2 lz4
make clean

file programs-amd64-ltrk2/lz4
file programs-s390x-ltrk2/lz4

programs-amd64-ltrk2/lz4 -1 README.md README.md.lz4-amd64-ltrk2
qemu-s390x-static -L /usr/s390x-linux-gnu programs-s390x-ltrk2/lz4 -1 README.md README.md.lz4-s390x-ltrk2

# note that output sizes are identical
ls -l | grep README.md.

# note that output from amd64 and s390x is now the same, and amd64 doesn't change
cmp README.md.lz4-amd64-dev README.md.lz4-amd64-ltrk2
cmp README.md.lz4-amd64-dev README.md.lz4-s390x-ltrk2

programs-amd64-ltrk2/lz4 -d README.md.lz4-s390x-ltrk2 out-3
qemu-s390x-static -L /usr/s390x-linux-gnu programs-s390x-ltrk2/lz4 -d README.md.lz4-amd64-ltrk2 out-4

# this is just a double-check, but for reasons of compatibility this is expected
cmp out-1 out-3
cmp out-1 out-4
```

ltrk2:
I implemented a feature flag called `LZ4_STATIC_LINKING_ONLY_ENDIANNESS_INDEPENDENT_OUTPUT` along the lines of what we discussed. @t-mat @Cyan4973 could you please review?

I tested it locally using the following scenario (the environment setup part from https://github.com/lz4/lz4/pull/1253#issuecomment-1677939136 might be necessary):
```
cp -r programs programs-s390x-ltrk2
cp -r programs programs-amd64-ltrk2

CC=gcc CFLAGS=-DLZ4_STATIC_LINKING_ONLY_ENDIANNESS_INDEPENDENT_OUTPUT make -C programs-amd64-ltrk2 lz4
make clean

CC=s390x-linux-gnu-gcc CFLAGS=-DLZ4_STATIC_LINKING_ONLY_ENDIANNESS_INDEPENDENT_OUTPUT make -C programs-s390x-ltrk2 lz4
make clean

programs-amd64-ltrk2/lz4 -1 README.md README.md.lz4-amd64-ltrk2
qemu-s390x-static -L /usr/s390x-linux-gnu programs-s390x-ltrk2/lz4 -1 README.md README.md.lz4-s390x-ltrk2

# note that the output matches
cmp README.md.lz4-amd64-ltrk2 README.md.lz4-s390x-ltrk2
```

In case the changes were to your liking, I'd appreciate it if we could get it merged. I'll make sure to follow up with tests afterwards in another PR based on the scenarios outlined above.

t-mat:
`LZ4_STATIC_LINKING_ONLY_ENDIANNESS_INDEPENDENT_OUTPUT` itself looks good :+1:

We may need some tweaks for other build systems, etc.  But we will treat them in other PRs.

ltrk2:
@Cyan4973 without any intention of rushing you or any checks you might wish to do, would it be possible to get this merged so that I can officially incorporate the functionality into ClickHouse?

Cyan4973:
I'm reviewing this PR and finds it generally well done.

The name of the proposed macro `LZ4_STATIC_LINKING_ONLY_ENDIANNESS_INDEPENDENT_OUTPUT` is quite a mouthful, 
but I get that it underlines its experimental nature, and follows the precedent established by `DISABLE_MEMORY_ALLOCATION`.

I agree with `@t-mat` comment that a CI test is really the only thing that will prevent this feature from being broken in the future. I'm also fine with having these tests provided as part of an additional separate PR.

Cyan4973 (approved):

ltrk2:
@Cyan4973 thank you very much for reviewing my changes. Please kindly go ahead with the merge. I shall follow up with tests in the new year.

PR #1254, opened 2023-07-31
Allow the usage of external xxHash

Ohjurot:
When building with CMake, lz4 compiles its own version of xxHash. This is not ideal since it disallows the simultaneous use of xxHash and lz4 in the same project without modifications in the build.

I'm using Conan as a package manager, and as soon as I use xxHash and lz4, I get linker issues because of multiple defined symbols (xxHash functions in lz4 and xxHash). This PR is the first step fixing this issue. I have added a CMake variable `LZ4_EXTERNAL_XXHASH` (by default OFF, thus not altering existing builds). When ON the variable will prevent compiling of the xxHash source and add xxHash as a CMake dependency and target link.

As soon as this is merged and included in an official release, I can go forward and refactor the Conan recipe, enable the variable during build, and add xxHash as a Conan dependency to lz4. Then xxHash and lz4 can coexist in Conan.

Assuming it's OK for you.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ _codelite_lz4/
 bin/
 *.zip
 *.swp
+compile_flags.txt
 
 # analyzers
 infer-out
--- a/programs/lz4cli.c
+++ b/programs/lz4cli.c
@@ -347,7 +347,8 @@ int main(int argc, const char** argv)
     /* Init */
     if (inFileNames==NULL) {
         DISPLAY("Allocation error : not enough memory \n");
-        return 1;
+        operationResult = 1;
+        goto _cleanup;
     }
     inFileNames[0] = stdinmark;
     LZ4IO_setOverwrite(prefs, 0);


t-mat:
Hi @Ohjurot ,

Nice catch!  But please consider to use standard `XXH_NAMESPACE`.

Because `lz4frame.c` uses it with `XXH_STATIC_LINKING_ONLY` mode.

https://github.com/lz4/lz4/blob/839edadd09bd653b7e9237a2fbf405d9e8bfc14e/lib/lz4frame.c#L74-L75

You can also refer other build scripts:

- https://github.com/lz4/lz4/blob/dev/lib/Makefile#L49
- https://github.com/lz4/lz4/blob/dev/programs/Makefile#L49
- https://github.com/lz4/lz4/blob/dev/ossfuzz/Makefile#L36
- https://github.com/lz4/lz4/blob/dev/tests/Makefile#L43
- https://github.com/lz4/lz4/blob/839edadd09bd653b7e9237a2fbf405d9e8bfc14e/build/meson/meson/meson.build#L18

Issue #880, opened 2020-06-20
Support mozilla/firefox non-standard lz4 header ("mozlz4")


afcady:
Mozilla & firefox starting using lz4 to compress its automatically-generated JSON files (used for configuration, history, bookmarks, etc.) before the lz4 on-disk format was standardized.  They invented their own disk format, now known as "mozlz4," which as far as I understand, is just lz4 but with a slightly different header.

Specifically, Mozlz4 uses the first 8 fixed bytes as a magic number, and then it uses 4 bytes to specify the uncompressed data length to be passed to `LZ4_decompress_safe` as the fourth argument.  I think the differences are that the standard lz4 format uses a different magic value, and a variable-width length field.

It's easy to decompress and compress mozlz4 files with the lz4 project's *libraries* -- several tools are available to do this -- but the lz4 project's command-line tool `lz4` does not recognize or produce the mozlz4 header.  So a custom tool is (currently) needed.

Here's an example tool that does both mozlz4 input and output using the lz4 libraries:

* https://github.com/jusw85/mozlz4/blob/master/src/main.rs

Mozilla's use of anything basically makes it kind of standard (probably lz4 is on more disks in the Mozlz4 format than in the standard format) and so there should be a standard, high-quality, widely-recognized, widely-distributed tool (such as lz4) to handle such a format.  There is a lot of data in that format (and it's valuable).

By the way, [Mozilla is not going to change their format ever](https://bugzilla.mozilla.org/show_bug.cgi?id=1209390).  That bug thread, among others, shows that it's not worth trying to get Mozilla to listen, IMO.  But it's anyway desirable to make existing data available even if new data is migrated to another format.

Cyan4973:
Given the presence of a magic number, transition from one format to another should be possible in a relatively graceful manner (support decoding of both formats, then switch to new format). It's strange this effort wasn't even started at some point, since it does not seem complex to do. I also don't see where it's specified that Mozilla will never update this format, but can understand that it's not happening due to lack of incentive.

It feels a bit backward to push that support requirement upwards, onto a non-participating party. To be fair, support for this custom format does not seem terribly complex, but comes at the cost of expected properties, which makes it more difficult to combine : 
- The format is non-streamable : the entire input must be loaded into memory before starting compression
- The format is non-concatenable : there is no "end" signal or distance, the format discovers its end because the input stream is completed. It makes it more difficult to properly size buffers in advance. It's also impossible to concatenate another frame after this one.

[...]

PR #1259, opened 2023-08-26
Recognize and decompress files created by Mozilla

UnitedMarsupials:
This aims to address the decompression part of the lz4/lz4#880.

The compression part -- _creation_ of files usable by Mozilla -- may be more contentious, given the reluctance to support the proliferation of the bogus formats.

(Personally, I need to be able to (sometimes) decompress Mozilla files, modify them, and then compress the modified data back for Firefox and/or Thunderbird to process.)

### DIFF
--- a/programs/lz4io.c
+++ b/programs/lz4io.c
@@ -1031,54 +1031,50 @@ LZ4IO_decodeMozilla(FILE* finput, FILE* foutput, const LZ4IO_prefs_t* prefs)
     char *out_buff;
 #ifdef __unix__
     int mmapped;
-#endif
 
-    {   size_t const sizeCheck = fread(&outputSize, 1, 4, finput);
-        if (sizeCheck != 4) END_PROCESS(73, "Read error: %s", strerror(errno));
-        outputSize = LZ4IO_readLE32(&outputSize);
-    }
     {   struct stat sb;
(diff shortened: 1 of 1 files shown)


Cyan4973:
This is good work,
but I'm not convinced that this repository is the best place to host it.
I would rather see it as a dedicated fork, or tool, focused on its mission to decode (and eventually produce) mozilla format.

UnitedMarsupials:
> I would rather see it as a dedicated fork

What good is a fork at all?!

> or tool, focused on its mission to decode (and eventually produce) mozilla format.

The existing `lz4` already supports multiple different formats -- recognizing a multitude of "magic" numbers. If _those_ aren't all split into separate tools of their own, why "pick" on the Mozilla's "flavor"?

Cyan4973:
If your concern is discoverability, a reasonable trade-off I can propose is to host a dedicated lz4-mozilla reader (and maybe generator) as part of the `contrib/` directory, within this repository.

UnitedMarsupials:
> host a dedicated lz4-mozilla reader

Such readers already exist :) I even maintain [one of them for FreeBSD](https://www.freshports.org/archivers/lz4json/). I deliberately made this change to match the style and the spirit of this repository...

I still don't understand, why you hesitate to add one more "flavor".

PR #1260, opened 2023-08-27
Move GNUInstallDirs include before it is referenced first

laszlo-dobcsanyi:
The lz4_shared and lz4_static targets specify their install include directories with the CMAKE_INSTALL_INCLUDEDIR variable that should be coming from include(GNUInstallDirs), however that file is only included later.

I think this results in an empty/missing INTERFACE_INCLUDE_DIRECTORIES property in the installed configuration files, making CMake not pick up the correct include directories for the installation.

This becomes an issue for example when cross compiling with a sysroot that does not has lz4 headers in standard include paths and the explicitly installed headers are not found either due to the empty INTERFACE_INCLUDE_DIRECTORIES property.

Please let me know if I'm missing something or you need more details

@nemequ

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -677,12 +677,10 @@ jobs:
 
     - name: cmake
       run: |
-        cd build/cmake
-        mkdir build
-        cd build
-        cmake ..
-        CFLAGS=-Werror make VERBOSE=1
-
+        CFLAGS=-Werror cmake -S build/cmake -B build -D CMAKE_INSTALL_PREFIX=install
+        VERBOSE=1 cmake --build build --target install
+        cmake -S tests/cmake -B build_test -D CMAKE_INSTALL_PREFIX=install
+        VERBOSE=1 cmake --build build_test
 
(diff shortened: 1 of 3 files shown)


Cyan4973:
Would there be a way to test the mentioned scenario in CI ?

This would help justify the change,
even more importantly, it would help keeping it working in the future,
since future contributors may accidentally break untested features.

laszlo-dobcsanyi:
I think it is certainly possible to verify that the installed imported target is what is expected.

I see there is a cmake Github workflow that right now only builds with CMake and I could extend it to also install into a prefix, so the CMake installation logic would also be tested.
Then a sort of integration test in a form of another CMake file (possibly in the tests directory?) can be added that would try to call find_package(lz4 REQUIRED) and could verify that the imported library files and include directories are correct.
This test CMake script could be invoked in the cmake workflow.

Do you think this is a good way forward?

Cyan4973:
Yes, it's a good plan

Cyan4973 (approved):

PR #1263, opened 2023-09-06
lz4hc: increase count back search step

JunHe77:
Current LZ4HC_countBack search for common bytes
at a step of byte which is not efficient. Replace
this with a step of 4 bytes.

Change-Id: I3ba1287ebc2ea06bb65f4700fa454c506d4b0cfc

### DIFF
--- a/lib/lz4hc.c
+++ b/lib/lz4hc.c
@@ -143,6 +143,41 @@ LZ4_FORCE_INLINE void LZ4HC_Insert (LZ4HC_CCtx_internal* hc4, const BYTE* ip)
     hc4->nextToUpdate = target;
 }
 
+LZ4_FORCE_INLINE
+unsigned LZ4HC_NbCommonBytes32(U32 val)
+{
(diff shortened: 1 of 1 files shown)


JunHe77:
This patch improves compression performance of L3 - L9 from ~6% - 10% at various level.

Cyan4973:
Can you provide some benchmark results ?
Which corpus, platform and compiler have you been using ?

Cyan4973 (approved):
I could check the performance on 2 platforms (`i7-9700k` and `M1 Pro`) and 2 compilers (`gcc` and `clang`) and 3 corpuses.
In most cases, I observed speed improvements for levels 3-9 (impacted by this PR),
though they sometimes can be small, especially at lower levels.
I even have a few isolated events of minor regressions at level 3, but nothing that would change the overall picture.
In all cases, the produced compressed artifacts remain identical, so it's purely a speed update.

I think this is a rather beneficial PR, and it feels worth merging.

PR #1277, opened 2023-09-26
fix examples

### LINKED ISSUE #1276: The statistics for the example of compress_functions are incorrect!
**Describe the bug**
image
if time_taken__default > 2147483647, the result of `(int)time_taken__default       / iterations` maybe a negative number.
It'll be better to change it to `(int)(time_taken__default       / iterations)`

Cyan4973:
streamlined `Makefile`.
avoid usage of `-lrt` (problematic for portability). 
now `bench_functions` (renamed) can be part of `make test`.

fix #1276


Cyan4973:
So, as expected, all `x32` tests are failing.

But I'm not sure how to fix that.

Nor if it should be fixed: is that a temporary mishap from Github ? Or is `x32` support permanently dropped ? ( I haven't found any source on the last one)

Note: it seems that `20.04` is still working fine for `x32` support

Cyan4973:
Rebased on top of `dev` fixes for `x32` tests

PR #1285, opened 2023-10-09
chore: suppress warning C6385 for MSVC 17.7

t-mat:
MSVC 17.7 has been released [1].  Their new `_MSC_VER` is `1937`.
But we're still hitting same false positive warnings.

[1]: https://devblogs.microsoft.com/visualstudio/visual-studio-2022-17-7-now-available/


Cyan4973:
It's a weird warning.
It seems that, if it was true, we would hit this condition with `asan` or `msan` enabled.
I suppose their analyzer is confused and believes there's only 2 bytes ready for read at this address ?

Cyan4973 (approved):

t-mat:
yeah it's really strange.  I suspect their analyzer doesn't understand how `memcpy()` works at all.

PR #1287, opened 2023-10-09
Introduce the `.clang-format` rule file 

localspook:
This PR introduces [clang-format](https://clang.llvm.org/docs/ClangFormat.html) into the codebase.

## Motivation

Currently, the codebase is, to put it lightly, inconsistent. There is variation even in such basic things as the number of spaces with which we indent. Consistent code is easier to read and understand, and with clang-format we can ensure consistency across the whole codebase. It makes contributing to the project easier too; people don't need to read any style guide or try and figure out the conventions we use; they can write code however they want and run clang-format on it before they submit a PR.

## Key changes

My starting settings are quite aggressive. I'm happy to make the formatter more lenient or discuss specific settings. I strongly believe that even a very lenient formatter is better than no formatter at all.

- A for loop, while loop, or case label and block on one line is disallowed:

```c
// before
while (true) bar();
for (int i = 0; i < 5; ++i) { baz(); fiz(); }
case 1: biz();

// after
while (true) {
    bar();
}
for (int i = 0; i < 5; ++i) { 
    baz(); 
    fiz(); 
}
case 1: 
    biz();
```

- Consecutive declarations are always aligned:

```c
// before
LZ4_stream_t* const lz4Stream = LZ4_createStream();
const size_t cmpBufBytes = LZ4_COMPRESSBOUND(messageMaxBytes);
char* const cmpBuf = (char*) malloc(cmpBufBytes);
char* const inpBuf = (char*) malloc(ringBufferBytes);
int inpOffset = 0;

// after
LZ4_stream_t* const lz4Stream   = LZ4_createStream();
const size_t        cmpBufBytes = LZ4_COMPRESSBOUND(messageMaxBytes);
char* const         cmpBuf      = (char*)malloc(cmpBufBytes);
char* const         inpBuf      = (char*)malloc(ringBufferBytes);
int                 inpOffset   = 0;
```
- Consistent handling of functions with long argument/parameter lists (clang-format calls this "binpacking"):
```c
// before
static void roundTripTest(void* resultBuff, size_t resultBuffCapacity,
                          void* compressedBuff, size_t compressedBuffCapacity,
                    const void* srcBuff, size_t srcSize)

// after
static void roundTripTest(void*       resultBuff,
                          size_t      resultBuffCapacity,
                          void*       compressedBuff,
                          size_t      compressedBuffCapacity,
                          const void* srcBuff,
                          size_t      srcSize)
```

## Possible improvements

- Integrate clang-format with GitHub (I'm not familiar with that so it's not part of this PR).

---

I realize this may be an overwhelming PR. Please keep in mind that these are changes to *style*, not functionality, and so not every single line deserves close scrutiny. Thank you very much for your time.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,132 @@
+# This is the configuration file for clang-format, an automatic code formatter.
+# Introduction: https://clang.llvm.org/docs/ClangFormat.html
+# Supported options: https://clang.llvm.org/docs/ClangFormatStyleOptions.html
+
+Language: Cpp
+Standard: Latest
+
+ColumnLimit: 110
+
+UseTab: Never
+IndentWidth: 4
+PPIndentWidth: 2
+ContinuationIndentWidth: 4
+
+LineEnding: LF
+InsertNewlineAtEOF: true
+KeepEmptyLinesAtTheStartOfBlocks: false
+MaxEmptyLinesToKeep: 1
+
+IndentCaseBlocks: false
+IndentCaseLabels: false
+IndentGotoLabels: false
+IndentPPDirectives: AfterHash
+IndentWrappedFunctionNames: false
+
+AlignAfterOpenBracket: Align
+AlignArrayOfStructures: Right
+AlignEscapedNewlines: Left
+AlignOperands: Align
+AlignConsecutiveAssignments:
+    Enabled:          true
+    AcrossEmptyLines: false
+    AcrossComments:   false
+    AlignCompound:    false
+    PadOperators:     false
+AlignConsecutiveBitFields:
+    Enabled:          true
+    AcrossEmptyLines: false
+    AcrossComments:   false
+AlignConsecutiveDeclarations:
+    Enabled:          true
+    AcrossEmptyLines: false
+    AcrossComments:   false
+AlignConsecutiveMacros:
+    Enabled:          true
+    AcrossEmptyLines: false
+    AcrossComments:   false
+AlignTrailingComments:
+    Kind:           Leave
+    OverEmptyLines: 0
+
+BinPackArguments: true
+BinPackParameters: false
+AllowAllArgumentsOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: false
+
+AllowShortBlocksOnASingleLine: Never
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortEnumsOnASingleLine: true
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: AllIfsAndElse
+AllowShortLoopsOnASingleLine: false
+
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+BraceWrapping:
+    AfterCaseLabel:        false
+    AfterControlStatement: Never
+    AfterEnum:             false
+    AfterExternBlock:      false
+    AfterFunction:         true
+    AfterStruct:           false
+    AfterUnion:            false
+    BeforeElse:            false
+    BeforeWhile:           false
+    IndentBraces:          false
+    SplitEmptyFunction:    false
+    SplitEmptyRecord:      false
+BreakAfterAttributes: Never
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Custom
+BreakBeforeInlineASMColon: OnlyMultiline
+BreakBeforeTernaryOperators: true
+
+DerivePointerAlignment: false
+PointerAlignment: Left
+
+QualifierAlignment: Custom
+QualifierOrder: ["inline", "static", "volatile", "restrict", "const", "type"]
+
+ReflowComments: false
+BreakStringLiterals: false
+RemoveSemicolon: true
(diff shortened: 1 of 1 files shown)


Cyan4973:
I'm at least favorable to the idea.

The most important part of the PR is the `.clang-format` rule file, so we should at least spend a bit of time talking about it.
The rest of the PR is a good illustration of the kind of outcome it should produce.
(I note that, btw, many tests are failing after this PR, which points at changes that are a bit more than cosmetic).

Now onto a more detailed feedback on proposed rules : 

> A condition or case label and block on one line is disallowed
> Omitting braces is disallowed

Well, short `if()` exist, and look good for brevity.
One of the most important asset when browsing code is vertical density, to ensure having as many elements as possible right under the eyes. So it shouldn't be wasted by introducing too many new lines for no good reasons.

I think I would agree with this rule for the `for(;;) { ... }` pattern.
By extension, it's likely wise to adopt the same rule for the `while() { ... }` pattern.
But short `if() ...;` exist, and I would prefer to favor brevity in this case.
Finally, regarding `case X:`, I'm definitely against the idea of _requiring_ brackets all the time.

> Consecutive declarations are always aligned

Yes, good idea

> Consistent handling of functions with long argument/parameter lists (clang-format calls this "binpacking")

Well, I tend to like the idea of having `Type* ptr, size_t size` on the same line,
to make it clear that both variables are actually related,
but I acknowledge that it might be a little too difficult to ask such a nuance for a parser.

Anyway, I suspect this is only for functions with a lot of parameters,
and some clarity would be welcomed for these few cases.

t-mat:
Hi @LocalSpook, thanks for sending PR.
It seems you're not familiar with our codebase.  And we may need to discuss many details. I think "small start" may be better strategy.
I mean starting to target with more smaller amount of code. For example, applying it to `examples/` only and aim to merge smaller change.  After that you can extend your target code easier.

Also just an FYI, you can locally invoke `make test` for basic testing.

localspook:
> But short if() ...; exist, and I would prefer to favor brevity in this case.

Changed `AllowShortIfStatementsOnASingleLine` to `AllIfsAndElse`.

> I'm definitely against the idea of requiring brackets all the time.

Changed `InsertBraces` to `false`.

> (I note that, btw, many tests are failing after this PR, which points at changes that are a bit more than cosmetic).

Yep, that turned out to be (I believe) due to some header files depending on include order, which broke when the includes were rearranged.

`programs/platform.h` has (effectively) this content:
```c
#define _POSIX_C_SOURCE
#include <stdio.h> // Brings in fileno() because _POSIX_C_SOURCE is defined
```

Then `programs/util.h` had:
```c
#include "platform.h"
#include <stdio.h>

fileno(/* ... */);
```
which got rearranged to 
```c
#include <stdio.h>
#include "platform.h"

fileno(/* ... */);
```
which is equivalent to 
```c
#include <stdio.h>
#define _POSIX_C_SOURCE
#include <stdio.h> // <stdio.h> was already included, so this does not bring in fileno()!

fileno(/* ... */); // Undefined symbol :(
```
I don't know if this should be fixed somehow; for now I've disabled include sorting to be safe.

> Finally, regarding `case X:`

I'm not sure I understand what you mean here.

> I think "small start" may be better strategy.

I'm not pushing hard to merge the actual reformatting in this PR; it's there to demonstrate the possible changes. If they look good, we can remove them from this PR and just merge the `.clang-format` file.

Cyan4973:
> Yep, that turned out to be (I believe) due to some header files depending on include order, which broke when the includes were rearranged.

It should probably be fixed,
but that should be the purpose of a different PR.

localspook:
I've updated the PR description to reflect the new changes and enabled removal of redundant parentheses.

Cyan4973:
OK,
I believe the part of this PR we could merge without too much difficulties are
the `.clang-format` file, and eventually the `.gitignore` one.

For the rest, I believe it would be easier to make them belong to their separate PR.

In particular, I note a few changes to several build scripts,
which seems separate from `.clang-format` scope.
The changes are likely good, but deserve a review of their own (one PR per build system for proper attention).

As for source code, I guess we could update it on the go, as the `.clang-format` rulebook get used.
It might also be worth creating some dedicated PR focusing on some of these files
(the `libzlz4` source code seems like a prime target for example).

localspook:
The build script changes are all conceptually one fix. Each of those build scripts parses `lz4.h` to get the version numbers, using a regex. The CMake regex failed when the whitespace was changed a little. The Makefile regexes happened to work, but are fragile in the same way. These changes should be merged before (or at the same time as) any reformatting, otherwise someone might get strange build failures when they run clang-format. Would you still like me to split them into a separate PR (or 2 PRs for each build system)?

Cyan4973:
> These changes should be merged before (or at the same time as) any reformatting, otherwise someone might get strange build failures when they run clang-format. 

Good point

> Would you still like me to split them into a separate PR (or 2 PRs for each build system)?

Yes please

localspook:
I've divided up the PR into separate PRs.

Cyan4973:
> I've divided up the PR into separate PRs.

Great !

Cyan4973 (approved):

PR #1290, opened 2023-10-12
Make Makefile version number parsing more robust

localspook:
As discussed in PR #1287.


Cyan4973 (approved):

Cyan4973:
minor comment:
it seems we are using this version extraction routine in many places,
to the point that I wonder if it would be better to have a single centralized version somewhere
that would be accessible by all these build systems.

This could be a future follow up.

PR #1295, opened 2023-10-13
Remove redundant null pointer checks before `free()`

localspook:
`free()` [already checks if it was given a null pointer](https://en.cppreference.com/w/c/memory/free#Notes), and therefore a user-defined `LZ4_free()` should too, so we don't need to check for null pointers ourselves.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1545,7 +1545,6 @@ void LZ4_resetStream_fast(LZ4_stream_t* ctx) {
 #if !defined(LZ4_STATIC_LINKING_ONLY_DISABLE_MEMORY_ALLOCATION)
 int LZ4_freeStream (LZ4_stream_t* LZ4_stream)
 {
-    if (!LZ4_stream) return 0;   /* support free on NULL */
     DEBUGLOG(5, "LZ4_freeStream %p", LZ4_stream);
     FREEMEM(LZ4_stream);
     return (0);
@@ -2497,7 +2496,6 @@ LZ4_streamDecode_t* LZ4_createStreamDecode(void)
 
 int LZ4_freeStreamDecode (LZ4_streamDecode_t* LZ4_stream)
 {
-    if (LZ4_stream == NULL) { return 0; }  /* support free on NULL */
     FREEMEM(LZ4_stream);
     return 0;
 }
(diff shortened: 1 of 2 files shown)


Cyan4973:
That's true of `free()`.
But `FREEMEM` makes it possible to replace `free()` by some user defined function.
In which case, the user-defined function would also have to support `NULL`.
Which, if you ask me, would be good practice, but 
it's different from saying that it has to be a mandatory part of the contract.

And I'm reluctant changing the contract now.
Who knows if there is some scenario depending on that out there.
Meanwhile, such a precaution doesn't hurt much.

localspook:
In `lz4hc.c`, in function `LZ4HC_compress_optimal()`, there's code that boils down to:

```c
LZ4HC_optimal_t* const opt = (LZ4HC_optimal_t*)ALLOC(sizeof(LZ4HC_optimal_t) * (LZ4_OPT_NUM + TRAILING_LITERALS));
if (opt == NULL) goto _return_label;

_return_label:
FREEMEM(opt);
```

Should that be considered a bug then?

Cyan4973:
I should look at the code to be sure,
but generally the pattern `goto _return_label;`
tends to work as a kind of generic "cleaning" terminating pattern,
and there might be several `goto` places leading to the same label.
So one would have to be sure that all of them don't need the following `FREEMEM(opt)` before removing it.

localspook:
The `FREEMEM()` does need to be there for other exit paths. I'm asking whether it's a bug that the code jumps to `_return_label`  when `opt` is a null pointer, and passes that null `opt` to `FREEMEM()`. So, I'm asking if it should be changed to something like:

```c
if (opt == NULL) return 0;
```

And if that is a bug and we're guaranteeing that the user-supplied free function will never be called with a null pointer, that should be documented somewhere.

Cyan4973:
That's a good point.
Yes, for consistency, the same rule regarding `FREEMEM()` should be applied everywhere.
Since the API doesn't make it compulsory (yet) for `FREEMEM()` to support `NULL`,
this scenario should be taken care of by the code, so an additional `if (opt)` test should be added.

localspook:
I opened PR #1297 to add that check. We could add a comment above the `LZ4_free()` declaration that says something like "Like `free()`, your custom `LZ4_free()` must be able to handle null pointers." It wouldn't break any current code, since we never actually give `LZ4_free()` null pointers, but it might encourage people to check the correctness of their `LZ4_free()`. I don't feel strongly about the idea though.

PR #1299, opened 2023-10-19
added new qemu targets for CI (MIPS, M68K, RISC-V)

Cyan4973:


Cyan4973:
Interesting,
I see the new `mips` and `m68k` tests being run,
but not `risc-v`.
What's going on ...

_edit_ : found it, small typo

PR #1300, opened 2023-10-20
fix incorrect AND of condition check

JunHe77:
Missing one & in pointers check makes this a bitwise AND instead of logical AND. This was introduced in commit 3c0332600490003b1267a9f0a51a6a315b52891e.

Change-Id: Ice50d240e3f52e38e2b3456dd2c5bc5cefb70e0d

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1085,7 +1085,7 @@ LZ4_FORCE_INLINE int LZ4_compress_generic_validated(
 
         /* Catch up */
         filledIp = ip;
-        while (((ip>anchor) & (match > lowLimit)) && (unlikely(ip[-1]==match[-1]))) { ip--; match--; }
+        while (((ip>anchor) && (match > lowLimit)) && (unlikely(ip[-1]==match[-1]))) { ip--; match--; }
 
         /* Encode Literals */
         {   unsigned const litLength = (unsigned)(ip - anchor);


Cyan4973:
It is correct, and it is intended to be a bitwise operation.

This is supposed to nudge the compiler towards replacing a branch by an `AND` operation.

t-mat:
I also supposed this code should be warned by `ctocpptest`.  I expected to see warnings something like: "implicit casting from `int` to `bool`".  But actually not.

```
echo "int main() { if(2 & 2) return 123; else return 4; }" > c.cpp

g++ -Wall -Wextra -Wconversion -Wbool-operation ./c.cpp && ./a.out
# no error / warning(s)
echo $?
# 123

clang++ -Wall -Wextra -Wconversion -Wbool-operation ./c.cpp && ./a.out
# no error / warning(s)
echo $?
# 123
```

`if(1 & 2) {}` is perfect code in both of C and C++.
I've been infected with many other languages which don't allow implicit cast from integer to boolean (Java, C#, etc) 🤦

Cyan4973:
`C90` doesn't even have `boolean` as a separate type.
It's just mapped onto integer, with `0` meaning `false`, and anything else meaning `true`.

Even `C99` booleans are just macros, still employing integer underneath.

JunHe77:
> It is correct, and it is intended to be a bitwise operation.
> 
> This is supposed to nudge the compiler towards replacing a branch by an `AND` operation.

Understood.
Maybe add this comment in code would help others understand it. :)

PR #1301, opened 2023-10-23
lz4: remove unnecessary check of ip

JunHe77:
ip has been advanced before the match finder loop, so (ip>anchor) will be always true before entering the loop.
This change generates better codegen with clang.

Change-Id: I2567cb7f4a16cc28496fc88c07dcd99601bc97b3

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1085,7 +1085,10 @@ LZ4_FORCE_INLINE int LZ4_compress_generic_validated(
 
         /* Catch up */
         filledIp = ip;
-        while (((ip>anchor) & (match > lowLimit)) && (unlikely(ip[-1]==match[-1]))) { ip--; match--; }
+        assert(ip > anchor); /* this is always true as ip has been advanced before entering the main loop */
+        if ((match > lowLimit) && unlikely(ip[-1] == match[-1])) {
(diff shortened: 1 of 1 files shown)


JunHe77:
With clang-16 and silesia, observed ~1.8 - 3% compression uplift for level-1/2 on Arm N1/N2, Apple M1 and Xeon 5218.
No obvious regression observed for gcc-13.

lib/lz4.c
```diff
@@ -1085,7 +1085,9 @@ LZ4_FORCE_INLINE int LZ4_compress_generic_validated(
 
         /* Catch up */
         filledIp = ip;
-        while (((ip>anchor) & (match > lowLimit)) && (unlikely(ip[-1]==match[-1]))) { ip--; match--; }
+        if ((match > lowLimit) && unlikely(ip[-1] == match[-1])) {
```
Cyan4973, line 1089: We should probably add an `assert(ip > anchor)` before this line.
I understand this condition is supposed to be correct, it's just not "obvious enough" on reading this line.

JunHe77: Thanks for that. Updated in the new version.

Cyan4973 (approved):

PR #1308, opened 2023-11-10
Add LZ4_compress_fast_extState_destSize() API

tristan957:
Allow a user to provide an external state, but also understand how much of the source buffer was compressed.

### DIFF
--- a/lib/lz4.h
+++ b/lib/lz4.h
@@ -243,6 +243,13 @@ LZ4LIB_API int LZ4_compress_fast (const char* src, char* dst, int srcSize, int d
 LZ4LIB_API int LZ4_sizeofState(void);
 LZ4LIB_API int LZ4_compress_fast_extState (void* state, const char* src, char* dst, int srcSize, int dstCapacity, int acceleration);
 
+/*! LZ4_compress_fast_extState_destSize() :
+ *  Same as LZ4_compress_fast(), but compresses as much data as possible from 'src' buffer into
+ *  already allocated buffer 'dst', of size >= 'dstCapacity'. This function either compresses the
+ *  entire 'src' content into 'dst' if it's large enough, or fills 'dst' buffer completely with as
+ *  much data as possible from 'src'.
+ */
+LZ4LIB_API int LZ4_compress_fast_extState_destSize(void* state, const char* src, char* dst, int *srcSizePtr, int dstCapacity, int acceleration);
 
 /*! LZ4_compress_destSize() :
  *  Reverse the logic : compresses as much data as possible from 'src' buffer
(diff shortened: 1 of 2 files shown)


tristan957:
I would like to make use of such an API in GLib in order to make an implementation of [`GConverter`](https://docs.gtk.org/gio/iface.Converter.html).

lib/lz4.h
```diff
@@ -243,6 +243,13 @@ LZ4LIB_API int LZ4_compress_fast (const char* src, char* dst, int srcSize, int d
 LZ4LIB_API int LZ4_sizeofState(void);
 LZ4LIB_API int LZ4_compress_fast_extState (void* state, const char* src, char* dst, int srcSize, int dstCapacity, int acceleration);
 
+/*! LZ4_compress_fast_extState() :
```
Cyan4973: You probably mean `LZ4_compress_fast_extState_destSize()` here ?

lib/lz4.h
```diff
@@ -243,6 +243,13 @@ LZ4LIB_API int LZ4_compress_fast (const char* src, char* dst, int srcSize, int d
...
+ *  Same as LZ4_compress_fast(), but compresses as much data as possible from 'src' buffer into
+ *  already allocated buffer 'dst', of size >= 'targetDestSize'. This function either compresses the
```
Cyan4973: There is no parameter `targetDestSize` in this function's signature.

tristan957:
@Cyan4973 did you have any more feedback? I'd like to get all my work done for the next GLib minor release in April-ish.

Cyan4973:
The proposed modifications look fine to me, from a correctness perspective.

I just need to ensure that the patch does not accidentally impact the performance of regular `LZ4_compress_fast_extState()`.

Cyan4973:
I ran a few performance tests on several different platforms using different compilers,
and I could not detect any significant performance for `LZ4_compress_fast_extState()`.

So as far as I'm concerned, this is good to go.

Cyan4973 (approved):

tristan957:
@Cyan4973 do you know when we might see this in a release?

Cyan4973:
There is no set date yet.
I would prudently consider something like Summer 2024,
but these things can change, in both directions, depending on priorities.

Cyan4973:
I've been testing the new entry-point `LZ4_compress_fast_extState_destSize()` exposed in this PR,
and from what I can tell so far, it doesn't seem to work as intended:

Faced with an output buffer smaller than would be required to store the entire compressed input,
it just fails, as a regular compression function, instead of being successful by compressing less input.

PR #1310, opened 2023-11-21
Update deprecation warning

tomerv:
No use in the deprecated warning pointing to another deprecated function - update to point directly to the recommended function.

### DIFF
--- a/doc/lz4_manual.html
+++ b/doc/lz4_manual.html
@@ -560,12 +560,12 @@ LZ4_DEPRECATED("use LZ4_compress_fast_continue() instead") LZ4LIB_API int LZ4_co
 LZ4_DEPRECATED("use LZ4_compress_fast_continue() instead") LZ4LIB_API int LZ4_compress_limitedOutput_continue  (LZ4_stream_t* LZ4_streamPtr, const char* source, char* dest, int inputSize, int maxOutputSize);
 </b><p></p></pre><BR>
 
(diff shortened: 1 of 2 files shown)


Cyan4973:
Problem is, these proposed alternatives are not interchangeable.

`LZ4_uncompress()` requires the _decompressed_ block size, like `LZ4_decompress_fast()`,
whereas `LZ4_decompress_safe()` requires the _compressed_ size.
Therefore, a use case having access to the decompressed size only cannot transition to the `_safe()` variant.

I haven't looked at it recently. There might be a way to employ the `_destSize()` variant to achieve the same functionality as `_uncompress()`, i.e. being able to decompress a LZ4 block while knowing only its decompressed size. If that is the case, it would be a proper update recommendation.

tomerv:
> Problem is, these proposed alternatives are not interchangeable.

The same is true for `LZ4_decompress_fast()` and `LZ4_decompress_safe()`, right? But the `LZ4_DEPRECATED` message for `LZ4_decompress_fast()` still recommends using `LZ4_decompress_safe()` - so it should at least be consistent.
https://github.com/lz4/lz4/blob/dev/lib/lz4.h#L841

Maybe it's just a matter of wording? Would it be better to write "Consider using LZ4_decompress_safe() instead" for `LZ4_uncompress()`?

Cyan4973:
Let me have another look at this topic.
It's likely that there are consistency issues (such as the ones you underline) that can be improved.

Cyan4973:
Looking at the source code,
the documentation already states that these variants should be replaced by `LZ4_decompress_safe_partial()`,
however, it hasn't updated the compilation warning message accordingly.
So I'll put up a PR which does just that.

PR #1312, opened 2023-12-05
decomp: refine read_variable_length codegen layout

JunHe77:
The probability of multiple-bytes length is low.
Split the first byte extraction from the loop,
add hint of comparison to get more pipeline friendly code, and move the reset of loop to a seperate cold path. See: https://godbolt.org/z/z4jd5z7z9

Change-Id: Ic93ae717954574435357fdb84624d00b0c280331

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1931,6 +1931,17 @@ read_variable_length(const BYTE** ip, const BYTE* ilimit,
     if (initial_check && unlikely((*ip) >= ilimit)) {    /* read limit reached */
         return rvl_error;
     }
+    s = **ip;
+    (*ip)++;
+    length += s;
+    if (unlikely((*ip) > ilimit)) {    /* read limit reached */
+        return rvl_error;
+    }
+    /* accumulator overflow detection (32-bit mode only) */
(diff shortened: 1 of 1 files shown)


JunHe77:
Tested this patch using silesia.tar with gcc-13 and clang-16 on various platforms.
For GCC-13 it is about 0.7% ~1.2% on Arm N1/N2 cores, -0.8% ~ -1% on xeon5218 and epyc7513.
For clang-16 it is about ~2% on Arm N1/N2, -1.5% on xeon5218 and ~1.5% on epyc7513.

Cyan4973 (approved):
The explanation makes sense, as well as the expected performance benefit.

As usual unfortunately, observing the performance benefit is difficult,
since on many platforms, other side effects such as random instruction alignment 
trigger larger performance changes than the patch itself.

Testing and comparing this PR over a wide range of compilers and compiler versions, performance impact is all over the place, sometimes positive, sometimes not, likely due to stated unrelated random side-effects.

Yet, I will err towards the side of accepting this PR,
as the reasoning looks correct to me and the effect looks sound and is well documented.

PR #1314, opened 2023-12-08
updated code documentation

Cyan4973:
notably around migration advice from deprecated `_fast`  functions.

fix #1310.

### DIFF
--- a/lib/lz4.h
+++ b/lib/lz4.h
@@ -144,22 +144,22 @@ LZ4LIB_API const char* LZ4_versionString (void);   /**< library version string; 
 
 
 /*-************************************
-*  Tuning parameter
+*  Tuning memory usage
 **************************************/
-#define LZ4_MEMORY_USAGE_MIN 10
-#define LZ4_MEMORY_USAGE_DEFAULT 14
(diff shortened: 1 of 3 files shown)


lib/lz4.h
```diff
@@ -546,9 +546,9 @@ LZ4_decompress_safe_partial_usingDict(const char* src, char* dst,
 #define LZ4_STATIC_3504398509
 
 #ifdef LZ4_PUBLISH_STATIC_FUNCTIONS
-#define LZ4LIB_STATIC_API LZ4LIB_API
+# define LZ4LIB_STATIC_API LZ4LIB_API
```
tomerv, line 549: How does this work with the space after the `#`?

t-mat: https://gcc.gnu.org/onlinedocs/cpp/The-preprocessing-language.html

> 1.4 The preprocessing language
> ... Whitespace is allowed before and after the '`#`' ...

See also C11 draft - 6.10 Preprocessing directives
https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf

PR #1323, opened 2023-12-19
remove Windows-specific `$(EXT)` suffix

Cyan4973:
`Makefile` is only understood and therefore useful for a few specific environments under Windows.

When tested within supported environments (`mingw64`, `msys2` and `cygwin`), the current `Makefile` is able to produce Windows executable binaries with the expected `*.exe` suffix without any need to add this suffix manually in the recipe.
This works even for extended command such as `$(RM)`, which will correctly remove `target.exe` even if only the name `target` appears in the list.

This makes it possible to consume a regular `Makefile` under these environments, simplifying the redaction, clarity and maintenance of these files.

This approach is proposed as an alternative to #1243.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ lz4 : liblz4.a
 lz4-release : lib-release
 lz4 lz4-release :
 	$(MAKE) -C $(PRGDIR) $@
-	$(LN_SF) $(PRGDIR)/lz4$(EXT) .
+	$(LN_SF) $(PRGDIR)/lz4 .
 
 .PHONY: examples
 examples: liblz4.a
@@ -83,7 +83,7 @@ clean:
 	$(MAKE) -C $(EXDIR) $@ > $(VOID)
 	$(MAKE) -C $(FUZZDIR) $@ > $(VOID)
 	$(MAKE) -C contrib/gen_manual $@ > $(VOID)
-	$(RM) lz4$(EXT)
+	$(RM) lz4
 	$(RM) -r $(CMAKE_BUILD_DIR)
 	@echo Cleaning completed
 
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -57,14 +57,12 @@ WINBASED    = yes
       else
 LIBLZ4_NAME = liblz4
 WINBASED    = no
-EXT         =
       endif
     endif
   endif
 endif
 
 ifeq ($(WINBASED),yes)
-EXT        = .exe
 WINDRES ?= windres
 endif
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -102,6 +102,7 @@ build_script:
       msbuild "build\VS2010\lz4.sln" /m /verbosity:minimal /property:PlatformToolset=v140 /t:Clean,Build /p:Platform=%PLATFORM% /p:Configuration=%CONFIGURATION% /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll" &&
       COPY build\VS2010\bin\%PLATFORM%_%CONFIGURATION%\*.exe programs\
     ) else (
+      DIR tests\ &&
       COPY tests\*.exe programs\
     )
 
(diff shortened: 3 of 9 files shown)


Cyan4973:
I see one test failing, 
could it be related to `mingw32`?

Cyan4973:
OK, looking into the one failing test on Appveyor.
Not sure if I understand it properly.

Compilation seems to work fine,
however, on reaching the test section, we see: 
```
if [%COMPILER%]==[gcc] (COPY tests\*.exe programs\)
tests\*.exe 
The system cannot find the file specified.
```

So it seems the issue comes on trying to execute the `COPY` command.

It's not clear what kind of environment is this, since it issues typical Windows `cmd` commands, such as `COPY`, instead of the more usual `cp` posix command. So it does not seem to be a typical `mingw64` `mintty` `posix` environment. It's a pity because it's very different from local reproduction scenarios, and iterating over `appveyor` CI service is pretty slow, to put it mildly.
It's not clear, from the log, if the `COPY` command is executed, or if only the argument remains.

Cyan4973:
OK, so on [this particular test](https://ci.appveyor.com/project/YannCollet/lz4-1lndh/builds/48788354/job/534gk20qjgq3j1vl), for some reason,
the `Makefile` produces executable binaries without the `*.exe` suffix after this PR.

Why on this one scenario, but not on any other Windows scenario, neither the 6 other ones on Appveyor, nor the other ones on Github Actions, that's not clear.

As mentioned earlier, it's not clear what's the weird environment in which this specific test is run. All I know is that's not one I've been able to reproduce on any of my Windows test systems so far.
And it's also unclear if it's a relevant scenario, or just a weird corner case.

Cyan4973:
OK, so I believe I have reproduced an environment which is closer to what Appveyor uses.

Instead of relying on the environment provided by `mingw64`, which involves a `mintty` shell coupled with `bash` or `zsh` for a more `posix` console experience, appveyor uses Windows `cmd` directly. This explain use of commands like `COPY` or `DIR`.
However, binaries from `mingw64` are also made accessible by editing this `PATH` variable, so it makes commands such as `cp` or `ls` also available, as well as the more important `make`, `gcc` and `clang`.

This feels like a hackish Frankensteinian environment, and it makes me wonder about its relevance to real life.

Anyway, point is, even after reproducing this environment locally on a Windows 10 system,
_it still results in producing correct executables_ like `lz4.exe` with the proper `.exe` suffix.

There are some remaining differences,
notably my `mingw64` environment is up to date, while the appveyor one is not.
Main visible differences are :
- `make`: 4.4.1 vs 4.2.1
- `clang`: 17.0.6 vs 9.0.1

It's extremely challenging to install older versions of a package on `mingw64`, so I'm not going to try.

My current suspicion is a change of behavior of `clang`, because `make` + `gcc` doesn't have the same issue, it properly produces `lz4.exe`, even though the `make` version is the same 4.2.1. Which leaves the `clang` version difference as the remaining suspect.

Anyway, most evidences so far point to an evolving situation, which used to require the `$(EXE)` manual suffix trick, but does not seem necessary anymore, provided users employ up to date tools.

It seems this is enough to justify the direction, aka remove the `$(EXE)` trick which makes the `Makefile` more difficult to read and maintain. In which case, there is still a need to figure out what to do about the currently failing `appveyor` `clang` test.

Cyan4973:
___Follow up___ : 

Indeed, `gcc` introduced this behavior (implicit `*.exe` suffix) for `gcc-8` in 2017: https://gcc.gnu.org/git/?p=gcc.git;a=commitdiff;h=5bc86b599054f494ec0a45e49b82749320eaa9c4;hp=77efd154f33154e819186de79687ea55552139f4
`clang` copied it later (2020) : https://reviews.llvm.org/D88005
That's why the explicit `$(EXT)` manual suffix trick is no longer necessary going forward.

While manually adding `$(EXT)` suffix is _possible_, it comes at subtle cost, being at odd with default construction rules, preventing recipe simplification, and generally getting in the way of `Makefile` refactoring, ultimately blocking more ambitious attempts at providing advanced functionalities. It's a hidden "lost opportunity" cost.

That being said, my main concern is that 2020 is still not that old, and such a change (removing manual `$(EXT)` setting) would require all `make` users to employ some recent enough compiler. Henceforth, an open question is: is it too soon to remove $(EXT) ?

Cyan4973:
It might be a little too early to remove the `$(EXT)` work around for old compilers.
Let's keep around for 1 or 2 more years.

Issue #1334, opened 2023-12-29
report thread: WIP `asyncio` (aka multi-threading)


t-mat:
Let's discuss about / play with [`asyncio` branch](https://github.com/lz4/lz4/tree/asyncio).

Since we don't have GitHub ["Discussions"](https://github.com/features/discussions), I'd like to put some report here.

Please feel free to post your own experiment.

t-mat:
I've tested new WIP `asyncio` branch a bit on Windows.  Even on WSL2, it works great.

- Target file : the almost incompressable file [ggml-alpaca-7b-q4.bin](https://github.com/ggerganov/llama.cpp)
- Target environments
  - WSL2 (ubuntu 22, gcc-11.4, on NTFS ([slow file I/O](https://github.com/microsoft/WSL/issues/4197)))
  - Windows 11 (Visual Studio 12, cl.exe 19.38.33133)

In this case WSL (gcc, pthread) version with `--best` works *7 times faster* than single threaded MSVC version (95.0 sec -> 13.1 sec).

Task manager's CPU graph reports that WSL version works fine and it utilizes all cores well.

<img width="582" alt="lz4-multi-threaded" src="https://github.com/lz4/lz4/assets/898396/6e4d6a37-333c-4072-a940-ffb3134fc4e3">

It seems `--fast` on WSL has some issue.  But I think it's mainly caused by slow NTFS I/O.  I'll check it with WSL's "native" file system later.

<details><summary>Log</summary>

```
# branch

> git branch -v
* asyncio c6762ec [behind 3] fixed meson build
```

```
# WSL

$ ./programs/lz4 --version
*** lz4 v1.9.5 64-bits multithread, by Yann Collet ***

$ time ./programs/lz4 -f --fast ./ggml-alpaca-7b-q4.bin
Compressed filename will be : ./ggml-alpaca-7b-q4.bin.lz4
Compressed 4212727017 bytes into 4212583972 bytes ==> 100.00%

real    0m19.549s
user    0m1.357s
sys     0m1.304s

$ time ./programs/lz4 -f --best ./ggml-alpaca-7b-q4.bin
Compressed filename will be : ./ggml-alpaca-7b-q4.bin.lz4
Compressed 4212727017 bytes into 4194322100 bytes ==> 99.56%

real    0m13.084s
user    2m2.647s
sys     0m2.135s
```

```
# MSVC (Powershell)

> .\bin\x64_Release\lz4.exe --version
[...]

PR #1336, opened 2023-12-29
Add multi-threading compression

Cyan4973:
This patch adds multi-threading support for compression to the `lz4` CLI.

Features :
- Display if the binary supports multithreading or not (welcome message) 
- Automatically select a nb of threads, based on local cpu
  + can also be controlled manually
- Provides a detailed summary in verbose mode, featuring speed and cpu utilization

Here are a few results using this patch : 

| cpu | corpus | c.level | v1.9.4 | this PR | speedup |
| --- | --- | --- | --- | --- | --- |
| i7-9700 | enwik9 | 1 | 2.3 s | 0.45 s | x5.1 |
| i7-9700 | silesia.tar | 9 | 5.5 s | 1.0 s | x5.5 |
| i7-9700 | silesia.tar | 12 | 17 s | 3.0 s | x5.6 |
| i7-9700 | enwik9 | 9 | 23 s | 4.1 s | x5.6 |
| M1 Pro | enwik9 | 1 | 2.2 s | 0.42 s | x5.2 |
| M1 Pro | silesia.tar | 9 | 5.1 s | 0.75 s | x6.8 |
| M1 Pro | silesia.tar | 12 | 16.5 s | 2.5 s | x6.6 |
| M1 Pro | enwik9 | 9 | 21 s | 3.0 s | x7.0 |

This is a pretty big patch, and I would expect various issues to crop up during testings.
Therefore it's going to be important to test this feature thoroughly before considering building a new release around it.

Open discussions on https://github.com/lz4/lz4/issues/1334

Known current limitations : 
- Requires `pthread` (i.e. mostly `posix` environments)
- Compression only
- Independent blocks only (revert to single thread for linked blocks)
- When there are multiple files, they are still processed sequentially


programs/util.c
```diff
@@ -0,0 +1,243 @@
...
+                        free(buffer);
+                    buffer = (PSYSTEM_LOGICAL_PROCESSOR_INFORMATION)malloc(returnLength);
+
+                    if (buffer == NULL) {
+                        perror("lz4");
+                        exit(1);
+                    }
+                } else {
+                    /* some other error */
+                    goto failed;
```
t-mat, line 99: It may not happen.  But we can think about the following scenario : 
- `glpi()` returns `ERROR_INSUFFICIENT_BUFFER` at the first time.
- `buffer` is allocated, `while()` loop continues.
- `glpi()` fails at the second time. (for some reason)
- Here, potentially `buffer` will be leaked.

PR #1341, opened 2023-12-31
Support C11 <threads.h> (optionally)

t-mat:
This PR adds simple compatibility macro for C11 `<threads.h>`.
These macros mimic `pthread_*` functions and types.

We can test this PR with the following command:
(EDIT 2024-04-08 : add `objdump` to check actual executable file)

```
cd
git clone -b tpool-c11-threads https://github.com/t-mat/lz4 t-mat-lz4
cd t-mat-lz4

make clean default
./lz4 --version
objdump -TC ./lz4 | grep -E "pthread|cnd_|thrd_|mtx_"
# POSIX pthread.h functions will be listed

make clean default CFLAGS=-DLZ4IO_USE_STD_THREADS_H=1
./lz4 --version
objdump -TC ./lz4 | grep -E "pthread|cnd_|thrd_|mtx_"
# C11 threads.h functions will be listed
```

## Potential issue#1 : Some macro ignores its argument.

Some macro ignores its argument.
It may cause problem.  For example, this PR ignores the second argument of `pthread_create(a, b++, c, d)`.

## Potential issue#2 : How `LZ4IO_USE_STD_THREADS_H` should be defined?

I hope `#if __has_include(<threads.h>)` will work properly.  But I don't know much about compatibility issue of this style.

### DIFF
--- a/programs/threadpool.c
+++ b/programs/threadpool.c
@@ -68,9 +68,34 @@ void TPOOL_completeJobs(TPOOL_ctx* ctx) {
 
 #else
 
-/* pthread only */
+/* pthread or standard <threads.h> */
 #include <stdlib.h>  /* malloc, free*/
-#include <pthread.h> /* pthread_* */
+
+#if defined(LZ4IO_USE_STD_THREADS_H) && (LZ4IO_USE_STD_THREADS_H)
+#  include <threads.h> /* thrd_*, mtx_*, cnd_* */
+#  define pthread_t                     thrd_t
+#  define pthread_create(a,b,c,d)       thrd_create(a,c,d)    /* ignore 2nd argument */
+#  define pthread_join                  thrd_join
+#  define pthread_mutex_t               mtx_t
+#  define pthread_mutex_init(a,b)       mtx_init(a,mtx_plain) /* ignore and replace 2nd argument */
+#  define pthread_mutex_destroy         mtx_destroy
+#  define pthread_mutex_lock            mtx_lock
+#  define pthread_mutex_unlock          mtx_unlock
+#  define pthread_cond_t                cnd_t
+#  define pthread_cond_init(a,b)        cnd_init(a)           /* ignore 2nd argument */
+#  define pthread_cond_destroy          cnd_destroy
+#  define pthread_cond_wait             cnd_wait
+#  define pthread_cond_signal           cnd_signal
+#  define pthread_cond_broadcast        cnd_broadcast
+#  define TPOOL_thread_ReturnType       int
+#  define TPOOL_thread_ReturnFailure    0
+#  define TPOOL_thread_ReturnSuccess    (!0)
+#else
+#  include <pthread.h> /* pthread_* */
+#  define TPOOL_thread_ReturnType       void*
+#  define TPOOL_thread_ReturnFailure    NULL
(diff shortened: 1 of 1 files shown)


t-mat:
Cirrus CI failed to update the FreeBSD repo.

image

Cyan4973:
> Cirrus CI failed to update the FreeBSD repo.

This is fixed in latest `dev` update

Cyan4973:
In which scenario does this approach provide a benefit ?
I presume, it would be an environment that does not provide `<pthread.h>` but would support `C11` `<threads.h>` ?
Windows maybe ?

t-mat:
> Windows maybe ?

Yes.  I mainly think about (next release of) MSVC.  Since the difference between `pthread.h` and `threads.h` is smaller than vs WIN32.

Cyan4973:
> I hope `#if __has_include(<threads.h>)` will work properly. But I don't know much about compatibility issue of this style.

`<threads.h>` being optional, testing its presence is indeed an issue, which is similar to `<pthread>`.

An approach, used in this `Makefile` for `<pthread.h>`, is to create a small source code which depends on the existence of this library, and test if compilations fails or succeed. And if it succeed, set a build macro which enables this code path.
This is similar in spirit to `./configure`. However, it's build driven, as opposed to source driven, so it doesn't work outside of `make`.

`__has_include()` is source-driven, so it would be more general. However, it's not part of the C standard, so it's a compiler feature, generally defined as a macro.
Therefore its existence must be tested, and only then employed accordingly.

Both approaches could be combined, it it makes sense (to cover non-overlapping sets of environments).
In the end, what's useful is that all detection methods converge towards controlling a build variable, that can be manually overridden by the user. A user may prefer disabling this feature, even if it's available. Or may want to ensure that the binary is built with the feature enabled only.

Cyan4973:
It would probably be a good idea to add a test for this feature.

_edit_ : there are a few Windows tests which currently produce single-threaded version of `lz4` (detectable thanks to the welcome message, such as `lz4 -V`). They could (maybe?) be refactored to employ this C11 alternative ?

leafi:
Hey, might it be worth including some kind of assert (static_assert? too new C std?) to make sure `thrd_success` from threads.h actually equals 0?

I don't know if any system actually exists where `thrd_success` is non-zero -- it's certainly 0 on this Linux system -- but [https://en.cppreference.com/w/c/thread/thrd_errors](https://en.cppreference.com/w/c/thread/thrd_errors) claims it's undefined and ^F in [https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1548.pdf](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1548.pdf) doesn't show it being defined as 0.

t-mat:
@leafi, thank you for pointing this out!

You're right.  Their return value compatibility must be checked as you suggested.  Or all macro must return pthread compatible value.

t-mat:
> Hey, might it be worth including some kind of assert (static_assert? too new C std?) to make sure `thrd_success` from threads.h actually equals 0?

dada21b addressed this issue.

Cyan4973:
In #1429 , I've added multithreading support for Windows using a different method,
by providing a dedicated implementation of ThreadPool for Windows, based on Completion Ports.

This works fine because `lz4io` does not need any other synchronization primitive beyond the Thread Pool.

Consequently, there isn't a need to "translate" `<pthread>` for Windows.

PR #1343, opened 2023-12-31
Compile-time constants

Cyan4973:
now regrouped in `programs/lz4conf.h`.

### DIFF
--- a/programs/lz4io.c
+++ b/programs/lz4io.c
@@ -32,6 +32,21 @@
 
 
 /*-************************************
+*  Compile-time parameters
+**************************************/
+/* Determines if multithreading is enabled or no
+ * Default: disabled */
+#ifndef LZ4IO_MULTITHREAD
+# define LZ4IO_MULTITHREAD 0
+#endif
+
+/* Maximum nb of threads that can selected at runtime */
+#ifndef LZ4IO_NB_WORKERS_MAX
+# define LZ4IO_NB_WORKERS_MAX 125
+#endif
+
+
+/*-************************************
 *  Compiler options
 **************************************/
 #ifdef _MSC_VER    /* Visual Studio */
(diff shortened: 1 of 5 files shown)


programs/lz4io.c
```diff
@@ -31,6 +31,21 @@
...
+/* Determines if multithreading is enabled or no
+ * Default: disabled */
+#ifndef LZ4IO_MULTITHREAD
+# define LZ4IO_MULTITHREAD 0
+#endif
+
+/* Maximum nb of threads that can selected at runtime */
+#ifndef LZ4IO_NB_WORKERS_MAX
+# define LZ4IO_NB_WORKERS_MAX 125
+#endif
```
dearblue: How about putting a header file that consolidates configuration definitions under the `programs/` directory?
For example, `programs/lz4conf.h`, `programs/lz4def.h`, etc.?

I think it would also be useful as an alternative to documentation.

t-mat: Hi @dearblue,

Could you think about to post your opinion under ["Issues"](https://github.com/lz4/lz4/issues/new/choose) (if you have a plan) or ["Discussions"](https://github.com/lz4/lz4/discussions/new/choose) (if it's still vague)?

Although it'll be your burden, but since this Pull Request will be merged/closed, your opinion may also be forgotten  with it.

dearblue: @t-mat, thanks for your reply.
I posted it at https://github.com/lz4/lz4/discussions/1344 as an idea.

Cyan4973: > How about putting a header file that consolidates configuration definitions under the `programs/` directory?
> For example, `programs/lz4conf.h`, `programs/lz4def.h`, etc.?

Indeed, I was also pondering about the same idea.

I initially considered keeping this topic for a later PR,
but alternatively, if it feels preferable, this PR could be refactored to include this concept.

dearblue: I'm sorry I disrupted your schedule.
And, nice change!
\# :crying_cat_face: However, for some reason I can't find "Resolve conversation" button, so I can't fold this review. Please forgive me.

PR #1353, opened 2024-01-23
Fix Python 3.6 string interpolation

likema:


Cyan4973 (approved):
I'm not the best at parsing python code invoking regular expression,
but it looks to me that this simplification should indeed achieve the same outcome, 
on top of being more readable.

PR #1356, opened 2024-01-28
added a lorem ipsum generator

Cyan4973:
and use it as primary synthetic source for the internal benchmark program
because it produces a more satisfying speed / ratio
than the previous purely statistical generator
which was primarily designed for Entropy compressors
and doesn't match well the expected structure of natural text.


data-man:
> GPL v2 License

Very sad.

PR #1360, opened 2024-01-29
add sparc compilation test

Cyan4973:


Cyan4973:
There's something weird going on.
I have an almost identical test in another project, where [it works fine](https://github.com/facebook/zstd/actions/runs/7691177864/job/20956093323?pr=3886).
But here, just starting any of the produced binary [immediately crashes](https://github.com/lz4/lz4/actions/runs/7692278604/job/20958939861?pr=1360#step:7:82).
That's a pretty wild difference, and I don't see what could be specifically `lz4` to explain such a different outcome ...

_note_: 
although, there is a difference,
the working version uses `gcc` 9.4.0 and `qemu ` 4.2.1,
while the failing one uses `gcc` 11.4.0 and `qemu` 6.2.0,
courtesy of a difference in `ubuntu` version.

So there is a chance that the newer version could actually be buggy ?

Cyan4973:
Yep, that was it.
`qemu` for `sparc` as shipped in `ubuntu-22` is apparently broken.
Using `ubuntu-20` instead fixed it.

PR #1399, opened 2024-05-01
Added library.json for PlatformIO Compatibility

ab-tools:
Would you mind adding this `library.json` file to easily allow using this library within PlatformIO by just referencing your GIT repository?

Thanks
Andreas

### DIFF
--- a/library.json
+++ b/library.json
@@ -1,0 +1,20 @@
+{
+  "name": "lz4",
+  "version": "1.9.4",
+  "keywords": "lz4, compress, esp32, esp826, espressif",
+  "description": "lz4 data compress library, fork from awesome https://github.com/lz4/lz4",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/liux-pro/lz4"
+  },
+  "frameworks": "*",
+  "platforms": "*",
+  "build": {
+    "flags": [
+      "-DLZ4_HEAPMODE"
+    ],
+    "includeDir": "lib",
+    "srcDir": "lib",
+    "srcFilter": ["-<*>", "+<lz4.c>", "+<lz4.h>"]
+  }
+}
--- a/library.json
+++ b/library.json
@@ -1,11 +1,11 @@
 {
   "name": "lz4",
-  "version": "1.9.4",
+  "version": "1.10.0",
   "keywords": "lz4, compress, esp32, esp826, espressif",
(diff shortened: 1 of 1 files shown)


Cyan4973:
Where is this file added ? At the root on the repository ?
In which case, I have 2 questions : 
- Is it possible to add it to a dedicated subdirectory instead ? (exact path to be determined)
- Does the format of `library.json` allow comments, thus making it possible to document in the file why it's there and how it works ?

ab-tools:
Hello Yann,

first thanks for your reply.

According to the PlatformIO documentation the `library.json` file has to be in the root directory of the library:
https://docs.platformio.org/en/latest/librarymanager/creating.html#publishing

As it is plain JSON, it cannot contain comments. But you could add a note in the library `description` JSON element within this file that it is used for PlatformIO.

It would be ideal if the package would be officially published to the PlatformIO registry by the lz4 team here as well:
https://registry.platformio.org

Currently, there is only a very outdated fork available there:
https://registry.platformio.org/libraries/liux-pro/lz4

Thanks for considering
Andreas

library.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "lz4",
+  "version": "1.9.4",
```
Cyan4973: It seems this file implies a manual maintenance of the `version` field

library.json
```diff
@@ -0,0 +1,20 @@
...
+  "keywords": "lz4, compress, esp32, esp826, espressif",
+  "description": "lz4 data compress library, fork from awesome https://github.com/lz4/lz4",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/liux-pro/lz4"
```
Cyan4973: I presume the url should be updated

library.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "lz4",
+  "version": "1.9.4",
+  "keywords": "lz4, compress, esp32, esp826, espressif",
+  "description": "lz4 data compress library, fork from awesome https://github.com/lz4/lz4",
```
Cyan4973: Well, it's no longer a fork

library.json
ab-tools: Description corrected.

library.json
ab-tools: URL updated.

ab-tools:
Would be great if this could be merged, @Cyan4973!

Thanks
Andreas

PR #1400, opened 2024-05-06
Remove code branch

Nicoshev:
Hello, I hope you are doing well : )

The code branch being removed is equivalent to one iteration of the loop within LZ4_wildCopy8.
It should be slightly faster to remove the branch and let the loop iterate 1 more time.
Because LZ4_wildCopy8 has a do-while, before this removal a minimum of 16 bytes of memory were written when offset>=8.
Now if the match length is at most 8 bytes while offset>=8, only 8 bytes of memory are written.

Regarding to performance, variance across tests seems to be greater than this change's impact.
However, after several runs, the highest throughputs were achieved when this change was included.

Regards,
Nicolas

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -506,10 +506,6 @@ LZ4_memcpy_using_offset_base(BYTE* dstPtr, const BYTE* srcPtr, BYTE* dstEnd, con
         LZ4_memcpy(dstPtr+4, srcPtr, 4);
         srcPtr -= dec64table[offset];
         dstPtr += 8;
-    } else {
-        LZ4_memcpy(dstPtr, srcPtr, 8);
-        dstPtr += 8;
-        srcPtr += 8;
     }
 
     LZ4_wildCopy8(dstPtr, srcPtr, dstEnd);


Cyan4973:
Hi @Nicoshev 

Let's start by some benchmark results, comparing this PR with `dev` branch :

<details>
  <summary>Benchmark results, core i7-9700k</summary>
  <pre><code>
```
PR1400                                                                         | dev
compile clean lz4 with gcc-7                                                   │
116bce53d100657a8c9fd9437902e0dc  lz4                                          │f534dc4e198ab0e93f4e32e379240aee  lz4
Benchmark Decompression of LZ4 Frame _without_ checksum even when present      │Benchmark Decompression of LZ4 Frame _without_ checksum even when present
 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3361.7 MB/s │ 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 2822.3 MB/s
 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3221.4 MB/s │ 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 2557.8 MB/s
 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 2967.3 MB/s │ 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 2380.1 MB/s
compile clean lz4 with gcc-8                                                   │compile clean lz4 with gcc-8
0110fc1e8e2a360b548a294fd7ba3cbd  lz4                                          │53723ca58bcbca69ad31831bdeeef577  lz4
Benchmark Decompression of LZ4 Frame _without_ checksum even when present      │Benchmark Decompression of LZ4 Frame _without_ checksum even when present
 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3431.1 MB/s │ 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3678.7 MB/s
 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3361.4 MB/s │ 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3596.8 MB/s
 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 3066.6 MB/s │ 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 3284.2 MB/s
compile clean lz4 with gcc-9                                                   │compile clean lz4 with gcc-9
2d572e28bee6863189cbc8179a822499  lz4                                          │d2f0e2fb23363e7b376d311f1e6656ab  lz4
Benchmark Decompression of LZ4 Frame _without_ checksum even when present      │Benchmark Decompression of LZ4 Frame _without_ checksum even when present
 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 2788.4 MB/s │ 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3242.8 MB/s
 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 2544.8 MB/s │ 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3111.2 MB/s
 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 2365.1 MB/s │ 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 2855.8 MB/s
compile clean lz4 with gcc-10                                                  │compile clean lz4 with gcc-10
c64bbfcd432c8dc82ea80dbd1998000a  lz4                                          │03ec7af6f458531cc0ebd2f80dd7320e  lz4
Benchmark Decompression of LZ4 Frame _without_ checksum even when present      │Benchmark Decompression of LZ4 Frame _without_ checksum even when present
 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3239.6 MB/s │ 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 2834.8 MB/s
 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3115.7 MB/s │ 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 2569.7 MB/s
 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 2861.4 MB/s │ 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 2395.0 MB/s
compile clean lz4 with gcc-11                                                  │compile clean lz4 with gcc-11
460eefdb51b88981717bb89180af2090  lz4                                          │dd631a74f73a4d78f92b9aaec4457307  lz4
Benchmark Decompression of LZ4 Frame _without_ checksum even when present      │Benchmark Decompression of LZ4 Frame _without_ checksum even when present
 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3079.5 MB/s │ 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3074.9 MB/s
 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 2900.9 MB/s │ 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 2899.9 MB/s
 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 2669.0 MB/s │ 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 2666.3 MB/s
compile clean lz4 with clang-6.0                                               │compile clean lz4 with clang-6.0
b6444a79dde9857ba53282ab5136195f  lz4                                          │b45308aa293056f2a684cc2d7cf75579  lz4
Benchmark Decompression of LZ4 Frame _without_ checksum even when present      │Benchmark Decompression of LZ4 Frame _without_ checksum even when present
 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3600.0 MB/s │ 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3179.6 MB/s
 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3491.2 MB/s │ 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3073.4 MB/s
 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 3218.6 MB/s │ 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 2834.7 MB/s
compile clean lz4 with clang-7                                                 │compile clean lz4 with clang-7
4723a3236ba1ac6f21e1fb677295c198  lz4                                          │5872945ba59beaaf97ffc131235e8427  lz4
Benchmark Decompression of LZ4 Frame _without_ checksum even when present      │Benchmark Decompression of LZ4 Frame _without_ checksum even when present
 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3208.6 MB/s │ 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3585.3 MB/s
 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3087.9 MB/s │ 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3484.7 MB/s
 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 2854.3 MB/s │ 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 3204.8 MB/s
compile clean lz4 with clang-8                                                 │compile clean lz4 with clang-8
0aeec8d327b8e1b222b5909206907d7d  lz4                                          │999021f090ba65dfdf344c498c911140  lz4
Benchmark Decompression of LZ4 Frame _without_ checksum even when present      │Benchmark Decompression of LZ4 Frame _without_ checksum even when present
 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3604.2 MB/s │ 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3118.9 MB/s
 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3501.5 MB/s │ 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3044.5 MB/s
 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 3228.2 MB/s │ 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 2789.1 MB/s
compile clean lz4 with clang-9                                                 │compile clean lz4 with clang-9
4cf97f7280edc377353f7a74c58e8622  lz4                                          │6c6674a049f1ea2c29575aa63beca9ae  lz4
Benchmark Decompression of LZ4 Frame _without_ checksum even when present      │Benchmark Decompression of LZ4 Frame _without_ checksum even when present
 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3324.3 MB/s │ 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3746.7 MB/s
 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3154.0 MB/s │ 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3588.2 MB/s
 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 2909.9 MB/s │ 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 3295.4 MB/s
compile clean lz4 with clang-10                                                │compile clean lz4 with clang-10
6fdd3e48fe8e24053312931b90119b77  lz4                                          │bd5a1126a70ea573d175a711c94524d8  lz4
Benchmark Decompression of LZ4 Frame _without_ checksum even when present      │Benchmark Decompression of LZ4 Frame _without_ checksum even when present
 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3322.4 MB/s │ 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3752.2 MB/s
 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3151.0 MB/s │ 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3581.1 MB/s
 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 2905.3 MB/s │ 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 3302.0 MB/s
compile clean lz4 with clang-11                                                │compile clean lz4 with clang-11
f29190a9f208589960c688882cff7312  lz4                                          │8fdf3b95762359425db720ba1f137931  lz4
Benchmark Decompression of LZ4 Frame _without_ checksum even when present      │Benchmark Decompression of LZ4 Frame _without_ checksum even when present
 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3779.9 MB/s │ 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3336.8 MB/s
 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3591.0 MB/s │ 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3163.5 MB/s
 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 3334.7 MB/s │ 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 2916.5 MB/s
compile clean lz4 with clang-12                                                │compile clean lz4 with clang-12
3b4db1f59f565a5c0a684a1a6eedd7f3  lz4                                          │51f2dbb3d1966cf9cadbdac5d92f7110  lz4
Benchmark Decompression of LZ4 Frame _without_ checksum even when present      │Benchmark Decompression of LZ4 Frame _without_ checksum even when present
 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3391.1 MB/s │ 1#lesia.tar.L12.lz4 : 211957760 ->  77382713 (2.739),   0.0 MB/s, 3393.6 MB/s
 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3328.6 MB/s │ 1#lgary.tar.L12.lz4 :   3153920 ->   1185988 (2.659),   0.0 MB/s, 3328.2 MB/s
 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 3035.9 MB/s │ 1#enwik9.L12.lz4    :1000000000 -> 372443347 (2.685),   0.0 MB/s, 3038.3 MB/s
```
</code></pre>
</details>

As one can see, results are all over the place.
I can't tell if this change is beneficial or not from these measurements.
But clearly, the variance per compiler version is very high, and very likely unrelated to the change itself, but rather to the side effect of the change, influencing instruction alignment, with surprisingly high consequences (~+30% difference between slowest and fastest generated binaries).

So the next step is reasoning about the change.

Your claim is that, by removing this branch, the cpu will do one less `copy8` operation when `matchLen <= 8`, so it's less work, so it should be faster.

Alas, if only it was that simple. With modern OoO architectures, "amount of work" is no longer a sufficient unit, especially at extremely low instruction level, and other forms of irregularities tend to have way worse consequences.

I remember in the early days of LZ4, the code was much simpler back then, but there was an equivalent code path dealing with overlapping matches. I then noticed that having this unconditional `copy8` operation in the `else` branch was pretty important, resulting in a measurable performance benefit. 
My tentative explanation of this effect was likely completely wrong at the time, but the way I would see it today is as follows : 
With the unconditional `copy8` operation followed by `wildcopy8` which also starts by an unconditional `copy8` operation, there are now 16 bytes which are copied unconditionally, before reaching the first branch.
When removing the first `copy8` operation, there are now only 8 bytes copied before reaching the first branch.
Consequently, the statistics associated with this first branch are wildly different, and in the first case (copy 16 bytes then test), it becomes highly likely to return `false`, meaning the branch is fairly predictable, meaning its cost is very low. On the other hand, when only 8 bytes are copied, this first branch becomes less predictable, and such a difference was important enough to produce a measurably worse speed outcome.

As one can see, it changes the focus, from "amount of work" to "predictability of the code path".

So, according to above explanation, this change should be worse for performance.
Now how to test that ?
As mentioned before, benchmarks are inconclusive.

Well, first let's reason about this code path. When is it activated ?
This function, `LZ4_memcpy_using_offset_base()`, is only invoked once, within `LZ4_memcpy_using_offset()`.
`LZ4_memcpy_using_offset()` will essentially take care of `offset==1`, `2` and `4`, and delegate all other cases to `LZ4_memcpy_using_offset_base()`.
In turn, `LZ4_memcpy_using_offset()` is invoked only once, within `LZ4_decompress_generic()`. It's part of `LZ4_FAST_DEC_LOOP`, which a section which specializes in "short" sequences, aka `litLen <= 14` and `matchLen <= 18`.
Within this setup, `LZ4_memcpy_using_offset()` is only invoked if `offset<16`. And then, the impact of the removed branch is only active if `offset >= 8`.

So this code path only matters for short matches at an `offset` within `[8-15]`. That's a seriously reduced range. No wonders it's difficult to measure the impact: this code path is probably very rarely taken.

So, in order to measure something, one could try to create a "fake" file, designed to trigger this code path very often, in order to magnify the impact of the change and measure it.
And now it becomes a more complex / longer work, knowing full well that the impact of the change, whether negative or positive, is likely going to be very small, if perceptible at all, on real data.

Nicoshev:
Hey @Cyan4973, thanks a lot for the detailed response

Regarding the change's motivation:
Aside from reducing the work needed, I thought that removing a code branch should make execution more predictable.
As code size shrinks, the CPU sees the branch further in advance and has more time to make a decision.

Regardsing testing:
It is indeed very difficult to test this kind of changes, even with an input-based approach.
Some of these changes impact performance in an opposite manner depending on how often the affected branch is executed.
That is, the change decreases CPU time when the branch tends to be taken and increases CPU time when the branch is not taken often (or viceversa). 
I guess that taking it or not taking it also affects future predictions.
When using a given compiler to test a change, sometimes I get different performance results depending on the file being used.

I was testing changing the LZ4_wildCopy8 within LZ4_memcpy_using_offset_base by something like:

```
do { 
        LZ4_memcpy(dstPtr,srcPtr,8); 
        LZ4_memcpy(dstPtr+8,srcPtr+8,8); 
        LZ4_memcpy(dstPtr+16,srcPtr+16,8); 
        LZ4_memcpy(dstPtr+24,srcPtr+24,8); 
        dstPtr+=32; srcPtr+=32; 
} while (dstPtr<dstEnd);
```

This is actually legal to do, because LZ4_memcpy_using_offset_base is called within the fast dec loop

That change tends to improve performance, but again, not for absolutely every kind of file.

So then, decreasing CPU time when decompressing a given file does not seem to be a strong enough argument to merge a change.

I was also testing changing LZ4_wildCopy32 by the following:

```
LZ4_FORCE_INLINE void
LZ4_wildCopy64(void* dstPtr, const void* srcPtr, void* dstEnd)
{
    BYTE* d = (BYTE*)dstPtr;

    do { 
        LZ4_memcpy(d,s,16); 
        LZ4_memcpy(d+16,s+16,16); 
        LZ4_memcpy(d+32,s+32,16); 
        LZ4_memcpy(d+48,s+48,16); 
        d+=64; s+=64; 
    } while (d<e);
}
```

That change does seem to improve throughput by 1-4% in all modern CPUs.
Although I haven't proposed them formally yet due to not having it test it in all possible compilers.
But even if it still proves to be faster in most compilers, it may not be faster in weaker CPUs.

Aside from actually merging changes, I enjoy the learning these kind of experiments provide.
As you said, out of order execution engines are so complex that it's nice to test root causes we may theorize.

PR #1401, opened 2024-05-06
Update function comment

Nicoshev:
Hello, I hope you are doing well :)

LZ4_memcpy_using_offset documentation suggests it can overwrite up to 8 bytes after dstEnd.
However, LZ4_memcpy_using_offset_base actually writes at least 16 bytes from dstPtr.
Assuming dstEnd==dstPtr+4, LZ4_memcpy_using_offset can actually overwrite 12 bytes after dstEnd.

This does not seem to be an issue, as LZ4_memcpy_using_offset is only called within the fast dec loop.
Updating the function's documentation may still be worth it for future uses.

Regards,
Nicolas


Cyan4973 (approved):
I agree,
the documentation is outdated, 
and deserves to be fixed.

PR #1403, opened 2024-05-07
len should be unsigned

SiliconA-Z:
len will never underflow, as it is already bounds-checked, and casting is potentially another instruction.


Cyan4973:
why ?

SiliconA-Z:
> why ?

Because it will never underflow, and casting is potentially another instruction.

Cyan4973 (approved):

PR #1405, opened 2024-05-07
Define mlen = MINMATCH at the start of the loop

SiliconA-Z:


Cyan4973 (approved):
This is very, very minor, but ok.

PR #1413, opened 2024-05-18
[cmake] Always create lz4 target.

teo-tsirpanis:
See facebook/zstd#3965.

### DIFF
--- a/build/cmake/CMakeLists.txt
+++ b/build/cmake/CMakeLists.txt
@@ -116,16 +116,13 @@ if(BUILD_STATIC_LIBS)
     POSITION_INDEPENDENT_CODE ${LZ4_POSITION_INDEPENDENT_LIB})
   list(APPEND LZ4_LIBRARIES_BUILT lz4_static)
 endif()
-# Add unified target if building only a shared or static library.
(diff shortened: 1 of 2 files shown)


teo-tsirpanis:
CI failures are unrelated; the CMake jobs are green.

t-mat:
Sorry for CI issue.  #1416 will fix it.

Cyan4973:
Could you give an example of what was not working before and that works fine now with this PR ?

teo-tsirpanis:
In lz4 installled packages with both a static and shared library built, the unified `lz4::lz4` CMake target is not available. This PR defines it always, allowing developers to reference it with confidence and support both static and dynamic linkage of lz4 in their apps.

I opened this PR to mirror a similar PR to zstd, with the intent to proactively address an issue like facebook/zstd#3859 on lz4.

Cyan4973:
Is there any test we could do in CI that would "prove" the problem (aka fail) and succeed after this PR is applied ?

teo-tsirpanis:
Good idea, I added a check in the CMake tests.

Cyan4973 (approved):

Cyan4973:
Thanks @teo-tsirpanis !

PR #1431, opened 2024-07-06
build: minor: fix lz4 project in VS2017 solution

Cyan4973:
Note: it would probably be better to stop shipping and maintaining separate Visual Studio solutions manually,
and instead have them generated procedurally from (typically) `cmake` recipe before release.
But that's a project in itself...

### DIFF
--- a/build/VS2017/lz4/lz4.vcxproj
+++ b/build/VS2017/lz4/lz4.vcxproj
@@ -155,6 +155,9 @@
     <ClCompile Include="..\..\..\programs\lorem.c" />
     <ClCompile Include="..\..\..\programs\lz4cli.c" />
     <ClCompile Include="..\..\..\programs\lz4io.c" />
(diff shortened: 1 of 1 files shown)


t-mat:
Strongly :+1: for getting rid of the pure random `.vcxproj` format.

But for just in case, if we don't use Visual Studio IDE, we can use wildcard for `<ClCompile>` and `<ClInclude>`.

e.g. `<ClCompile Include="..\..\..\lib\*.c" />`

They've been supporting it with the following versions:

- [VS2017](https://learn.microsoft.com/en-us/cpp/build/reference/vcxproj-files-and-wildcards?view=msvc-150)
- [VS2019](https://learn.microsoft.com/en-us/cpp/build/reference/vcxproj-files-and-wildcards?view=msvc-160)
- [VS2022](https://learn.microsoft.com/en-us/cpp/build/reference/vcxproj-files-and-wildcards?view=msvc-170)

PR #1436, opened 2024-07-08
Linked Blocks compression (-BD) can employ multiple threads

Cyan4973:
also works with dictionary (`-D dictFile`).

Note: after this change, keeping a single-threaded version of the compressor in the source code becomes less relevant. All compression modes can now employ the multi-threaded version. This version silently turns into a single-threaded one when `LZ4IO_MULTITHREAD==0`.
This could be a good reason to reduce the code size.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -530,21 +530,22 @@ jobs:
           #   xemu : QEMU static emulator executable.
           #   os   : GitHub Actions YAML workflow label.  See https://github.com/actions/virtual-environments#available-environments
 
-          { type: ARM,      pkgs: 'qemu-system-arm   gcc-arm-linux-gnueabi',     xcc: arm-linux-gnueabi-gcc,     xemu: qemu-arm-static,     os: ubuntu-latest, },
-          { type: ARM64,    pkgs: 'qemu-system-arm   gcc-aarch64-linux-gnu',     xcc: aarch64-linux-gnu-gcc,     xemu: qemu-aarch64-static, os: ubuntu-latest, },
-          { type: PPC,      pkgs: 'qemu-system-ppc   gcc-powerpc-linux-gnu',     xcc: powerpc-linux-gnu-gcc,     xemu: qemu-ppc-static,     os: ubuntu-latest, },
-          { type: PPC64LE,  pkgs: 'qemu-system-ppc   gcc-powerpc64le-linux-gnu', xcc: powerpc64le-linux-gnu-gcc, xemu: qemu-ppc64le-static, os: ubuntu-latest, },
-          { type: S390X,    pkgs: 'qemu-system-s390x gcc-s390x-linux-gnu',       xcc: s390x-linux-gnu-gcc,       xemu: qemu-s390x-static,   os: ubuntu-latest, },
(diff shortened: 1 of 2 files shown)


Cyan4973:
There is a CI issue reported for the `m68k` `qemu` test, 
but it's unclear if this comes from this code base, or if it's a bug in either `qemu` or the kernel for `m68k`.

Cyan4973:
Given that the `m68k` architecture is pretty old and quite different from all other platforms, and given that all other emulated and native platforms seem to work fine, the suspicion is on the `m68k` platform or its emulation.

A work around could be to disable Multithreading when testing with `qemu-m68k-static`.

t-mat:
I've also tested this branch on my environment (Ubuntu 22.04, WSL2).

It ends with same error:

<details>

<summary>test lz4 qemu-m68k-static platform</summary>

```
 ---- test lz4 qemu-m68k-static platform ----
qemu-m68k-static ./datagen -g16KB  | qemu-m68k-static ../programs/lz4 -9     | qemu-m68k-static ../programs/lz4 -t

stdin                          : decoded 16384 bytes
qemu-m68k-static ./datagen         | qemu-m68k-static ../programs/lz4        | qemu-m68k-static ../programs/lz4 -t

stdin                          : decoded 65536 bytes
qemu-m68k-static ./datagen -g256MB | qemu-m68k-static ../programs/lz4 -vqB4D | qemu-m68k-static ../programs/lz4 -qt
using blocks of size 64 KB

The futex facility returned an unexpected error code.
...
qemu: uncaught target signal 6 (Aborted) - core dumped
Aborted
lz4: lz4io.c:2192: LZ4IO_decompressLZ4F: Assertion `lastStatus == 0' failed.
qemu: uncaught target signal 6 (Aborted) - core dumped
Aborted
make[1]: *** [Makefile:310: test-platform] Error 134
```

</details>


But when I split piped commands into separated commands:

- Compression program (`lz4 -vqB4D`) aborts
  - But it generates proper `.lz4` file
- Decompression program (`lz4 -d`) decompress it properly
- Original file and decompressed file are same.

It may suggest that program/process itself is working correctly but it doesn't terminate something gracefully.

<details>

<summary>Same test, but splitted</summary>

```
cd ./tests/

qemu-m68k-static ./datagen -g256MB > data-256MB
# OK

qemu-m68k-static ../programs/lz4 -vqB4D data-256MB
using blocks of size 64 KB
Compressed filename will be : data-256MB.lz4
The futex facility returned an unexpected error code.
qemu: uncaught target signal 6 (Aborted) - core dumped
Aborted
# FAILED ?

qemu-m68k-static ../programs/lz4 -d data-256MB.lz4 data-256MB.lz4.dec
data-256MB.lz4                 : decoded 268435456 bytes
# OK

cmp data-256MB data-256MB.lz4.dec
# OK
```

</details>

Cyan4973:
Thanks for the analysis @t-mat .
So the whole multithreaded compression works fine, 
it's closing a futex, at end of compression, that causes a segfault.

Cyan4973:
Let's test `m68k` without multithreading then. It's not a good fit anyway (multi-cores `m68k` doesn't exist).

Issue #1386, opened 2024-03-28
PoC: Use MSVC without project files


t-mat:
MSVC project files (.sln and friends) have been our burden.  Because their format, structure and convention have been (will be) changed randomly.

To ease this problem, I made a simple batch script which invokes `cl.exe` (MSVC compiler/linker driver) and `rc.exe` directly.

It's cumbersome, but its core is the following one line:

```
cl.exe %COMPILER_OPTIONS% %SRCS% /link %LINKER_OPTIONS% lz4.res
```

It passes compiler options, source files, linker options and resource file.

t-mat:
[gist](https://gist.github.com/t-mat/de819af37d3dd399147a75f9af1b034e)

Procedure:

```
cmd.exe
cd /d "%USERPROFILE%\Documents"
md lz4-issue-1386
cd lz4-issue-1386
git clone https://github.com/lz4/lz4.git
cd lz4

md build\visual_studio\lz4
cd build\visual_studio\lz4

copy ..\..\VS2022\lz4\lz4.rc .

curl -JOL https://gist.githubusercontent.com/t-mat/de819af37d3dd399147a75f9af1b034e/raw/build-vs2022.cmd
type build-vs2022.cmd | find "" /V > build-vs2022.cmd.bat
.\build-vs2022.cmd

.\lz4.exe --version
```

Cyan4973:
Yes, this is a good idea.

Cyan4973:
#1440  uses a somewhat different method to generate the Visual Studio solution, by leveraging `cmake` directly.

But this approach requires `cmake` to be installed and accessible on the local computer.

If this condition is deemed too limitative, one potential mitigation could be to provide at least one pre-generated solution within the repository.

Note that the  manually-defined `VS2022` solution is still present within `build/`, in part because it's used by some tests, and in another part because its content differs from the Solutions built from `cmake`, so existing users could be surprised by this difference. But this is a temporary situation, manually-defined solutions are set to be removed in a future version.

t-mat:
Since Visual Studio and its installer [supports `cmake`](https://learn.microsoft.com/en-us/cpp/build/cmake-projects-in-visual-studio?view=msvc-170#installation), I think requirement of `cmake` is 100% okay for them who can install Visual Studio.

I believe [`vsTests`](https://github.com/lz4/lz4/tree/vsTests) made our situation far better.  And it surpassed this PoC.

[...]

PR #1440, opened 2024-07-12
Generate Visual Studio solutions from cmake

Cyan4973:
Here is a potential follow up for #1386,
the idea is to no longer manually maintain Visual Studio solutions,
but have them generated from the `cmake` script instead.

This is achieved in this PR, by providing `generate_vs####.cmd` scripts within `build/visual/`.

The downsides are:
- It requires `cmake` to be installed on the user's system in order to successfully run the script
    + This could be mitigated by providing some pre-generated solutions as part of the repository before release 
- The list of projects generated is different from the list of projects currently provided in the manual solutions
    + This is not necessarily an issue, because it's dubious that users would be interested in the tests tools. So a reduced list of projects which are more likely to be wanted by users is actually better, to reduce confusion.
- When running `cmake` to generate Visual solutions, there are a number of warning messages and tests that feel inconsistent with Windows. Also, the generated Visual solution works fine, but the projects' names displayed in Visual look strange. It probably hints at a desirable refactoring of the `cmake` script to make the experience smoother for MS Visual.
    + `cmake` has been so far community maintained, but it seems unsustainable to rely on 3rd-party support to maintain `cmake` for an extended scope. This points at a need to take over this part of the project, which is neither simple nor free. I'm not too thrilled about it, given how complex `cmake` scripts quickly become, but it seems unavoidable.

Also: once it's clear this is the solution we want to support VS Solutions going forward, the manual ones should probably be removed, with the corresponding impact on tests and documentation.


Cyan4973:
I was initially concerned by discovering the existence of projects `ZERO_CHECK`, `PACKAGE`, `INSTALL` and `ALL_BUILD` in the generated Solution.

But after looking around, it appears to be normal practice for `cmake`, not a quirk of locally defined `CMakeLists.txt`.

So I guess it's "working as intended" ...

PR #1451, opened 2024-07-17
update changelog for v1.10

Cyan4973:


NEWS
```diff
@@ -1,11 +1,26 @@
-v1.9.5 (in dev)
-perf : Slightly faster (+5%) HC compression speed (levels 3-9)
-lib  : lz4frame states can now be reset and reused after a processing error (described by @QrczakMK)
+v1.10.0
+cli  : multithreading compression support: improves speed by X times threads allocated
```
ottokang, line 2: Do not have multithreading decompression ?

Cyan4973: > Do not have multithreading decompression ?

nope, only overlapping of i/o with decompression.
That's because, so far, lz4 decompression is faster than any i/o I can throw at it.

PR #1468, opened 2024-07-22
Fix stack alignment of _start() in freestanding test

### LINKED ISSUE #1466: Test '4/17 freestanding' fails with SIGSEGV when built with -O2 (which enables -ftree-vectorize)
**Describe the bug**
During an update to 1.10.0 Gentoo wired up the test suite and noticed a test failure in '4/17 freestanding':
https://bugs.gentoo.org/936480

This turned out to be reproducible across several build/toolchain configurations.

**Expected behavior**
Test should succeed

**To Reproduce**
1. build 1.10.0 with gcc 13/14 and at least `-O2`, which enables `-ftree-vectorize` by default
2. run tests

 - OS: Gentoo Linux
 - Compiler: gcc 13/14
 - Build System: meson

**Additional context**
- The problem was reproduced on several systems
- Build flag reduction shows that building `freestanding.c` (which includes the lz4 code as .c) with `-O1` or more precisely `-fno-tree-vectorize` makes the test work. This indicates that something is off with gcc's vectorizer passes, as far as the test in question is concerned.
- In addition building `freestanding.c` with clang (18.1.8) fails with several undefined references to `__stack_chk_fail` due to Gentoo's clang with default-enabled stack protector. Building with `-fno-stack-protector` makes the test pass _regardless of optimization level_, i.e. everything passes with `-O3`.

hhoffstaette:
When the freestanding test is built with any kind of optimization that enables vectorized loops, special care must be taken to align the stack for _start() at a 16-byte boundary.

Fixes: #1466

### DIFF
--- a/tests/freestanding.c
+++ b/tests/freestanding.c
@@ -225,7 +225,7 @@ EXTERN_C int memcmp(const void *s1, const void *s2, size_t n) {
 
 
 //
-EXTERN_C void _start(void) {
+EXTERN_C void __attribute__((force_align_arg_pointer)) _start(void) {
     test();
     MY_exit(0);
 }


Cyan4973:
cc @t-mat

t-mat:
Thanks for the patch!  :+1: for this change.

We may need more tests for this issue.  But they should be another one.  (Gentoo environment, recent versions of compiler, etc)

hhoffstaette:
> Thanks for the patch! 👍 for this change.

Excellent!

> We may need more tests for this issue. But they should be another one. (Gentoo environment, recent versions of compiler, etc)

I believe one reason this wasn't found earlier is that the Makefiles are broken, CMake does not seem to build the tests at all, and meson for some reason has [this default option](https://github.com/lz4/lz4/blob/dev/build/meson/meson/tests/meson.build#L54), which the Gentoo build probably overrides. So to reproduce it should be enough to remove the `optimization=1` bit and use a recent gcc (>=12 enabled vectorization in -O2 if I remember correctly).

Cyan4973 (approved):

thesamesam:
That said, I'm confident one of us can look into adding Gentoo to the CI if interested.

PR #1469, opened 2024-07-23
Fix building freestanding test with clang & enabled stack protector

hhoffstaette:
The freestanding test is special as it requires reimplementation of runtime/toolchain functionality. When building with clang where the compiler might be configured to enable the stack-protector by default, building the test fails due to missing __stack_chk_fail symbols; gcc automatically disables the stack-protector when `-ffreestanding` is enabled.
Thefore disable stack protection specifically for this test.

### DIFF
--- a/build/meson/meson/tests/meson.build
+++ b/build/meson/meson/tests/meson.build
@@ -47,7 +47,7 @@ test_exes = {
   },
   'freestanding': {
     'sources': files(lz4_source_root / 'tests/freestanding.c'),
-    'c_args': ['-ffreestanding', '-Wno-unused-parameter', '-Wno-declaration-after-statement'],
+    'c_args': ['-ffreestanding', '-fno-stack-protector', '-Wno-unused-parameter', '-Wno-declaration-after-statement'],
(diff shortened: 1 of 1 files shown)


hhoffstaette:
This is a followup to the clang-related problem mentioned in #1466 and fixes building & running the full test suite on Gentoo with clang. :rocket:

Cyan4973:
cc @t-mat

hhoffstaette:
Just to be clear: this simply makes existing implicit gcc behaviour explicit and consistent across compilers.
```
$diff <(gcc -Q --help=optimize | grep stack-protector) <(gcc -ffreestanding -Q --help=optimize | grep stack-protector)
4c4
<   -fstack-protector-strong    		[enabled]
---
>   -fstack-protector-strong    		[disabled]
```

t-mat:
LGTM!
Sorry for the delay.

@Cyan4973 Since I'm moving to new house, my reply may be delayed.  🙏

Cyan4973 (approved):

PR #1472, opened 2024-07-24
Fix file size check in LZ4F_readOpen

axiqia:
When the size of the opened compressed file is less than LZ4F_HEADER_SIZE_MAX (19) bytes, LZ4F_readOpen() will return an io_read error code. The minimum compressed file size is LZ4F_HEADER_SIZE_MIN + LZ4F_ENDMARK_SIZE.

Fix to check the file size with minimum size.

Test with data [a.txt]( https://corpus.canterbury.ac.nz/descriptions/artificl/a.html)
Before this patch:
./fileCompress a.txt
inp = [a.txt]
lz4 = [a.txt.lz4]
dec = [a.txt.lz4.dec]
compress : a.txt -> a.txt.lz4
a.txt: 1 → 16 bytes, 1600.0%
compress : done
decompress : a.txt.lz4 -> a.txt.lz4.dec
LZ4F_readOpen error: ERROR_io_read
compression error: Unspecified error code

After this patch:
$./fileCompress a.txt
inp = [a.txt]
lz4 = [a.txt.lz4]
dec = [a.txt.lz4.dec]
compress : a.txt -> a.txt.lz4
a.txt: 1 → 16 bytes, 1600.0%
compress : done
decompress : a.txt.lz4 -> a.txt.lz4.dec
decompress : done
verify : a.txt <-> a.txt.lz4.dec
verify : OK

### DIFF
--- a/lib/lz4file.c
+++ b/lib/lz4file.c
@@ -93,7 +93,7 @@ LZ4F_errorCode_t LZ4F_readOpen(LZ4_readFile_t** lz4fRead, FILE* fp)
 
   (*lz4fRead)->fp = fp;
   consumedSize = fread(buf, 1, sizeof(buf), (*lz4fRead)->fp);
-  if (consumedSize != sizeof(buf)) {
+  if (consumedSize < LZ4F_HEADER_SIZE_MIN + LZ4F_ENDMARK_SIZE) {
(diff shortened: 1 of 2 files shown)


Cyan4973 (approved):
Thanks @axiqia , that's a good fix

PR #1492, opened 2024-08-16
PoC: Add support for system xxHash

kou:
In general, packaging systems such as .deb want to use packaged dependencies instead of bundled dependencies. LZ4 bundles xxHash but system xxHash is preferred.

Changes:

* Move `lib/xxhash.[ch]` to `lib/xxhash/xxhash.[ch]` and use `#include "xxhash/xxhash.h"` instead of `#include "xxhash.h"` to distinct bundled xxHash and system xxHash
* Add `LZ4_USE_SYSTEM_XXHASH` macro to use system xxHash
* Use `#include "..."` instead of `-I...` to use our header files
* Only CMake build supports system xxHash by `-DLZ4_USE_SYSTEM_XXHASH` for now


kou:
What do you think about this proposal?
This is not completed yet but is there a chance that this proposal is accepted?

PR #1508, opened 2024-09-07
minor: fix test name

Cyan4973:
obvious insertion mishap


Cyan4973:
The VS2022 tests seem blocked in an infinite loop at the exact same place where other sessions fail (`fullbench` decompression test 12). So no change on this front (as expected).

PR #1510, opened 2024-09-12
CMakeLists.txt: respect global CMAKE_POSITION_INDEPENDENT_CODE flag

ds-sloth:
This is a minor change to the CMake build script to facilitate integrating this library as a (possibly nested) dependency of a project which disables PIC.

I'm aware that upstream code would be capable of setting `LZ4_POSITION_INDEPENDENT_LIB`, but this solution is the most natural to allow a top-level project to configure the option for many dependent libraries at once without needing to maintain patches for each one.

Thanks for your consideration!

(Sorry for the force push. I only realized that the patch should be based on the dev branch after seeing GitHub's UI.)

### DIFF
--- a/build/cmake/CMakeLists.txt
+++ b/build/cmake/CMakeLists.txt
@@ -85,11 +85,13 @@ list(APPEND LZ4_CLI_SOURCES ${LZ4_SOURCES}) # LZ4_CLI always use liblz4 sources 
 # Whether to use position independent code for the static library.  If
 # we're building a shared library this is ignored and PIC is always
 # used.
-if(NOT DEFINED CMAKE_POSITION_INDEPENDENT_CODE)
-  option(LZ4_POSITION_INDEPENDENT_LIB "Use position independent code for static library (if applicable)" ON)
+if(NOT DEFINED CMAKE_POSITION_INDEPENDENT_CODE OR CMAKE_POSITION_INDEPENDENT_CODE)
+  set(LZ4_POSITION_INDEPENDENT_LIB_DEFAULT ON)
 else()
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):
Looks good to me !

ds-sloth:
While looking back at these changes on my side, I realized that this doesn't fully account for the case where somebody *enables* the `CMAKE_POSITION_INDEPENDENT_CODE` flag.

What do you think the behavior should be in that case: force `LZ4_POSITION_INDEPENDENT_LIB` to be on, or just enable the option?

Cyan4973:
Here is a proposed logic: 

- When `CMAKE_POSITION_INDEPENDENT_CODE` is _not_ set :
  keep the current (before patch) behavior.

- When `CMAKE_POSITION_INDEPENDENT_CODE` is set :
  make it the default starting value of `LZ4_POSITION_INDEPENDENT_LIB`.

- When `LZ4_POSITION_INDEPENDENT_LIB` is explicitly set:
  let it take over the decision, irrespective of `CMAKE_POSITION_INDEPENDENT_CODE`.

ds-sloth:
That does sound like the desired behavior to me also.

Is it acceptable that the obvious solution (which I just committed) results in the default value of `LZ4_POSITION_INDEPENDENT_LIB` entering cache and persisting even if `CMAKE_POSITION_INDEPENDENT_CODE` is changed by the user in subsequent configures?

Cyan4973 (approved):

Cyan4973:
> Is it acceptable that the obvious solution (which I just committed) results in the default value of LZ4_POSITION_INDEPENDENT_LIB entering cache and persisting even if CMAKE_POSITION_INDEPENDENT_CODE is changed by the user in subsequent configures?

I have no idea. It just does not look problematic to me.

PR #1522, opened 2024-10-06
build/meson: avoid passing long options to `ln`

bgilbert:
On macOS, `ln` only supports short options.

Also update directory path in README.

### DIFF
--- a/build/meson/README.md
+++ b/build/meson/README.md
@@ -10,7 +10,7 @@ This Meson build system is provided with no guarantee.
 
 ## How to build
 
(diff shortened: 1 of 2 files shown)


t-mat:
Nice.  `ln -s -f` is POSIX compatible switches.  But long name options aren't.

- https://pubs.opengroup.org/onlinepubs/9699919799/utilities/ln.html

We may need to add test for POSIX conformance though.

Cyan4973 (approved):

PR #1523, opened 2024-10-06
build: use `uname -m` to detect processor

bgilbert:
Ubuntu [patches](https://git.launchpad.net/ubuntu/+source/coreutils/tree/debian/patches/80_fedora_sysinfo.patch?h=ubuntu/oracular) `uname -p` to return the processor type, and Fedora also did [before Fedora 38](https://src.fedoraproject.org/rpms/coreutils/c/cd953e11dd78cada371f0389171cea671949141b), but in general, `uname -p` returns `unknown`.  Use `uname -m` instead.

Fixes build error:

    /usr/bin/ld: /tmp/ccgv9E4e.o: in function `_start':
    freestanding.c:(.text+0x6255a): multiple definition of `_start'; /usr/lib/gcc/x86_64-redhat-linux/14/../../../../lib64/crt1.o:(.text+0x0): first defined here
    collect2: error: ld returned 1 exit status

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -367,7 +367,7 @@ test-decompress-partial : decompress-partial decompress-partial-usingDict
 # freestanding test only for Linux x86_64
 #-----------------------------------------------------------------------------
 UNAME_S ?= $(if $(filter Windows_NT,$(OS)),Windows,$(shell uname -s))
-UNAME_P ?= $(if $(filter Windows_NT,$(OS)),Unknown,$(shell uname -p))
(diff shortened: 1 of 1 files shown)


t-mat:
Good catch.   `uname -m` is portable / POSIX compatible.  But `uname -p` is not.

- https://www.man7.org/linux/man-pages/man1/uname.1.html
- https://pubs.opengroup.org/onlinepubs/9699919799/utilities/uname.html

Cyan4973 (approved):

PR #1524, opened 2024-10-06
build/meson: force `LZ4_DEBUG=0` for `tests/freestanding`

bgilbert:
Match the Makefile, which never sets `LZ4_DEBUG` for this test.  If `LZ4_DEBUG` is non-zero, `lib/lz4.c` `#include`s `<assert.h>`, the musl version of which has a [conflicting definition](https://www.openwall.com/lists/musl/2019/03/04/6) for `__assert_fail`.

Fixes build failure:

    ../../../tests/freestanding.c:173:6: error: conflicting types for '__assert_fail'; have 'void(const char *, const char *, unsigned int,  const char *)'
      173 | void __assert_fail(const char * assertion, const char * file, unsigned int line, const char * function) {
          |      ^~~~~~~~~~~~~
    In file included from ../../../tests/../lib/lz4.c:270,
                     from ../../../tests/freestanding.c:49:
    /usr/include/assert.h:19:16: note: previous declaration of '__assert_fail' with type 'void(const char *, const char *, int,  const char *)'
       19 | _Noreturn void __assert_fail (const char *, const char *, int, const char *);
          |                ^~~~~~~~~~~~~

### DIFF
--- a/build/meson/meson/tests/meson.build
+++ b/build/meson/meson/tests/meson.build
@@ -47,7 +47,7 @@ test_exes = {
   },
   'freestanding': {
     'sources': files(lz4_source_root / 'tests/freestanding.c'),
-    'c_args': ['-ffreestanding', '-fno-stack-protector', '-Wno-unused-parameter', '-Wno-declaration-after-statement'],
+    'c_args': ['-ffreestanding', '-fno-stack-protector', '-Wno-unused-parameter', '-Wno-declaration-after-statement', '-DLZ4_DEBUG=0'],
     'link_args': ['-nostdlib'],
     'build': cc.get_id() in ['gcc', 'clang'] and
       host_machine.system() == 'linux' and host_machine.cpu_family() == 'x86_64',


Cyan4973 (approved):

Cyan4973:
Great debugging @bgilbert !

Cyan4973:
Regarding tests:
our current CI runs the `freestanding` test with `make`, but not with `meson`.
and I was wondering if this is something that could be added.

Cyan4973:
Rectification:
there is a test named `freestanding` which is [run during `meson` CI test ](https://github.com/lz4/lz4/actions/runs/11201357856/job/31136190272?pr=1524#step:7:47):
```
 4/17 datagen                        OK               0.02s
 5/17 freestanding                   OK               0.02s
 6/17 lz4-contentSize                OK               4.15s
 ```
 
 I'm not sure what it does exactly, maybe it just compiles `tests/freestanding`.
Moreover, it was obviously working fine so far, therefore it did not underline the need for `LZ4_DEBUG=0`.
 Though maybe triggering this specific issue requires `musl` too, which is not standard.

bgilbert:
> Rectification: there is a test named `freestanding` which is [run during `meson` CI test ](https://github.com/lz4/lz4/actions/runs/11201357856/job/31136190272?pr=1524#step:7:47):
> 
> [...]
> 
> I'm not sure what it does exactly, maybe it just compiles `tests/freestanding`.

Meson compiles it and runs it.

> Moreover, it was obviously working fine so far, therefore it did not underline the need for `LZ4_DEBUG=0`. Though maybe triggering this specific issue requires `musl` too, which is not standard.

That's right, the problem only occurs with musl.

PR #1544, opened 2024-11-18
Fix examples dictionary random access

### LINKED ISSUE #1542: examples/dictionaryRandomAccess.c fails when given file is smaller than 1024 bytes
**Describe the bug**
examples/dictionaryRandomAccess.c doesn't work with a small file which is less than 1024 bytes (`DICTIONARY_BYTES`).

**To Reproduce**

```
make

# OK
./examples/dictionaryRandomAccess README.md README.md 0 10

# NG (for small files)
./examples/dictionaryRandomAccess .gitattributes .gitattributes 0 10
```

**Additional context**

I'll write a PR for this issue.  But the following patch may work.

https://github.com/lz4/lz4/blob/v1.10.0/examples/dictionaryRandomAccess.c#L254

```C
-       test_decompress(outFp, inpFp, dict, DICTIONARY_BYTES, offset, length);
+       test_decompress(outFp, inpFp, dict, dictSize, offset, length);
```

t-mat:
This PR adds a new test to reproduce (and will prevent) the issue, and fixes it.

- 184ffbd fails the test (as intended)
- 855a097 pass the test

<details>

<summary>How to test this PR</summary>

```sh
cd
git clone https://github.com/lz4/lz4.git lz4-pr-1544
cd lz4-pr-1544
git pull origin pull/1544/head

# NG
git checkout 184ffbd
make -C examples clean test
#
# === Dictionary Random Access ===
# ...
# verify : .gitignore <-> .gitignore.lz4s-1024.dec
# verify : NG
#


# OK
git checkout 855a097
make -C examples clean test
#
# === Dictionary Random Access ===
# ...
# verify : .gitignore <-> .gitignore.lz4s-1024.dec
# verify : OK
#
```

</details>

### DIFF
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -35,6 +35,7 @@ WFLAGS     = -std=gnu99 -Wall -Wextra -Wundef -Wshadow -Wcast-align -Wstrict-pro
 CFLAGS     = $(WFLAGS) -O2 $(USERCFLAGS)
 
 TESTFILE   = Makefile
+TESTFILE_SMALL= .gitignore
 SLIBLZ4   := $(LIBDIR)/liblz4.a
 LZ4DIR     = ../programs
 LZ4        = $(LZ4DIR)/lz4
@@ -80,6 +81,7 @@ test : all $(LZ4)
 	./blockStreaming_lineByLine$(EXT) $(TESTFILE)
 	@echo "\n=== Dictionary Random Access ==="
 	./dictionaryRandomAccess$(EXT) $(TESTFILE) $(TESTFILE) 1100 1400
+	./dictionaryRandomAccess$(EXT) $(TESTFILE_SMALL) $(TESTFILE_SMALL) 0 32
 	@echo "\n=== Frame compression ==="
 	./frameCompress$(EXT) $(TESTFILE)
 	$(LZ4) -vt $(TESTFILE).lz4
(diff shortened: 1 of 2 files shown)


t-mat:
{8905CC2D-7A08-4CC4-9D3B-0C5E51E34C27}

Ugh, `CFLAGS='-mx32'` test finally fails with `gcc-8` and `gcc-9`.  We have no compiler which passes `-mx32` test.  There're no other [`x32: 'true'`](https://github.com/lz4/lz4/blob/v1.10.0/.github/workflows/ci.yml#L54-L55) compilers :(
We may need serious investigation or just abandon them for now.  Since compiler itself is same, I suspect recent version of `gcc-multilib` is not compatible with x32 though.

Cyan4973:
> Ugh, `CFLAGS='-mx32'` test finally fails with `gcc-8` and `gcc-9`. We have no compiler which passes `-mx32` test. There're no other [`x32: 'true'`](https://github.com/lz4/lz4/blob/v1.10.0/.github/workflows/ci.yml#L54-L55) compilers :(

Yes, it's a recent development, it seems that `x32` is no longer supported by github actions at this point.
I don't expect a reverse of fortune,
so it is probably simpler to just drop this requirement, at least stop testing it on Github Actions.

Maybe other CI could cover that blind spot if deemed important enough.

Cyan4973 (approved):

PR #1564, opened 2025-01-13
fixed minor static analyzer warning in fileCompress example

Cyan4973:
In the process, discovered a test that was no longer firing in case of error.
Fixing it uncovered a bug that has been silently around since a few PR : 
flushing a single byte with `LZ4F_flush()` would impact the internal indexing scheme within the `lz4hc` state.
So fixing it too, and in the process improved traces a bit, which ended up being a big lion share of this PR.
After that, found another (test) bug, which might have been drowned by the previous issue, and fixed it too.

### DIFF
--- a/examples/fileCompress.c
+++ b/examples/fileCompress.c
@@ -28,7 +28,7 @@ static size_t get_file_size(char *filename)
         return 0;
     }
 
-    return statbuf.st_size;
+    return (size_t)statbuf.st_size;
 }
 
 static int compress_file(FILE* f_in, FILE* f_out)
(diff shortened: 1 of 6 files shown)


Cyan4973:
The LZ4F multi-segments compression test seems broken.

Unfortunately, the test itself was failing to report the issue as a clear error signal (just outputting some logs), so it's difficult to tell for how long this issue has been present.

It's unclear at this point if the issue is in the test, or if it is in the `LZ4F` library.

_edit_ : it seems the issue isn't present in latest release `v1.10`, so it has been triggered more recently.
_edit 2_: apparently introduced in 9b9e03ff0929850ea0342fb81391974e60e31069 , but so far it's difficult to see the relation between the commit's content and the test issue.
Btw, the test issue can be observed by running `frametest`, which will output a bunch of `locateBuffDiff: looking for error position` in the log. In `dev`, it just burdens the log, but in this `PR`, it properly triggers an error.
_edit 3_: found it, it's a bug triggered by `LZ4F_flush()`, when trying to flush a single byte into an empty buffer. 
It's btw not clear if this is a good idea to do that (write into a limited size buffer) since it may break some compression for some incompressible blocks, and I'm not sure about the state of the system after that ....

Cyan4973:
All tests fixed.
This should restore our clean green CI signal for future PRs.

PR #1565, opened 2025-01-14
Update util.h

zkh8227:
double-free risk in UTIL_realloc :
The UTIL_realloc function is used to resize a memory block allocated by malloc, calloc, or realloc. In glibc, when realloc(ptr, 0) is called, it is equivalent to free(ptr) and returns NULL. This behavior is implementation-defined, and while glibc handles it safely, it can lead to confusion and potential double-free issues if not managed correctly.

### DIFF
--- a/programs/util.h
+++ b/programs/util.h
@@ -191,7 +191,7 @@ UTIL_STATIC void* UTIL_realloc(void* ptr, size_t size)
 {
     void* const newptr = realloc(ptr, size);
     if (newptr) return newptr;
-    free(ptr);
+    if (size!=0) free(ptr);
     return NULL;
 }
 


t-mat:
Hi @zeux, thanks for the PR!

Here is a note for future contributors:
In C23, the behavior of `realloc(ptr, 0)` is officially undefined.

See also: [WG14 n2464: Zero-size Reallocations are Undefined Behavior](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2464.pdf)

It seems that `gcc-14` doesn't report anything about it though

```C
// realloc-0-0.c
// gcc-14 -std=c23 -Wall -Wextra -Wundef -Wshadow -Wcast-align -Werror -Wpedantic ./realloc-0-0.c
// # no error/warning(s)
#include <stdlib.h>
int main(int, char**) {
  return realloc(0, 0) == 0;
}
```

Therefore, it may be difficult to test this UB in our CI so far.

programs/util.h
```diff
@@ -191,7 +191,7 @@ UTIL_STATIC void* UTIL_realloc(void* ptr, size_t size)
 {
     void* const newptr = realloc(ptr, size);
     if (newptr) return newptr;
-    free(ptr);
+    if (size!=0) free(ptr);
```
Cyan4973, line 194: I believe it's preferable to do this check at the beginning the of function, and then exit early with `return NULL`.
This avoids any risk associated with `realloc(ptr, 0)`, which may or may not act, free or not the original buffer, depending on `libc` implementation, and we don't want to be dependent on a specific behavior.

t-mat: Possible fix for further discussion:
```C
UTIL_STATIC void* UTIL_realloc(void* ptr, size_t size)
{
  /* We explicitly check size != 0 since realloc(ptr, 0) is UB in C23 (see WG14 n2464).
     Though when size == 0, UTIL_realloc() frees the memory to mimic classic glibc's behavior. */
  if (size != 0) {
    void* const newptr = realloc(ptr, size);  /* Note: realloc() accepts ptr == NULL. */
    if (newptr) return newptr;
  }
  free(ptr); /* Note: free() accepts NULL */
  return NULL;
}
```

Cyan4973:
Please rebase this PR on top of `dev`,
there are errors in CI currently, which are probably not related to this PR, and therefore are expected to be fixed after the rebase.

zkh8227:
> Hi @zeux, thanks for the PR!
> 
> Here is a note for future contributors: In C23, the behavior of `realloc(ptr, 0)` is officially undefined.
> 
> See also: [WG14 n2464: Zero-size Reallocations are Undefined Behavior](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2464.pdf)
> 
> It seems that `gcc-14` doesn't report anything about it though
> 
> ```c
> // realloc-0-0.c
> // gcc-14 -std=c23 -Wall -Wextra -Wundef -Wshadow -Wcast-align -Werror -Wpedantic ./realloc-0-0.c
> // # no error/warning(s)
> #include <stdlib.h>
> int main(int, char**) {
>   return realloc(0, 0) == 0;
> }
> ```
> 
> Therefore, it may be difficult to test this UB in our CI so far.

There is indeed no problem with your realloc-0-0.c code. This is because you used 0 instead of a Buf as the first parameter. You can try the following code and test it with gcc -14.
```
#include <stdlib.h>
int main (int ,char**) {
        char buf[20];
        return realloc(buf, 0) == 0;
}
```

> Hi @zeux, thanks for the PR!
> 
> Here is a note for future contributors: In C23, the behavior of `realloc(ptr, 0)` is officially undefined.
> 
> See also: [WG14 n2464: Zero-size Reallocations are Undefined Behavior](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n2464.pdf)
> 
> It seems that `gcc-14` doesn't report anything about it though
> 
> ```c
> // realloc-0-0.c
> // gcc-14 -std=c23 -Wall -Wextra -Wundef -Wshadow -Wcast-align -Werror -Wpedantic ./realloc-0-0.c
> // # no error/warning(s)
> #include <stdlib.h>
> int main(int, char**) {
>   return realloc(0, 0) == 0;
> }
> ```
> 
> Therefore, it may be difficult to test this UB in our CI so far.

Cyan4973:
Replaced by #1577

PR #1578, opened 2025-02-25
fix FreeBSD test on Cirrus CI

Cyan4973:
underlying VM was removed, update to a new version

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -2,7 +2,7 @@ task:
   name: FreeBSD
   freebsd_instance:
     matrix:
-      image_family: freebsd-14-1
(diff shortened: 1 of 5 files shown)


Cyan4973:
ubsan tests are now failing, and that's new.
Obviously unrelated to this PR, so maybe a change of environment ?

Cyan4973:
And of course, it doesn't reproduce locally ....

Cyan4973:
I'm currently suspecting a problem in `test-lz4-list.py`.

Apparently, the test is order-sensitive during its checking stage, 
yet doesn't properly ensure corresponding order during its generation stage.

PR #1581, opened 2025-03-18
Replaced uint8_t with unsigned char in BYTE typedef for post C99/C++

KSSAB:
Uint8_t is not necessarily a typedef for char or unsigned char (or std::byte), for this reason, it can be undefined behavior to read or modify arbitrary data through it.

Ideally, we would select uint8_t anyway if it happens to alias one of the types permitted to read or modify arbitrary data, but I don't think it can be implemented without _Generic() 

The types in header are commented as not part of public api so this should not result in ABI break for those who heed the warning. There is no such warning in the .c but I would presume that is only because it is taken for granted.

### DIFF
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -301,12 +301,12 @@ static int LZ4_isAligned(const void* ptr, size_t alignment)
 #include <limits.h>
 #if defined(__cplusplus) || (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) /* C99 */)
 # include <stdint.h>
-  typedef  uint8_t BYTE;
-  typedef uint16_t U16;
-  typedef uint32_t U32;
-  typedef  int32_t S32;
-  typedef uint64_t U64;
-  typedef uintptr_t uptrval;
+  typedef unsigned char BYTE; /*uint8_t not necessarily blessed to alias arbitrary type*/
+  typedef uint16_t      U16;
+  typedef uint32_t      U32;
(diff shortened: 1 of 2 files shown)


t-mat:
I think it's practically minor issue, since we've never reporeted any issue which is caused by (non) strict definiton of byte, octet, uint8_t and uint16_t.

But I'm also realized that how LZ4 Block Format and Frame Format is defined (and be well).

LZ4 Block Format
- ture: "LZ4 is an LZ77-type compressor with a fixed byte-oriented encoding format."
- ?: LZ4 Block Format depends "byte".  But it doesn't specify what "byte" is.  It may (not) be octet (8-bits).
- Also the block format document implcitly expects that "consequential 2 bytes of little endian bytes = single 16-bits little endian value"

`unsigned char`
- true: `unsigned char` is always "byte".
- true: It is always true that `sizeof(unsigned char) == 1`.
- false: `CHAR_BIT` in `limits.h` may not equal to `8`.

`std::byte`
- true: `std::byte` is always "byte".
- true: It is always true that `sizeof(std::byte) == 1`.
- false: `CHAR_BIT` in `limits.h` may not equal to `8`.

`uint8_t`
- true: Number of bits of `uint8_t` is always exactly 8. uint8_t is octet.
- true: little endian `uint8_t, uint8_t` becomes little endian `uint16_t`

Cyan4973 (approved):
It's probably not changing anything much,
but I get the argument about type aliasing,
so I'm fine with the proposed patch.

PR #1586, opened 2025-03-25
fix(segv): add  null pointer assertions for file pointers in example files

louislafosse:
I'm using some of the binaries in `./examples` for some unit tests and noticed that most of them don't check with assertions when opening a file descriptor for the input/output files.
I have also fixed the `INT32_MAX` value for the offset and length to prevent integer overflow.


Cyan4973 (approved):

Cyan4973 (approved):
Thanks Louis !

PR #1593, opened 2025-03-31
fix(null) : improve error handlings when passing a null pointer to some functions from lz4frame

louislafosse:
Hi,
Notice some unhandled null pointer args that leads to segv


louislafosse:
any news ?

lib/lz4frame.c
```diff
@@ -1488,6 +1495,10 @@ LZ4F_errorCode_t LZ4F_getFrameInfo(LZ4F_dctx* dctx,
                                    LZ4F_frameInfo_t* frameInfoPtr,
                              const void* srcBuffer, size_t* srcSizePtr)
 {
+    RETURN_ERROR_IF(dctx == NULL, parameter_null);
```
Cyan4973: This is the only line whose benefits still merit discussion; I’m satisfied with every other proposed change.

Why could this line be debatable? Because it touches on API consistency.

Most functions in this API follow the pattern `result function(object*, params)`, where `object*` is a pointer to the (mostly opaque) object being operated on. By convention—common in C codebases that emulate OOP—the caller must supply a valid, non-NULL pointer; passing `NULL` is considered undefined behavior (UB).

Adding a `NULL` check inside one of these functions is not wrong; the function would simply return a defined error instead of triggering UB. The issue is consistency: if only a few functions treat `object* == NULL` gracefully while the rest do not, users have to memorize arbitrary exceptions.

Some symbols intentionally support `NULL` — `object_free(object*)`, like `free(NULL)`, do accepts `NULL` — but such cases are rare, generally expected and documented.

Here, letting `LZ4F_getFrameInfo()` accept a `NULL` `dctx` and return an error is harmless in isolation. Yet guaranteeing the same behavior across all related functions may be difficult, if not nonsensical for some calls. Exploring that broader design space takes time, so we usually deal with it separately, as a dedicated topic, rather than blend it with other modifications and risk blocking the other already approved changes.

One way to emphasize that, at this point, the condition `dctx != NULL` must be true, is to add an `assert()`. This is consistent with the definition of UB, aka a contract that must not be violated.

Cyan4973:
Hi @louislafosse ,
I know you have been waiting a while for this review, and I'm sorry for the delay.
The PR is pretty good, almost entirely validated,
and there is just an opened discussion on one of the modified lines worth closing.

lib/lz4frame.c
louislafosse: Hello Yann,
Hope you are well !
It's Ok for me, I forget that this codebase emulate OOP and in fact it will be inconsistent to do accordingly.
As you suggest, I'm gonna replace it by an assert.
I also understand that the code is now too old to rewrite all the parameter passages in a safe way (because most of the time it's done by wrappers around the library and it may imply modifications on their sides).

louislafosse: I have amended the changes to the first commit.

Cyan4973 (approved):

Cyan4973:
Let it be registered clearly that I dispute the characterization of this patch as a `CVE fix`.
Especially, the current statement:

> LZ4 through 1.10.0 allows attackers to cause a denial of service (application crash) or possibly have unspecified other impact when the application processes untrusted LZ4 frames.

is just completely false.

This patch does not impact the code path of LZ4 frame at any moment, not even close.

This patch changes a narrow contract (the ***out*** parameter `frameInfoPtr` of function `LZ4F_getFrameInfo()` must not be `NULL`, which is both obvious *and* documented) into an explicit input sanitization step, which is possible in this case because `LZ4_getFrameInfo()` can return an error signal. But fundamentally, this is equivalent to writing `memcpy(NULL, src, size);` => segfault => where is the CVE against `memcpy()` ?
And more importantly, such case is a *bad API usage*, and in no way related to the consumption of LZ4 frames.

louislafosse:
Hi Yann,

Thanks for taking the time to share your perspective and for clarifying how you see the patch in relation to the library’s contract.
I completely understand your point that this change doesn’t affect the LZ4 frame parsing path and that the API contract already specifies non-null parameters.

My intent in reporting and submitting this fix was not to overstate its impact or cause unnecessary noise, but rather to address what I saw as a potential robustness issue specifically, that certain calling contexts (for example, wrappers or fuzzers) could end up passing unexpected NULL pointers and trigger a crash.
From a defensive programming and safety perspective, explicitly validating parameters like this helps make the library a bit more resilient, even if the responsibility lies primarily with the caller.

I agree that the current CVE wording may overstate the impact. It was meant to document an improvement in input validation, not to imply a remotely exploitable issue or any change to the frame parsing path.

From a procedural standpoint, this type of input validation improvement does fit the CVE methodology specifically under the “denial of service via input validation flaw” category, where untrusted input or misuse of exposed API parameters can result in abnormal termination.
There are many precedents of similar cases being assigned CVEs, such as CVE-2024-7006 (libtiff) and CVE-2016-10087 (libpng), both of which involved NULL pointer dereferences leading to local crashes when invalid input is provided.

PR #1595, opened 2025-04-07
Add Functions that Report LZ4F Contexts' Memory Footprints

felixhandte:


Cyan4973:
The CI test issue has been fixed in #1602.
If you rebase your PR on top of `dev`, it should fix the test errors.

Cyan4973:
Any plan to complete this PR ?
It seems it mostly deserves a `rebase` to get back to clean CI status,
and then fix the rest from there, if there is any remaining issue.

Cyan4973 (approved):
Interface and implementation are both clear and well implemented.

All the maintenance complexity is concentrated in the test case.
The risk here is that, if something breaks this test in the future, then understanding what the test does and why it fails, and therefore what needs to be fixed, can be a bit challenging.

I don't have a different test to offer, I just wonder if understanding the test can be made easier for an external reviewer, notably if it happens to break in the future (typically as a consequence of some (possibly erroneous) future change).

felixhandte:
@Cyan4973, I actually expect the test to be fairly robust: it tracks the actual live memory that an object manages, via the `Test_alloc_state` / `lz4f_cmem_test`, and compares that against the memory usage that these new APIs report. In the event that these diverge, it should be high signal that either:

1. You don't intend to change the allocations owned by the context objects, and your changes are leaking memory, which this is detecting.
2. You did intend to change the allocations owned by the context objects, and you need to update the size-reporting implementations to reflect those changes.

In both cases I think it's pretty clear. You'll get a message: `"%llu allocated in cctx but it says its size is %llu.\n"`.

PR #1601, opened 2025-04-20
Update cmake minimum required version

GeorgH93:
Update cmake minimum required version to work with CMake v4 and shut up about the deprecation warning (for cmake_minimum_required < 3.10) on CMake v3.31+

### DIFF
--- a/build/cmake/CMakeLists.txt
+++ b/build/cmake/CMakeLists.txt
@@ -8,7 +8,7 @@
 # For details, see <http://creativecommons.org/publicdomain/zero/1.0/>.
 
 
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.5...3.16)
 
 set(LZ4_TOP_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../..")
 


GeorgH93:
@Cyan4973 looks like [Github has turned off their Ubuntu 20.04 runners](https://github.com/actions/runner-images/issues/11101) a week ago, and that's why some of the tests aren't running. Could you please look into that?

Cyan4973:
Thanks for the notification.
The CI test issue has been fixed in #1602.
If you rebase your PR on top of `dev`, it should fix the test errors currently experienced.

GeorgH93:
> Thanks for the notification. The CI test issue has been fixed in #1602. If you rebase your PR on top of `dev`, it should fix the test errors currently experienced.

Yes, all the tests work now.

Cyan4973:
On the topic of the PR itself:
the text specifies :

> work with CMake v4 and shut up about the deprecation warning (for cmake_minimum_required < 3.10) on CMake v3.31+

However, the code update `cmake_minimum_required(VERSION 3.5)` into:

```
cmake_minimum_required(VERSION 3.5...3.16)
```

which keeps the minimum at `3.5` but make the policy fluctuate up to `3.16`.

Somehow, the PR impact feels different from its description.

Why not a more direct : 
```
cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
```

GeorgH93:
That would block cmake versions older than 3.10 from using it altogether, since the project isn't using any cmake features requiring a newer version, I thought that might be an undesirable side effect.
The 3.16 comes from our internal testing (since it's our go-to legacy test version), so I know that the policy changes up to that version don't cause issues with this lib.

It's up to you, if you preferrer to hard limit to 3.10 we can do that too.

Cyan4973:
OK, if I understand correctly,
with `cmake_minimum_required(VERSION 3.5...3.16)`,
a modern version of `cmake`, say for example `4.0`,
will behave as if it was `3.16`,
i.e. if a there is a change of semantic or default behavior between `3.16` and `4.0`, it will use the `3.16` one.
Importantly, because it will not have to emulate `3.5` behavior, it will stop complaining about the old version number.

Then, you selected `3.16` specifically because it's the last version known to work fine to you, since it's the one used in your production environment.

Question: if the script is used in another setting with a more recent version of `cmake` and works fine there, then it's ground to increase the second number ?
For example, my desktop provides `cmake 3.28`, and it works fine there without any concern.
Hence it could be ground to increase the second number to `3.28` ?

GeorgH93:
Pretty much yes.

We are using 3.16 at work to verify our builds (and our dependencies) to work with 3.16 behaviour, so I know that 3.16 works for this lib. Newer versions would probably work fine too, the cmake script isn't really doing anything crazy, but I haven't tested that for anything but Linux static lib.

If you aren't seeing any deprecation warnings, in any code path and platform, you could also bump it up further.
On my Arch system where I am running CMake 4.0, which prompted me to make that PR, I am only building the static lib, so I have no clue if there might be something hidden behind the other options, so I went with what I know to be a safe behavior version.

Cyan4973 (approved):

PR #1634, opened 2025-07-21
lz4_Frame_format.md: document kernel legacy last block

Artoria2e5:
Since https://github.com/torvalds/linux/commit/2c484419efc09e7234c667aa72698cb79ba8d8ed, the Linux Kernel is able to discern the end of a legacy LZ4 frame by the presence of four 0 bytes, parsed as a B.CSize of 0 in the decoder. This is originally intended to make it mix well with grub initrd concatenation behavior (which uses a clause in the initramfs format about allowing arbitrary zero paddings), but I think it is a pretty good idea for any application limited to the legacy format in general.

### DIFF
--- a/doc/lz4_Frame_format.md
+++ b/doc/lz4_Frame_format.md
@@ -358,7 +358,9 @@ Main characteristics of the legacy format :
 - All blocks are always compressed, even when compression is detrimental.
 - The last block is detected either because
   it is followed by the “EOF” (End of File) mark,
-  or because it is followed by a known Frame Magic Number.
+  or because it is followed by a known Frame Magic Number.<br>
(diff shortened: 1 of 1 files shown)


Cyan4973 (approved):

Artoria2e5:
Yay! But now I’m second-guessing myself: would it be better to instead document it not just as a Linux kernel behavior, but also as a generally applicable recommendation (i.e. add a paragraph to the EndMark section)? That would be an actual (albeit minor) format change, which I didn’t want to deal with…

Cyan4973:
I would have folded the specific Linux kernel behavior within this sentence :
"or because it is followed by a known Magic Number."

with _known_ being effectively context dependent.

PR #1639, opened 2025-08-07
Makefile: support UNALIGNED_ACCESS_SUPPORTED for RISC-V

Polaris-911:
**Description**
When building on RISC-V, allow users to opt in to unaligned access by setting the UNALIGNED_ACCESS_SUPPORTED This appends
-DUNALIGNED_ACCESS_SUPPORTED to CFLAGS, which in turn enables LZ4_FORCE_MEMORY_ACCESS=2 under RISC-V only.

Usage:
    make UNALIGNED_ACCESS_SUPPORTED=1

**Reason for modification**
During testing, we found that direct access to unaligned memory is currently the best method on RISC-V, similar to the method on ARMv6. However, not all RISC-V servers support unaligned memory, and the GCC compiler does not have a macro to distinguish whether unaligned memory is supported. Therefore, we chose to add a new macro to manually allow users to enable direct access on RISC-V servers that support unaligned memory to achieve performance improvement.

**Testing**
Testing of this modification on a RISC-V server found that the compression rate increased by 30%. The test data is as follows:
Tool: lzbench
Unaligned_Access_Supported disabled:
```shell
lzbench 2.1 | GCC 13.2.1 | 64-bit Linux |

lz4 1.10.0               96.8 MB/s   206 MB/s   100881340  47.60 ../silesia.tar
[Params] cIters=5 dIters=5 cTime=0.0 dTime=0.0 chunkSize=1706MB cSpeed=0MB
```
Unaligned_Access_Supported enabled:
```shell
lzbench 2.1 | GCC 13.2.1 | 64-bit Linux |

lz4 1.10.0                124 MB/s   209 MB/s   100881340  47.60 ../silesia.tar
[Params] cIters=5 dIters=5 cTime=0.0 dTime=0.0 chunkSize=1706MB cSpeed=0MB
```

### DIFF
--- a/.github/workflows/build-systems.yml
+++ b/.github/workflows/build-systems.yml
@@ -23,7 +23,7 @@ jobs:
 
     steps:
     - name: Checkout
-      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
     - name: Build Win32 Debug
       # Switching to debug mode, since vs2022 optimizations seem to introduce a bug since July 2024
@@ -66,7 +66,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
-      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
     - name: Setup Python
       uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
--- a/.github/workflows/cmake-test.yml
+++ b/.github/workflows/cmake-test.yml
@@ -38,7 +38,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
 
     - name: Set up MSVC (Windows)
       if: runner.os == 'Windows'
@@ -79,7 +79,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
 
     - name: Set up CMake
       uses: jwlawson/actions-setup-cmake@v2
@@ -122,6 +122,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
     - name: Test via Makefile
       run: make clean && make cmakebuild
--- a/.github/workflows/code-quality.yml
+++ b/.github/workflows/code-quality.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
-      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
     - name: Install cppcheck
       run: |
@@ -38,7 +38,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
-      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
 
     - name: Install clang-tools
       run: |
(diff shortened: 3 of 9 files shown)


t-mat:
Hi @Polaris-911 thanks for your effort and the PR!

For future contributors, could you provide your environment and actual GCC compiler options?
I think Basic info would be enough for now.
```
cat /proc/device-tree/model
uname -a
```

Since RISC-V processors are extremely diverse, I also hope RISC-V consortium will define standard hadware feature macro or special vendor pragmas.  Dynamic dispatch may be great, but for embeded platforms, they may not be able to accept it.

Polaris-911:
Hello, my environment information is as follows
```shell
[root@fedora-riscv lzbench-master]# uname -a
Linux fedora-riscv 6.1.31 #1 SMP Thu Jun 15 01:30:00 CST 2023 riscv64 GNU/Linux
[root@fedora-riscv lzbench-master]#  cat /proc/device-tree/model
Sophgo Mango
```
During the test, I compiled lzbench with the following commands:
```shell
make DONT_BUILD_TORNADO=1 -j
```
```shell
make UNALIGNED_ACCESS_SUPPORTED=1 DONT_BUILD_TORNADO=1 -j
```
Performance has been improved by adding the option UNALIGNED_ACCESS_SUPPORTED=1

Thank you for your suggestion. I will continue to look for other better implementation methods. Under the current limited conditions, this is the most appropriate modification I can think of. It can ensure that we have a significant improvement on risc-v without affecting other architectures.

Polaris-911:
> Since RISC-V processors are extremely diverse, I also hope RISC-V consortium will define standard hadware feature macro or special vendor pragmas. Dynamic dispatch may be great, but for embeded platforms, they may not be able to accept it.

Hi @t-mat Thank you for your suggestion. Regarding the embedded system issue you're concerned about, the custom macro I modified is not enabled by default and will not affect the corresponding system. However, for RISC-V, which supports unaligned memory access, enabling this feature will significantly improve performance. I hope you can assist in reviewing this pull request, and I will actively cooperate with any suggested modifications.

Polaris-911:
Current testing indicates that this change can improve RISC-V performance. I understand the community's high priority on code quality and stability , I sincerely hope this optimization will be helpful to the project. If there are any improvements or additions to the implementation, or coding style that need improvement in the pull request, I would be happy to assist with the revision.

build/make/lz4defs.make
```diff
@@ -113,3 +113,6 @@ INSTALL_PROGRAM ?= $(INSTALL) -m 755
 INSTALL_DATA    ?= $(INSTALL) -m 644
 MAKE_DIR        ?= $(INSTALL) -d -m 755
 
+ifdef UNALIGNED_ACCESS_SUPPORTED
+  CFLAGS += -DUNALIGNED_ACCESS_SUPPORTED
```
Cyan4973: This should rather be `CPPFLAGS`.

Furthermore, for a more generic impact (beyond just `risc-v`), one could set `-DLZ4_FORCE_MEMORY_ACCESS=2` directly here. It would have the same impact for `risc-v`, but also be potentially useful for other systems. 
This scope extension seems legitimate as the `Makefile` variable `UNALIGNED_ACCESS_SUPPORTED` does not have anything related to `risc-v` in the name.

I would also prefer renaming it `LZ4_FORCE_UNALIGNED_ACCESS`, since it's not a standard variable (so `LZ4_` prefix), and what it does is indeed force the compiler to write the code using unaligned memory access, no matter what, even if this capability is in fact not supported by target system.
Finally, this new variable should be documented in the `README.md` file to be useful.

Note that the only reason for this new `Makefile` variable to exist would be to express `LZ4_FORCE_MEMORY_ACCESS=2` in a slightly more convenient way.

Cyan4973:
Support of unaligned memory access is indeed generally important for speed, so it's a good idea to enable it for `risc-v`.

Reading this PR, a few things come to mind :
 
 - I don't have a real `risc-v` system around to test, so I can only reason and make assumptions, without verification. 
     - Furthermore, given the diversity of the `risc-v` ecosystem, it's likely that conclusions regarding performance optimizations will be different depending on exact chipset and compiler.
 - The patch seems to have the same impact as `LZ4_FORCE_MEMORY_ACCESS=2`. Did you try this CPP compilation flag ?
 - We tend to discourage `LZ4_FORCE_MEMORY_ACCESS=2` mode, because it's in essence "lying" to the compiler by pretending that the pointer is aligned, while it's probably not. It can sometimes end up badly, depending on exact hardware. For example, some ARM cpus supports unaligned access in normal `scalar` mode, but not in vector `neon` mode, and such a setting will segfault. 
     - That being said, in some specific cases, it might be the only way to make the compiler generate code exploiting unaligned memory access. Hence why it's still there: discouraged, but still available if really there is no other choice.
- I'm surprised that enabling unaligned memory access only improves compression speed, but not decompression speed. I would expect such a feature to have a major impact on decompression speed performance.

Cyan4973:
A safer option seems to be:
- Declare to the compiler that your target cpu supports unaligned memory access (something like `gcc -munaligned-access`
- Combined with `CPP` flag `LZ4_FORCE_MEMORY_ACCESS=1`, which is a lot safer than `=2`, but will only trigger if the compiler agrees that the target architecture supports unaligned memory access
   - Note that, for `risc-v`, the source code will default to `LZ4_FORCE_MEMORY_ACCESS=0`, which is more likely to generate align-safe code generation. But sometimes, the compiler nonetheless manages to exploit unaligned memory access at assembly level even with the `=0` setting.

Polaris-911:
Hi @Cyan4973  Thanks for your suggestion, I implemented unaligned memory access in another way, please refer to PR：https://github.com/lz4/lz4/pull/1648

PR #1641, opened 2025-08-14
Add RVV vector optimization to xxHash 32/64-bit align functions

Dayuxiaoshui:
Add RISC-V Vector Extension (RVV) support with conditional includes and macros for compatibility
Implement RVV-optimized loops in XXH32_endian_align and XXH64_endian_align for vectorized round computations
Add fallback to scalar implementation for non-RVV builds
Include performance notes and TODO for further parallelization of accumulators


Dayuxiaoshui:
I'll continue to commit code based on this, some of which are still being tested, including some experimental tests that are still ongoing,thanks

tests/rvv_detect
t-mat, line 1: Please remove this binary executable from the PR.  Since we don't accept binaries.
Maybe adding `rvv_detect` to `.gitignore` will help avoid the same issue in the future.

NOTE for @t-mat: add another test to detect binary executables.

PR #1666, opened 2025-10-23
Improve compression speed through data preloading and reordering of search range upper bound check

BiplabRaut:
- Data is preloaded for byte matching within the inner loop in LZ4_compress_generic_validated.
    - Order of LZ4_DISTANCE_MAX check for match is changed to allow better branch predictor behavior for noDict directive.

### DIFF
--- a/build/make/lz4defs.make
+++ b/build/make/lz4defs.make
@@ -88,6 +88,11 @@ else
 POSIX_ENV = No
 endif
 
+# `AOCL_PRELOAD_BRANCH_OPT` Enable data preload for byte matching within the inner loop in 
+# LZ4_compress_generic_validated and better branch predictor behavior for noDict directive.
+
+
+USEROPTIONALFLAGS ?= -DAOCL_PRELOAD_BRANCH_OPT=1
 # Avoid symlinks when targeting Windows or building on a Windows host
 ifeq ($(WINBASED),yes)
 LN_SF = cp -p
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -58,6 +58,7 @@ SRCFILES := $(sort $(wildcard *.c))
 OBJFILES = $(SRCFILES:.c=.o)
 
 include ../build/make/lz4defs.make
+CFLAGS += $(USEROPTIONALFLAGS)
 
 # default is defined as the first target in Makefile, hence before multiconf.make
 .PHONY: default
--- a/lib/lz4.c
+++ b/lib/lz4.c
@@ -1045,6 +1045,80 @@ LZ4_FORCE_INLINE int LZ4_compress_generic_validated(
             const BYTE* forwardIp = ip;
             int step = 1;
             int searchMatchNb = acceleration << LZ4_skipTrigger;
+#if defined(AOCL_PRELOAD_BRANCH_OPT) && (AOCL_PRELOAD_BRANCH_OPT == 1)
+            U32 ipData;
+
+            do {
+                U32 const h = forwardH;
+                U32 const current = (U32)(forwardIp - base);
+                U32 matchIndex = LZ4_getIndexOnHash(h, cctx->hashTable, tableType);
+                U32 matchData;
+
+                assert(matchIndex <= current);
+                assert(forwardIp - base < (ptrdiff_t)(2 GB - 1));
+                ip = forwardIp;
+                forwardIp += step;
+                step = (searchMatchNb++ >> LZ4_skipTrigger);
+
+                if (unlikely(forwardIp > mflimitPlusOne)) goto _last_literals;
+                assert(ip < mflimitPlusOne);
+
+                if (dictDirective == usingDictCtx) {
+                    if (matchIndex < startIndex) {
+                        /* there was no match, try the dictionary */
+                        assert(tableType == byU32);
+                        matchIndex = LZ4_getIndexOnHash(h, dictCtx->hashTable, byU32);
+                        match = dictBase + matchIndex;
+                        matchIndex += dictDelta;   /* make dictCtx index comparable with current context */
+                        lowLimit = dictionary;
+                    } else {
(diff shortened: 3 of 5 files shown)


Cyan4973:
- On which platforms is this PR presumed to be beneficial ?
- Please document the expected gains in these cases
- Are there other platforms on which this patch is suspected to be detrimental ?

Cyan4973 (changes requested):
The asan / valgrind errors are very consequential.
They must be fixed before moving the patch forward.

BiplabRaut:
> The asan / valgrind errors are very consequential. They must be fixed before moving the patch forward.

Hi Yann,
The issues have been fixed and all the tests are now passing. Thanks

BiplabRaut:
> * On which platforms is this PR presumed to be beneficial ?
> * Please document the expected gains in these cases
> * Are there other platforms on which this patch is suspected to be detrimental ?

The code change adds two types of optimizations to speed up compression (also mentioned in the commit log message) :
- Prefetching of input and match bytes by accessing them before they get used for comparison
- Delay the condition checking involving LZ4_DISTANCE_MAX for noDict case for helping branch prediction.

These optimizations help compression speed on x86 CPUs in general and may also extend the performance benefits to Arm CPUs (although not tested).
As these changes are mostly generic memory and code optimizations without any assembly/architecture specific changes, so we don't expect any platforms to suffer in any way.

In our tests on Zen4 and Zen5 based AMD CPUs, we observed the below performance benefits (with GCC 14.2) :
**8.8%** compression speed up for Silesia
**13%** compression speed up for freeBSD

Thanks.

BiplabRaut:
HI @Cyan4973 , can you please take up this PR for review

build/make/lz4defs.make
```diff
@@ -88,6 +88,11 @@ else
 POSIX_ENV = No
 endif
 
+# `AOCL_PRELOAD_BRANCH_OPT` Enable data preload for byte matching within the inner loop in 
+# LZ4_compress_generic_validated and better branch predictor behavior for noDict directive.
+
+
+USEROPTIONALFLAGS = -DAOCL_PRELOAD_BRANCH_OPT=1
```
Cyan4973: nit: I have a preference for :

```
USEROPTIONALFLAGS ?= -DAOCL_PRELOAD_BRANCH_OPT=1
```

so that `USEROPTIONALFLAGS` can also be set via environment variable (on top of command line argument, which remains higher priority).

Also: if the patch is beneficial to some targets and detrimental to others, we might have some conditional platform logic to implement here.

Cyan4973:
> HI @Cyan4973 , can you please take up this PR for review

Hi @BiplabRaut ,

sorry, we had some consequential activity to deal with these past few weeks.

Back to some reasonable availability.

I guess the questions asked at the beginning of this post remain valid:

- On which platforms is this PR presumed to be beneficial ?
   I'm presuming AMD, but maybe only "recent" cpus, or maybe x64 in general.
- Please document the expected gains for these known cases
- Are there other platforms (like `arm`) to check to ensure this patch (which is currently applied unconditionally) is not detrimental ?

On the plus side, the patch is (mostly) concentrated in one place, which makes gating it easy.
What I don't know at this point is if it's expected to produce generally positive speed outcomes, and at worse neutral, thus being a replacement candidate for the existing code, 
or if its impact varies (sometimes positive, sometimes negative) across platforms, thus requiring to keep the alternative around and cleverly select the correct variant for the local environment.

BiplabRaut:
> > HI @Cyan4973 , can you please take up this PR for review
> 
> Hi @BiplabRaut ,
> 
> sorry, we had some consequential activity to deal with these past few weeks.
> 
> Back to some reasonable availability.
> 
> I guess the questions asked at the beginning of this post remain valid:
> 
> * On which platforms is this PR presumed to be beneficial ?
>   I'm presuming AMD, but maybe only "recent" cpus, or maybe x64 in general.
> * Please document the expected gains for these known cases
> * Are there other platforms (like `arm`) to check to ensure this patch (which is currently applied unconditionally) is not detrimental ?
> 
> On the plus side, the patch is (mostly) concentrated in one place, which makes gating it easy. What I don't know at this point is if it's expected to produce generally positive speed outcomes, and at worse neutral, thus being a replacement candidate for the existing code, or if its impact varies (sometimes positive, sometimes negative) across platforms, thus requiring to keep the alternative around and cleverly select the correct variant for the local environment.

Thank you Yann for your further comments.

We will get back with detailed test numbers/graphs and share here.

On your point regarding whether this code can be a replacement/separate candidate to the existing code, we think it would give more flexibility for its use across platforms.
Once we share our performance graphs, may be you can review and give your views on this.

build/make/lz4defs.make
BiplabRaut: Hi @Cyan4973, we have modified it based on your comment. Thanks

BiplabRaut:
Hi @Cyan4973,
Here is the summary of performance gains that we see with this PR:
(As you suggested, we have also benchmarked it on Arm)

**Performance Highlights** :- (No compression ratio change)
Benchmark result (single-threaded execution) graphs are shared below for the datasets: Silesia, Calgary, Canterbury and FreeBSD.
Other datasets like geekBench and Enwik showed similar trends.

image

_**Performance gains on AMD Zen4 based CPU Setup**_: AMD Genoa (Zen4) CPU, SMT OFF, GCC 14.2.1



image

_**Performance gains on AWS Arm based Graviton CPU Setup**_: AWS Graviton4 (Neoverse-V2) CPU, SMT OFF, GCC 13.3

BiplabRaut:
Hi @Cyan4973 , can you take a look at the perf numbers shared for your reference.
Let me know if there are any other review comments from you?

BiplabRaut:
Hi @Cyan4973 , Can you please prioritize this PR.
We have a few more PRs that build on top of this.

PR #1711, opened 2026-01-31
Fix compilation warnings

brainy:
In a custom amalgamated build of LZ4 with `XXH_PRIVATE_API` defined, `-std=c90` and stricter compilation warnings, three classes of warnings were emitted. This is a trivial fix with no functional change.

```
lz4.c:1590:5: warning: no previous prototype for function 'LZ4_loadDict_internal' [-Wmissing-prototypes]
 1590 | int LZ4_loadDict_internal(LZ4_stream_t* LZ4_dict,
```

```
lz4frame.c:180:31: warning: redefinition of typedef 'BYTE' is a C11 feature [-Wtypedef-redefinition]
  182 |   typedef unsigned char       BYTE;
```

```
xxhash.c:864:54: warning: cast from function call of type 'int' to non-matching type 'XXH_endianness' [-Wbad-function-cast]
  864 |     XXH_endianness endian_detected = (XXH_endianness)XXH_CPU_LITTLE_ENDIAN;
```

### DIFF
--- a/lib/xxhash.c
+++ b/lib/xxhash.c
@@ -217,10 +217,10 @@ typedef enum { XXH_bigEndian=0, XXH_littleEndian=1 } XXH_endianness;
 
 /* XXH_CPU_LITTLE_ENDIAN can be defined externally, for example on the compiler command line */
 #ifndef XXH_CPU_LITTLE_ENDIAN
-static int XXH_isLittleEndian(void)
+static XXH_endianness XXH_isLittleEndian(void)
 {
-    const union { U32 u; BYTE c[4]; } one = { 1 };   /* don't use static : performance detrimental  */
-    return one.c[0];
+    const union { U32 u; BYTE c[4]; } one = { XXH_littleEndian };   /* don't use static : performance detrimental  */
+    return (XXH_endianness)one.c[0];
 }
 #   define XXH_CPU_LITTLE_ENDIAN   XXH_isLittleEndian()
 #endif
@@ -398,7 +398,7 @@ XXH_PUBLIC_API unsigned int XXH32 (const void* input, size_t len, unsigned int s
     XXH32_update(&state, input, len);
     return XXH32_digest(&state);
 #else
-    XXH_endianness endian_detected = (XXH_endianness)XXH_CPU_LITTLE_ENDIAN;
+    XXH_endianness endian_detected = XXH_CPU_LITTLE_ENDIAN;
 
     if (XXH_FORCE_ALIGN_CHECK) {
         if ((((size_t)input) & 3) == 0) {   /* Input is 4-bytes aligned, leverage the speed benefit */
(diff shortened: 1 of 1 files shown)


lib/lz4.c
```diff
@@ -1587,7 +1587,7 @@ int LZ4_freeStream (LZ4_stream_t* LZ4_stream)
 
 typedef enum { _ld_fast, _ld_slow } LoadDict_mode_e;
 #define HASH_UNIT sizeof(reg_t)
-int LZ4_loadDict_internal(LZ4_stream_t* LZ4_dict,
+static int LZ4_loadDict_internal(LZ4_stream_t* LZ4_dict,
```
Cyan4973, line 1590: Sidenote: I'm really surprised this was not caught before. We may be missing a compilation flag to catch that.

Cyan4973:
Do you have a reproduction scenario, that would produce the warnings without this fix?

lib/lz4frame.c
```diff
@@ -169,17 +169,21 @@ static int g_debuglog_enable = 1;
 **************************************/
 #if !defined (__VMS) && (defined (__cplusplus) || (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 199901L) /* C99 */) )
 # include <stdint.h>
+# ifndef XXH_PRIVATE_API
```
Cyan4973, line 172: This one feels weird. Like a misplaced responsibility, or a fragile coupling.
But, it's not immediately obvious which better solution exists out there...

lib/lz4.c
Cyan4973: Clearly, we are missing `-Wmissing-prototypes`. This should be added. I would have caught these issues.

lib/lz4frame.c
Cyan4973: I think some proper solutions are:
- update `xxhash` to a newer version that doesn't use `U32` / `U16` (it moved to `xxh_u32` / `xxh_u16` since)
- if using the xxhash version bundled in `lz4`, `#define MEM_MODULE`, this will disable the `typedef U32` part within `xxhash`.
- rely on `C11` to make `typedef` redefinition valid
- silence the specific warning

Cyan4973:
`-Wmissing-prototypes` compilation warnings are now fixed and verified in latest `dev` commits.

lib/lz4frame.c
brainy: Apologies for the late reply. I found these while upgrading the optionally used LZ4 amalgamation that [ships with Subversion](https://svn.apache.org/repos/asf/subversion/branches/better-pristines/subversion/libsvn_subr/lz4/).

This code already implements my proposed patch.

Notes:

 - `lz4.h` was renamed to `lz4internal.h` to avoid header name clashes, since Subversion can be compiled with the included amalgamated LZ4 even when an external library is available.
 - The entry point for compiling LZ4 is `svnlz4.c`.
 - `svnlz4.h` defines some Subversion-specific options, like changing the visibility of the LZ4 API.

As for relying on C11 – Subversion compiles with `-std=c90` and it would make including LZ4 in our source a bit easier if C11 constructs could be avoided.

brainy:
I just resolved the conflicts and there's only one commit left. The changes you already made seem to work now in my environment.

lib/lz4frame.c
brainy: This `#ifdef` isn't needed any more.

Issue #1674, opened 2025-11-12
Possible off-by-one in LZ4F_compressBound_internal?


PQCraft:
**Describe the bug**
The description of `LZ4F_compressBound_internal` states: `When srcSize==0, LZ4F_compressBound() provides an upper bound for LZ4F_flush() and LZ4F_compressEnd() operations.`.
I am using the lz4file functions.
`LZ4F_writeOpen` does `writeFile->dstBufMaxSize = LZ4F_compressBound(0, prefsPtr);` which calls `LZ4F_compressBound_internal` and passes on that `srcSize` of `0`.
`LZ4F_compressBound_internal` returns a value of `65543` that is then used later with `LZ4F_compressUpdate`, which returns error `-11` (`ERROR_dstMaxSize_tooSmall`).
Editing `LZ4F_writeOpen` to pass `1` or `2` instead of `0` causes it to return `65544` which does not cause an error.
I did some poking around, and removing the ` - 1` from the `size_t const maxBuffered = blockSize - 1;` line in `LZ4F_compressBound_internal` makes it return `65544` when given `0`, which gets rid of the error, and does not affect the return value if given `1` or `2`.

**Expected behavior**
`LZ4F_compressBound_internal` should return `65544` for `0` like it does for other values (`1` and `2` were tested).

**To Reproduce**
1. Try to use writing part of the lz4file API
2. Encounter error `-11` (`ERROR_dstMaxSize_tooSmall`)
3. Inspect the code and find that `LZ4F_compressBound_internal` seems to return `1` less than it should when supplied with `0`
4. Remove the ` - 1` from the `size_t const maxBuffered = blockSize - 1;` line in `LZ4F_compressBound_internal`
5. Error goes away

**System (please complete the following information):**
 - OS: Arch Linux
 - Version 1.10.0, commit `c99cfe4792830ed66b3a44310abfcd122d605ab2`
 - Compiler: GCC 15.2.1
 - Build System: Makefile (files dumped into my project and compiled with its Makefile)
 - Other hardware specs: AMD Ryzen 7 5825U

**Additional context**
I am not sure if the minus 1 is intentional but it seems to be causing problems. I don't know enough about LZ4 to determine if this is actually incorrect.

PQCraft:
Fixed in #1714 👍

PR #1714, opened 2026-02-08
Fix compress bound off by one

jlee303:
Fix bug from #1674, thanks @PQCraft for finding bug!

LZ4F_writeOpen() allocates its output buffer using LZ4F_compressBound(0, prefs). However, LZ4F_compressBound(0) returns the bound for LZ4F_flush() and LZ4F_compressEnd() operations. Fix by using LZ4F_compressBound(blockSize, prefs)

Added test case to detect issue:
Here is the error message when using LZ4F_compressBound(0, prefs)
```
LZ4F_compressBound(0) =  65551 
LZ4F_compressBound(1 KB, autoFlush=1) =  1032 
LZ4F_compressBound(1 KB, autoFlush=0) =  65544 
LZ4F_write checksums disabled edge case : Error detected ! 
```

I did not see a test file for lz4file so I added the test case to frametest.c, let me know if I should move the test to a better location!


tests/frametest.c
```diff
@@ -442,25 +442,28 @@ static int unitTests(U32 seed, double compressibility)
         LZ4F_preferences_t writePrefs;
         size_t             blockSize;
         char*              data;
+        int                failed;
```
Cyan4973: minor: 
I would prefer initialization to happen at the same place as declaration.
This would give : `int failed = 0;` rather than 2 separated lines with an uninitialized state in between.

And yes, this comment is valid for all other variables declared in this block.

tests/frametest.c
```diff
@@ -442,25 +442,28 @@ static int unitTests(U32 seed, double compressibility)
...
 
         tmpFile   = tmpfile();
         writeCtx  = NULL;
         blockSize = LZ4F_getBlockSize(LZ4F_default);
         data      = (char*)malloc(blockSize);
+        failed    = 0;
 
         memset(data, 0, blockSize);
```
Cyan4973: The main issue, on reaching this code, is that `data` is not tested. So allocation could have failed. In which case, this `memset()` is writing into `NULL`.

2 solutions:
- allocate with `calloc()`, which automatically fills with zero on success
- test and branch out on allocation failure before `memset()`.

Note that `calloc()` spares you the `memset()`, but you'll still need to test and branch out on allocation failure.

tests/frametest.c
```diff
@@ -442,25 +442,28 @@ static int unitTests(U32 seed, double compressibility)
...
         memset(&writePrefs, 0, sizeof(writePrefs));
         writePrefs.frameInfo.contentChecksumFlag = LZ4F_noContentChecksum;
         writePrefs.frameInfo.blockChecksumFlag   = LZ4F_noBlockChecksum;
 
-        if (!tmpFile || !data) goto _output_error;
-        if (LZ4F_isError(LZ4F_writeOpen(&writeCtx, tmpFile, &writePrefs))) goto _output_error;
-        if (LZ4F_isError(LZ4F_write(writeCtx, data, blockSize - 1))) goto _output_error;
-        if (LZ4F_isError(LZ4F_write(writeCtx, data, 1))) goto _output_error;
+        if (!tmpFile || !data) failed = 1;
```
Cyan4973: checking `!data` here is already too late.

Also you may not need an additional variable is all you need is to scope out the next lines into a `{ ... }` block.

tests/frametest.c
```diff
@@ -442,25 +442,28 @@ static int unitTests(U32 seed, double compressibility)
...
+        if (!failed && LZ4F_isError(LZ4F_writeOpen(&writeCtx, tmpFile, &writePrefs))) failed = 1;
```
Cyan4973: I guess it works, but it feels unusual.

I would prefer a pattern like : 

```
do {
    if (status(operation1(arg1)) == fail) break; /* exit the while loop */

    if (status(operation2(arg2)) == fail) break; 

    // etc...
} while(0);
```

tests/frametest.c
```diff
@@ -51,24 +51,22 @@
...
 /* unoptimized version; solves endianness & alignment issues */
-static void FUZ_writeLE32 (void* dstVoidPtr, U32 value32)
+static void FUZ_writeLE32(void* dstVoidPtr, U32 value32)
 {
     BYTE* const dstPtr = (BYTE*)dstVoidPtr;
-    dstPtr[0] = (BYTE) value32;
-    dstPtr[1] = (BYTE)(value32 >> 8);
-    dstPtr[2] = (BYTE)(value32 >> 16);
-    dstPtr[3] = (BYTE)(value32 >> 24);
+    dstPtr[0]          = (BYTE)value32;
```
Cyan4973: minor: it seems this PR has been mixed with a formatting exercise.
It's not in itself bad, but it makes it harder to find the changes that matter, hence more difficult to review.
That's why formatting is typically left as a separate exercise.

lib/lz4file.c
```diff
@@ -275,7 +275,7 @@ LZ4F_errorCode_t LZ4F_writeOpen(LZ4_writeFile_t** lz4fWrite, FILE* fp, const LZ4
   writeFile->maxWriteSize = blockSize;
 
   /* Calculate and allocate destination buffer */
-  writeFile->dstBufMaxSize = LZ4F_compressBound(0, prefsPtr);
+  writeFile->dstBufMaxSize = LZ4F_compressBound(blockSize, prefsPtr);
```
Cyan4973, line 278: 👍 the change that matters

Cyan4973 (changes requested):
The library change itself is correct.

The focus in more on the test side.
There are a few patterns that deserve some attention.

tests/frametest.c
jlee303: Reverted formatting changes! Sorry auto format was on 😸

tests/frametest.c
jlee303: Updated! Lemme know if there's anything else :)

Cyan4973 (approved):
Great work @jlee303 !

PR #1732, opened 2026-04-04
fix heap buffer overflow in multi-file filename construction

metsw24-max:
The multi-file compression logic allocated insufficient space for
destination filenames `input_length + 20` and then appended an
unbounded suffix using strcat, allowing heap buffer overflow when
suffix length exceeded the assumed margin.

This patch:
- computes exact required buffer size (src + suffix + null)
- adds integer overflow protection
- replaces unsafe string operations with memcpy
- consolidates logic into LZ4IO_prepareDstFileName()

Fixes both legacy and non-legacy multi-file compression paths.

Additionally replaces sprintf with snprintf in LZ4IO_toHuman().


metsw24-max:
any updates on this pr?

PR #1735, opened 2026-04-13
heap buffer overflow in suffix-based filename construction

jmestwa-coder:
## Summary

Fix a heap buffer overflow in `lz4io` multi-file compression when constructing destination filenames with a suffix.

---

## Problem

The destination buffer is resized using a fixed heuristic:

ofnSize = ifnSize + 20;

But the actual required size is:

required = ifnSize + suffixSize + 1;

Since suffix is provided by the caller and not length-restricted, large values can exceed the allocated buffer, leading to overflow during:

strcpy(dstFileName, input);
strcat(dstFileName, suffix);

---

## Reproduction

Calling LZ4IO_compressMultipleFilenames() with a long suffix (e.g. 5000 bytes) causes a crash before this change.

---

## Fix

Compute exact required size:

neededSize = ifnSize + suffixSize + 1;

Add overflow guard for size calculation
Allocate buffer based on neededSize

---

## Behaviour

No change for normal inputs.
Previously crashing inputs are now handled safely using existing missed_files logic.

### DIFF
--- a/programs/lz4io.c
+++ b/programs/lz4io.c
@@ -951,6 +951,7 @@ int LZ4IO_compressMultipleFilenames_Legacy(
     for (i=0; i<ifntSize; i++) {
         unsigned long long processed = 0;
         size_t const ifnSize = strlen(inFileNamesTable[i]);
+        size_t neededSize;
         if (LZ4IO_isStdout(suffix)) {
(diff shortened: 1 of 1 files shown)


jmestwa-coder:
gentle ping

PR #1736, opened 2026-04-14
reject truncated frames in --list by validating full reads

jmestwa-coder:
### Summary

Fixes an inconsistency where --list could incorrectly accept truncated or corrupted LZ4 frames as valid.

### Problem

\--list does not strictly validate input length:

*   Block headers may be accepted on partial reads
    
*   Frame content checksum is skipped using fseek(), which does not verify data presence
    

This allows truncated inputs to be reported as valid, while -t correctly rejects them.

### Fix

*   Require exact-length reads for:
    
    *   Block headers
        
    *   Frame content checksum
        
*   Replace fseek() with fread() to ensure bytes are actually present
    

### Result

*   \--list now rejects truncated/corrupted frames
    
*   Behavior is consistent with -t
    
*   No change for valid inputs
    

### Scope

*   Localized to LZ4IO\_skipBlocksData()
    
*   No API changes
    
*   No impact on valid workflows
    

### Validation

*   Verified with truncated inputs (default and --no-frame-crc)
    
*   Verified valid inputs remain unaffected

### DIFF
--- a/programs/lz4io.c
+++ b/programs/lz4io.c
@@ -2608,8 +2608,7 @@ LZ4IO_skipBlocksData(FILE* finput,
     unsigned char blockInfo[LZ4F_BLOCK_HEADER_SIZE];
     unsigned long long totalBlocksSize = 0;
     for (;;) {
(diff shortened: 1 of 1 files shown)


jmestwa-coder:
any update?

PR #1741, opened 2026-04-30
prevent integer overflow in block count and allocation sizes

jmestwa-coder:
### Summary

Prevent integer overflow in `programs/bench.c` when computing block counts and allocation sizes.

---

### Problem

- `U32 maxNbBlocks` can overflow for large inputs
- `srcSize` is truncated when passed to `LZ4_compressBound`

This can lead to incorrect allocation sizes.

---

### Fix

- Use `size_t` for block count
- Guard allocation multiplication
- Clamp input to `LZ4_compressBound` to `INT_MAX`

---

### Impact

- Avoids overflow in allocation paths
- No change for typical inputs
- Limited to benchmark tool

### DIFF
--- a/programs/bench.c
+++ b/programs/bench.c
@@ -42,7 +42,7 @@
 #include <stdio.h>       /* fprintf, fopen, ftello */
 #include <time.h>        /* clock_t, clock, CLOCKS_PER_SEC */
 #include <assert.h>      /* assert */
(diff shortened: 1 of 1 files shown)


jmestwa-coder:
pushed a fix for the compiler test failures. SIZE_MAX isn't available in c90 so switched to (size_t)-1, matching the existing usage in lz4frame.c.

PR #1742, opened 2026-04-30
prevent overflow when growing buffer registry

jmestwa-coder:
### Summary

Prevent overflow when growing the buffer registry in `programs/lz4io.c`.

---

### Problem

Capacity growth uses unchecked arithmetic:

- `oldCapacity + addedCapacity` may overflow
- `newCapacity * sizeof(BufferDesc)` may overflow

This can result in incorrect allocation sizes.

---

### Fix

- Guard addition before computing new capacity
- Guard multiplication before allocation

---

### Impact
- Prevents overflow in buffer growth logic
- No behavior change for normal usage


jmestwa-coder:
pushed a fix for the compiler test failures: the overflow guard tripped -Wdeclaration-after-statement on the C90 builds, restructured so declarations come first.

PR #1757, opened 2026-06-03
ci: pin GitHub Actions to full commit SHAs


XananasX7:
Pin `actions/checkout`, `actions/upload-artifact`, `actions/download-artifact` from mutable version tags (`@v4`) to full commit SHAs. Mutable refs allow silent supply chain attacks. Recommended by GitHub's security hardening guide and OpenSSF Scorecard.


XananasX7:
Friendly ping — happy to address any review comments. Thank you!

PR #1760, opened 2026-06-11
ossfuzz: add lz4_frame_crc_fuzzer for content-checksum verification

XananasX7:
## Summary

The existing `decompress_frame_fuzzer` calls `LZ4F_decompress()` without enabling content-checksum verification. The **xxHash content-checksum path** — accumulated across multi-call incremental decode in `lz4frame.c` — is not fuzz-covered.

## New harness: `ossfuzz/lz4_frame_crc_fuzzer.c`

- Sets `dopt.stableDst = 1` to enable xxHash content-checksum accumulation
- Feeds input in small fuzz-derived chunks (1–32 bytes) across multiple `LZ4F_decompress()` calls
- Exercises the incremental xxHash state in `lz4frame.c` and the final checksum verification at frame end

Specifically targets the `XXH32_update` accumulation across chunk boundaries and the `LZ4F_decompress` frame-completion branch that validates the stored vs computed checksum.

### DIFF
--- a/ossfuzz/lz4_frame_crc_fuzzer.c
+++ b/ossfuzz/lz4_frame_crc_fuzzer.c
@@ -1,0 +1,50 @@
+/* lz4_frame_crc_fuzzer.c
+ *
+ * Fuzz target for LZ4F frame decompression with content-checksum verification.
+ * The existing decompress_frame_fuzzer calls LZ4F_decompress() without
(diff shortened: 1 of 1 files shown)


XananasX7:
Friendly ping — happy to address any review comments. Thank you!

PR #1769, opened 2026-07-03
fix short header read for dictID frames in lz4 --list

arshsmith1:
LZ4IO_getCompressedFileInfo already reads LZ4F_HEADER_SIZE_MIN bytes of the frame header, then only fetches the remaining bytes when hSize is greater than LZ4F_HEADER_SIZE_MIN + MAGICNUMBER_SIZE. A frame that sets the dictID flag but no content size has an 11-byte header, so that branch is skipped and the four dictID plus header-checksum bytes are left uninitialized on the stack, which makes `lz4 --list` reject a perfectly valid file as unrecognized. Comparing hSize against LZ4F_HEADER_SIZE_MIN reads the rest of the header for every layout; I added a golden sample with a dictID and a --list check that fails without this change.


arshsmith1:
any update?

PR #1773, opened 2026-07-16
Fix buffer under-allocation in multi-file compression and add tests

VinogradovDmitri:
- Fix heap overflow in LZ4IO_compressMultipleFilenames_Legacy() where output filename buffer was sized +20 regardless of suffix length – now uses + suffixSize + 1;
- Fix double-pointer bug in LZ4F_writeOpen() error path;
- Fix integer division in examples (percentage calculations);
- Add regression test (legacy_suffix_repro) that triggers the overflow (passes with ASan after fix);
- Enhance decompress-partial-usingDict with allocation tracking and OOM leak tests;
- Update build files to include new test dependencies.

**Note:** Perhaps I shouldn't have touched the legacy code, and perhaps it would have been better to leave it as it was if it was no longer in use. However, since it is still present and may cause issues, I have decided to address it. If this intervention is not desired, please let me know, and I will either redo or roll back this portion. Thank you for understanding!


VinogradovDmitri:
Any ideas on how to fix CI failing?

PR #1776, opened 2026-07-28
Reject negative decompress sizes and frame contentSize underflow

M0nd0R:
## Summary
- Reject negative `srcSize` in `LZ4_decompress_generic` (already rejected negative `outputSize`) so a negative `compressedSize` cannot wrap the input end pointer.
- Fail closed in `LZ4F_decompress` when a block would subtract past declared `contentSize` / `frameRemainingSize`.

## Test plan
- [x] `make -C lib` succeeds on macOS
- [ ] Run existing decompress / frame fuzzer or test suite
- [ ] Confirm `LZ4_decompress_safe(src, dst, -1, dstCapacity)` returns an error


M0nd0R:
Closing.

PR #1777, opened 2026-07-28
Fix heap overflow in multi-file output name buffer sizing

M0nd0R:
## Summary
- `LZ4IO_compressMultipleFilenames` / `_Legacy` grew the destination name buffer to `ifnSize + 20` and then `strcat`'d the caller-provided `suffix`. Any `suffix` longer than 20 bytes overflowed the heap allocation.
- Size the buffer to `ifnSize + suffixSize + 1` instead.
- In `LZ4IO_decompressMultipleFilenames`, validate the suffix length/match before computing the stripped output name, and grow to the exact stripped length.

## Test plan
- [x] ASAN PoC with short input path + 40-byte suffix: overflow before fix, clean after
- [ ] Existing multi-file compress/decompress CLI tests (`lz4 -m`, decompress with `.lz4` suffix)

### DIFF
--- a/programs/lz4io.c
+++ b/programs/lz4io.c
@@ -961,7 +961,7 @@ int LZ4IO_compressMultipleFilenames_Legacy(
 
         if (ofnSize <= ifnSize+suffixSize+1) {
             free(dstFileName);
-            ofnSize = ifnSize + 20;
+            ofnSize = ifnSize + suffixSize + 1;
             dstFileName = (char*)malloc(ofnSize);
             if (dstFileName==NULL) {
                 return ifntSize;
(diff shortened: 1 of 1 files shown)


fallenmi (approved):
Approved at exact head `f18a372`.

I reproduced the long-suffix bug independently against exact base `0774d05`: both the frame and legacy multi-file paths produce an ASan heap-buffer-overflow in `strcat` with a 70-byte suffix. On this head, both paths exit cleanly under ASan/UBSan, and each long-suffix output decompresses back through the matching custom suffix without a sanitizer finding. The focused upstream `test-lz4-multiple` and `test-lz4-multiple-legacy` targets also pass, and the diff is clean.

The two failing live checks are unrelated to this `programs/lz4io.c` patch (the known VS17-on-VS18 runner mismatch and the unchanged macOS example conversion warning); the other 53 checks succeed.

Disclosure: I used Codex to help analyze the diff and run the verification above.

M0nd0R:
Closing.

PR #1784, opened 2026-08-10
lz4io: parallelize chunk reads so MT file compression scales past 8 threads

ksenxx:
## Problem

Multithreaded file-mode compression (`lz4 -T#`) stops scaling around 8 threads and plateaus near 1.5–2 GB/s, regardless of core count. Profiling (`perf` on `lz4 -1 -T16`, 1.7 GB input in tmpfs) showed only ~5 of 16 workers busy: `LZ4IO_readAndProcess()` feeds every worker through a serial chain — a self-resubmitting job in the same thread pool that does, per 4 MB chunk: buffer allocation → `fread` → sequential XXH32 content-checksum update → dispatch → re-queue behind busy compression jobs.

## Change

For seekable regular files with independent blocks, dispatch lightweight jobs which each `pread()` their own 4 MB chunk at a known absolute offset (anchored on the stream's initial `ftello` position, EINTR/short-read safe, into pooled reused buffers), hash it in strict block order through a small ordered turnstile ("HashGate") so the frame's sequential XXH32 stays correct, then compress it. The serial read chain is gone from the critical path.

Because the frame content checksum is inherently sequential, it becomes the pipeline bottleneck at fast levels; worker count is clamped there (12) to limit memory-bandwidth contention against the hashing stage. The clamp does not apply to HC levels (`compressionLevel >= LZ4HC_CLEVEL_MIN`), which are compression-bound and need every worker.

The sequential path is preserved unchanged for pipes/stdin, dependent blocks (`-BD`), the legacy format, Windows, and non-POSIX platforms. Only `programs/lz4io.c` is touched; **compressed output is byte-identical in all modes**.

## Results

silesia ×8 (1.7 GB, tmpfs, output to /dev/null), 2× Xeon 16C/32T, best-of-3 wall clock, `lz4 -1 -T#`:

| Threads | before | after |
|---|---|---|
| 1 | 410 MB/s | 404 MB/s (parity) |
| 8 | ~1.5–2.0 GB/s | 2.9 GB/s |
| 16 | ~1.5–2.0 GB/s | 3.6 GB/s |
| 32 | ~1.5–1.9 GB/s | 3.6–3.7 GB/s |

(The "before" range spans a plain-`make` build ≈1.5 GB/s and an -O3/LTO/native build ≈2.0 GB/s; both plateau from 8 threads on.) With `--no-frame-crc` the same binary reaches 6.4–8.1 GB/s at T32, confirming the remaining ceiling is the frame format's sequential XXH32, not the I/O design. HC levels also benefit: `-9 -T32` ≈ +9%, `-3 -T32` ≈ +27% on the same corpus.

At 16–32 threads this makes the lz4 CLI faster than `pzstd -1` and `zstd -1 -T#` on the same input (~3.6 vs ~2.7–3.3 GB/s), where it previously lost.

## Correctness / verification

- Byte-identity oracle vs the pre-change binary: 105-case matrix (5 file shapes × `-T1/16/32` × 7 flag sets incl. `-B4..7`, `--no-frame-crc`, `--content-size`, HC levels) plus large-corpus md5s — all identical.
- Regular-file stdin redirected at a nonzero offset compresses the correct byte range (offsets anchored via `ftello`; falls back to the serial path if the position is unknown).
- Truncating the source mid-compression is reported as a clean read error (exit 32) — chunks are `pread` into pooled buffers, no mmap, so no SIGBUS window.
- `EINTR`/short reads are retried at the right offset.
- ASan/UBSan/TSan clean; `make -C tests test-lz4-basic`, `frametest`, and fuzzer runs pass; roundtrips verified with `lz4 -t` and md5.

The second commit contains fixes for 5 defects found by adversarial review of the first (wrong-offset stdin, mmap SIGBUS on concurrent truncation, HC over-clamping, fatal EINTR, `MAP_ANONYMOUS` portability); details in its commit message.

### DIFF
--- a/programs/lz4io.c
+++ b/programs/lz4io.c
@@ -730,7 +730,13 @@ typedef struct {
     pthread_mutex_t mutex;
     pthread_cond_t cond;
     XXH32_state_t* xxh32;
-    unsigned long long expectedRank;
+    /* The rank is deliberately 32-bit :
+     * several 32-bit targets have no native 64-bit atomics
+     * (the spin fast path below would not link, or would be misaligned),
+     * while a naturally-aligned unsigned is atomic-friendly everywhere.
+     * Equality comparison stays unambiguous under wraparound,
+     * since at most nbWorkers (<< 2^32) ranks are in flight at any time. */
+    unsigned expectedRank;
 } HashGate;
 
 static void HG_init(HashGate* hg, XXH32_state_t* xxh32, unsigned long long firstRank)
@@ -738,7 +744,7 @@ static void HG_init(HashGate* hg, XXH32_state_t* xxh32, unsigned long long first
     if (pthread_mutex_init(&hg->mutex, NULL) || pthread_cond_init(&hg->cond, NULL))
         END_PROCESS(46, "hash gate creation error");
     hg->xxh32 = xxh32;
-    hg->expectedRank = firstRank;
+    hg->expectedRank = (unsigned)firstRank;
 }
 
 static void HG_destroy(HashGate* hg)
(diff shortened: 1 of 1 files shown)


ksenxx:
Pushed 874fcc9 fixing the CI failures on 32-bit targets (ARM/MIPS/M68K/PPC QEMU, UBSan x86): the HashGate spin path used 64-bit atomics, which don't link on targets without native 8-byte atomic ops and trip `-Watomic-alignment` on x86 `-m32`. The gate's rank is now a naturally-aligned 32-bit `unsigned` (wraparound-safe: at most nbWorkers ranks in flight), with the spin gated on `__GCC_ATOMIC_INT_LOCK_FREE == 2`.

The two remaining failures appear unrelated to this PR — the same jobs fail on other recent PRs (e.g. the freebsd-ci PR from Aug 6):
- `macOS`: `-Werror -Wimplicit-int-float-conversion` in `examples/frameCompress.c:446` under the Xcode 26.6 toolchain (file untouched by this PR);
- `Core Tests (windows-latest, Release)`: CMake generator "Visual Studio 17 2022" not found on the current `windows-latest` runner image.

PR #1785, opened 2026-08-11
ci: pin third-party action references to full commit SHAs

Alb3e3:
## Summary

Two workflow files use mutable action references that could allow supply-chain attacks if the referenced repositories are compromised:

**`oss-fuzz.yml`** — both CI Fuzz steps reference `@master` (the highest-risk form: resolves to HEAD of `google/oss-fuzz` at runtime). Any commit pushed to that branch would immediately execute inside lz4's CI environment.

**`cmake-test.yml`** — three actions use version tags (`@v6.0.2`, `@v3`, `@v2`) instead of commit SHAs. This is inconsistent with the other eight workflow files in this repository, which all use the pinned SHA+comment form (e.g. `actions/checkout@de0fac2e... # v6.0.2`).

## Changes

| File | Action | Before | After |
|---|---|---|---|
| `oss-fuzz.yml` | `google/oss-fuzz/.../build_fuzzers` | `@master` | `@f9bb0e5...` |
| `oss-fuzz.yml` | `google/oss-fuzz/.../run_fuzzers` | `@master` | `@f9bb0e5...` |
| `cmake-test.yml` | `actions/checkout` | `@v6.0.2` | `@de0fac2...` (matches other workflows) |
| `cmake-test.yml` | `microsoft/setup-msbuild` | `@v3` | `@30375c6...` |
| `cmake-test.yml` | `jwlawson/actions-setup-cmake` | `@v2` | `@0d6a7d6...` |

No functional change. SHAs recorded are current as of 2026-08-11.

### DIFF
--- a/.github/workflows/cmake-test.yml
+++ b/.github/workflows/cmake-test.yml
@@ -38,14 +38,14 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v6.0.2
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
     - name: Set up MSVC (Windows)
       if: runner.os == 'Windows'
-      uses: microsoft/setup-msbuild@v3
+      uses: microsoft/setup-msbuild@30375c66a4eea26614e0d39710365f22f8b0af57 # v3
 
     - name: Set up CMake
-      uses: jwlawson/actions-setup-cmake@v2
+      uses: jwlawson/actions-setup-cmake@0d6a7d60b009d01c9e7523be22153ff8f19460d3 # v2
       with:
         cmake-version: latest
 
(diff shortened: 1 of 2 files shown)


Alb3e3:
Same as lz4/lz4#1786 — the `macOS` and `Core Tests (windows-latest, Release)` failures here are not from this PR, which only edits `.github/workflows/cmake-test.yml` and `oss-fuzz.yml`.

Both jobs die building `examples/frameCompress.c`:

```
frameCompress.c:446:40: error: implicit conversion from 'const unsigned long long' to 'double'
  may lose precision [-Werror,-Wimplicit-int-float-conversion]
```

Details and the one-line fix are in #1786.

Alb3e3:
Submitted the `examples/` fix as #1795.

One correction to what I posted here on 2026-08-24: **the one-line change I offered was not sufficient.** `make` stops at the first error, so the CI log only ever shows `frameCompress.c:446` — but `examples/Makefile` has `test : all`, and `bench_functions.c` has **ten more** instances of the same implicit `uint64_t`→`double` denominator conversion queued behind it. Fixing only line 446 moves the error rather than clearing it. #1795 fixes all eleven.

Verified by reproducing with `clang-18`, which emits the identical diagnostic to the Xcode 26.6 runner, then running both macOS CI commands verbatim — `CFLAGS="-Werror -O0" make default` and `CFLAGS="-O3 -Werror -Wconversion -Wno-sign-conversion" make all` — both exit 0 with zero warnings. `frameCompress` output is byte-identical before and after.

Also worth separating out: the `Core Tests (windows-latest, Release)` failure here is **not** the same problem and #1795 will not fix it. That one is `Generator "Visual Studio 17 2022" could not find any instance of Visual Studio` — a runner-image issue needing a workflow change, not a source change.

So this PR should go green on `macOS` once #1795 lands and this branch is rebased; Windows will need its own fix.

PR #1786, opened 2026-08-11
lib: check frame compression bound arithmetic

Alb3e3:
Compression bounds can wrap at `size_t` limits, and the `unsigned` block count can
truncate on 64-bit. For example, on 64-bit,
`LZ4F_compressBound(SIZE_MAX / 2, NULL)` returns **8**, although the correct bound
is **9224497936761618440**. These are allocation-size calculations; reproducing
the incorrect results does not require allocating the corresponding buffers.

Use `size_t` block counts and check buffered-input addition, payload plus block
overhead, and the frame-header addition. Unrepresentable bounds return the existing
`LZ4F_ERROR_srcSize_tooLarge` error. Both compression callers propagate that error,
and streaming compression validates the bound before computing the source-end
pointer. The API comments document checking bound results with `LZ4F_isError()`.
Representable bounds retain their mathematical values.

The context-capacity check now compares its ordered allocation type IDs directly.
This is equivalent to mapping both IDs to byte sizes and lets Clang's analyzer prove
that the one-shot stack context is never passed to `free()`.

This is preventive allocation-size hardening. The tests establish incorrect bounds
and correct rejection; they do not demonstrate an end-to-end exploitable heap
overflow with simultaneously valid source and destination allocations. Streaming
and one-shot compression require their respective bound APIs.

Validation:

- Regression fails against the previous PR head on both 32-bit and 64-bit.
- GCC 64-bit and GCC `-m32`: frame unit tests plus 1,000 randomized tests each.
- Clang 18 ASan/UBSan: frame unit tests plus 1,000 randomized tests.
- Clang 18 `scan-build --status-bugs`: no bugs found.
- C90 library build with warnings treated as errors; `git diff --check`.

The regression covers every block size, flushing/checksum combinations, truncated
block counts, input and overhead overflow, header-only overflow, and propagation
without touching the supplied small buffers.

AI assistance was used to develop and review the patch; the listed tests were run locally.


Alb3e3:
The two red checks here are not caused by this PR.

This PR touches only `lib/lz4frame.c`. The `macOS` and `Core Tests (windows-latest, Release)` jobs fail while building `examples/frameCompress.c`:

```
frameCompress.c:446:40: error: implicit conversion from 'const unsigned long long' to 'double'
  may lose precision [-Werror,-Wimplicit-int-float-conversion]
make[1]: *** [frameCompress] Error 1
```

That is this line:

```c
(double)ret.size_out / ret.size_in * 100);
```

`ret.size_out` is cast, but `ret.size_in` is still `unsigned long long` and gets implicitly converted. Clang in the current Xcode image (26.6) promotes `-Wimplicit-int-float-conversion` to an error under the `-Werror -Wconversion` flags this target builds with. The one-line fix is to cast both operands:

```c
(double)ret.size_out / (double)ret.size_in * 100);
```

Worth noting `dev` only *looks* green — its recent runs are the Scorecard workflow, not the full matrix, so nothing has rebuilt the examples on the newer image until these PRs did. #1785 fails identically for the same reason, and it only touches two workflow YAML files.

Happy to send that one-liner as its own PR if you would like it — it would unblock CI on anything touching this repo, not just these two.

Alb3e3:
Submitted the `examples/` fix as #1795.

One correction to what I posted here on 2026-08-24: **the one-line change I offered was not sufficient.** `make` stops at the first error, so the CI log only ever shows `frameCompress.c:446` — but `examples/Makefile` has `test : all`, and `bench_functions.c` has **ten more** instances of the same implicit `uint64_t`→`double` denominator conversion queued behind it. Fixing only line 446 moves the error rather than clearing it. #1795 fixes all eleven.

Verified by reproducing with `clang-18`, which emits the identical diagnostic to the Xcode 26.6 runner, then running both macOS CI commands verbatim — `CFLAGS="-Werror -O0" make default` and `CFLAGS="-O3 -Werror -Wconversion -Wno-sign-conversion" make all` — both exit 0 with zero warnings. `frameCompress` output is byte-identical before and after.

Also worth separating out: the `Core Tests (windows-latest, Release)` failure here is **not** the same problem and #1795 will not fix it. That one is `Generator "Visual Studio 17 2022" could not find any instance of Visual Studio` — a runner-image issue needing a workflow change, not a source change.

So this PR should go green on `macOS` once #1795 lands and this branch is rebased; Windows will need its own fix.

PR #1788, opened 2026-08-16
fix unvalidated blockSizeID in LZ4F_compressBegin_internal()

jaypatrickhoward:
### Summary

`LZ4F_getBlockSize()` returns an error code when `blockSizeID` is outside the documented `4..7` range, but `LZ4F_compressBegin_internal()` assigns that result to `cctx->maxBlockSize` without checking it. The error code is a huge `size_t`, so the buffer sizing that follows wraps around:

- `requiredBuffSize` becomes `~128 KB` instead of `maxBlockSize + 128 KB`, so `tmpBuff` is far too small;
- `LZ4F_compressBound()` wraps down to a handful of bytes.

A subsequent `LZ4F_compressUpdate()` then copies the input into that undersized `tmpBuff` and overflows it.

The decoder already rejects these values (`lz4frame.c`, `blockSizeID < 4` → `maxBlockSize_invalid`); this restores the same check on the compression side.

### Reproducer

```c
LZ4F_preferences_t prefs;
memset(&prefs, 0, sizeof(prefs));
prefs.frameInfo.blockSizeID = (LZ4F_blockSizeID_t)3;   /* invalid */

size_t cb = LZ4F_compressBound(1 MB, &prefs);          /* returns 4 (!) */
LZ4F_compressBegin(cctx, hdr, sizeof(hdr), &prefs);    /* returns 7, no error */
LZ4F_compressUpdate(cctx, dst, cb, src, 1 MB, NULL);   /* overflows */
```

Under ASan, before this patch:

```
ERROR: AddressSanitizer: heap-buffer-overflow
WRITE of size 1048576 ... 0 bytes after 131070-byte region
    #1 LZ4F_compressUpdateImpl lz4frame.c:1112
    allocated by LZ4F_compressBegin_internal lz4frame.c:766
```

Affected entry points, before / after:

| entry point | before | after |
|---|---|---|
| `LZ4F_compressBegin` + `LZ4F_compressUpdate` | heap-buffer-overflow | `ERROR_maxBlockSize_invalid` |
| `LZ4F_compressFrame` | accepted; emits `BD=0x30`, undecodable | `ERROR_maxBlockSize_invalid` |
| `LZ4F_writeOpen` + `LZ4F_write` (lz4file) | heap-buffer-overflow | `ERROR_maxBlockSize_invalid` |
| valid IDs (`0`, `4`..`7`) | ok | unchanged |

**The `lz4` CLI is not affected** — `LZ4IO_setBlockSizeID()` already clamps to `4..7`. This reaches third-party callers of the `LZ4F_*` / `lz4file` APIs, which is likely why fuzzing hasn't surfaced it: the fuzzers drive the decoder with hostile input but the encoder with valid preferences.

### Note on the shape of the fix

The obvious one-liner is *not* sufficient:

```c
cctx->maxBlockSize = LZ4F_getBlockSize(cctx->prefs.frameInfo.blockSizeID);
FORWARD_IF_ERROR(cctx->maxBlockSize);   /* insufficient */
```

That writes the error code into `cctx->maxBlockSize` before returning. Since `LZ4F_compressBegin_internal()` does not clear `cStage` on the error path, a context that had already been initialized stays at `cStage == 1`, and the next `LZ4F_compressUpdate()` runs with the poisoned value:

```
compressBegin(valid)    -> ok
compressBegin(INVALID)  -> ERROR_maxBlockSize_invalid
compressUpdate(1 MB)    -> heap-buffer-overflow
```

Assigning only after validation avoids this. The added test covers that sequence and fails against the one-line variant.

### Tests

`tests/frametest.c` gains four checks in `unitTests()`:

1. `LZ4F_compressBegin()` rejects `blockSizeID` ∈ `{1,2,3,8,9}`
2. `LZ4F_compressFrame()` never emits an undecodable frame for those IDs
3. valid IDs `{0 (default),4,5,6,7}` still compress
4. a failed `LZ4F_compressBegin()` leaves an already initialized `cctx` intact

The test fails on `dev` (`blockSizeID 1 not rejected !`) and passes with the fix, in a plain non-sanitizer build.

Verified: `frametest -T20s`, `frametest` under ASan, C90 and C99 builds with `-Werror -Wpedantic`, `-Wc++-compat -Wall -Wextra -Werror` on `lz4frame.c`, `unicode_lint.sh`.

### Deliberately out of scope

- `LZ4F_compressBound()` still returns a wrapped-small value for an invalid `blockSizeID`. It is harmless once `LZ4F_compressBegin()` rejects the request (nothing is ever written), and #1786 is already open on that function — keeping this patch disjoint so both can land independently.
- Check 2 above asserts "no undecodable frame" rather than plain rejection because `LZ4F_optimalBSID()` may clamp an ID *above* the range down into range when `srcSize` is small (e.g. `8` → `5` for a 256 KB input), producing a valid frame. That is pre-existing behaviour and memory-safe either way, so it is left untouched.

### DIFF
--- a/lib/lz4frame.c
+++ b/lib/lz4frame.c
@@ -715,6 +715,7 @@ static size_t LZ4F_compressBegin_internal(LZ4F_cctx* cctx,
 
     RETURN_ERROR_IF(dstCapacity < maxFHSize, dstMaxSize_tooSmall);
     if (preferencesPtr == NULL) preferencesPtr = &prefNull;
+    FORWARD_IF_ERROR( LZ4F_getBlockSize(preferencesPtr->frameInfo.blockSizeID) ); /* validate before modifying @cctx */
     cctx->prefs = *preferencesPtr;
     DEBUGLOG(5, "LZ4F_compressBegin_internal: Independent_blocks=%u", cctx->prefs.frameInfo.blockMode);
 
--- a/tests/frametest.c
+++ b/tests/frametest.c
@@ -728,6 +728,77 @@ static int unitTests(U32 seed, double compressibility)
         cctx = NULL;
     }
 
+    /* blockSizeID validation tests */
+    {   int const invalidBSID[] = { 1, 2, 3, 8, 9 };
+        int const validBSID[] = { 0 /* == default */, 4, 5, 6, 7 };
+        size_t const srcSize = 256 KB;
+        size_t nb;
+        LZ4F_preferences_t bsidPrefs;
+
+        assert(srcSize <= COMPRESSIBLE_NOISE_LENGTH);
+        CHECK( LZ4F_createCompressionContext(&cctx, LZ4F_VERSION) );
+
+        DISPLAYLEVEL(3, "LZ4F_compressBegin rejects invalid blockSizeID : ");
+        for (nb = 0; nb < sizeof(invalidBSID) / sizeof(invalidBSID[0]); nb++) {
+            memset(&bsidPrefs, 0, sizeof(bsidPrefs));
+            bsidPrefs.frameInfo.blockSizeID = (LZ4F_blockSizeID_t)invalidBSID[nb];
+            {   size_t const r = LZ4F_compressBegin(cctx, compressedBuffer, cBuffSize, &bsidPrefs);
+                if (!LZ4F_isError(r)) {
+                    DISPLAYLEVEL(3, "blockSizeID %i not rejected ! \n", invalidBSID[nb]);
+                    goto _output_error;
+        }   }   }
+        DISPLAYLEVEL(3, "OK \n");
+
+        DISPLAYLEVEL(3, "valid blockSizeID still accepted : ");
+        for (nb = 0; nb < sizeof(validBSID) / sizeof(validBSID[0]); nb++) {
+            memset(&bsidPrefs, 0, sizeof(bsidPrefs));
+            bsidPrefs.frameInfo.blockSizeID = (LZ4F_blockSizeID_t)validBSID[nb];
+            CHECK( LZ4F_compressFrame(compressedBuffer, cBuffSize,
+                                      CNBuffer, srcSize, &bsidPrefs) );
+        }
+        DISPLAYLEVEL(3, "OK \n");
+
+        DISPLAYLEVEL(3, "failed LZ4F_compressBegin preserves cctx : ");
+        {   BYTE* const ostart = (BYTE*)compressedBuffer;
+            BYTE* op = ostart;
+            size_t cErr;
+
+            memset(&bsidPrefs, 0, sizeof(bsidPrefs));
+            bsidPrefs.frameInfo.blockSizeID = LZ4F_max64KB;
+            bsidPrefs.frameInfo.contentChecksumFlag = LZ4F_contentChecksumEnabled;
+            CHECK_V(cErr, LZ4F_compressBegin(cctx, op, cBuffSize, &bsidPrefs));
+            op += cErr;
+
+            memset(&bsidPrefs, 0, sizeof(bsidPrefs));   /* must differ by more than blockSizeID */
+            bsidPrefs.frameInfo.blockSizeID = (LZ4F_blockSizeID_t)3;   /* invalid */
+            cErr = LZ4F_compressBegin(cctx, op, cBuffSize, &bsidPrefs);
+            if (!LZ4F_isError(cErr)) goto _output_error;
+
+            /* resume the frame started above */
+            CHECK_V(cErr, LZ4F_compressUpdate(cctx, op, cBuffSize - (size_t)(op-ostart),
+                                              CNBuffer, srcSize, NULL));
+            op += cErr;
+            CHECK_V(cErr, LZ4F_compressEnd(cctx, op, cBuffSize - (size_t)(op-ostart), NULL));
+            op += cErr;
+
+            {   size_t iSize = (size_t)(op - ostart);
+                size_t decodedSize = COMPRESSIBLE_NOISE_LENGTH;
+                LZ4F_decompressionContext_t dctx;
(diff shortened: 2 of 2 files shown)


fallenmi (changes requested):
I found one remaining state-preservation bug.

Reproduction on exact head a06d8a1c1b083ccde31eb5a1c43af53f4788f9fe:

1. Start a frame with blockSizeID=4, linked blocks, and the content checksum enabled.
2. Call LZ4F_compressBegin again on the same context with blockSizeID=3 and otherwise-default preferences; it correctly returns ERROR_maxBlockSize_invalid.
3. Continue the first frame with LZ4F_compressUpdate and LZ4F_compressEnd, then decode it.

The decoder consumes all 1347 frame bytes and produces all 262144 source bytes, but returns hint=4 because the checksum advertised by the original header was never written. The new validation happens after cctx->prefs = *preferencesPtr (and after context-management mutations), so the failed begin still replaces the active frame preferences. The added preservation test misses this because its valid and invalid preferences differ only in blockSizeID.

For comparison, exact base 0774d05 accepts invalid ID 1 (the original RED); this head rejects all five invalid IDs, accepts all five valid IDs, and preserves a context when the remaining preferences match. The divergent-preferences case above still fails deterministically. Moving validation of the effective blockSizeID before the first cctx mutation makes that same oracle pass; the early-validation variant also compiles cleanly as C90 and C99. Exact-head frametest -T20s completed, the matching-preferences oracle passed 30/30, and diff check is clean.

Please validate the effective blockSizeID before assigning cctx->prefs or changing the compression context, and extend the regression test with materially different preferences (content checksum is sufficient).

Review disclosure: OpenAI Codex assisted this review; the commands and results above were reproduced locally against the exact commits stated.

jaypatrickhoward:
Thanks — reproduced exactly, including the byte counts. On `a06d8a1c` I get `consumed 1104/1104, produced 262144/262144, hint=4`. You're right about the cause: the validation ran after `cctx->prefs = *preferencesPtr` and after the context-management block, so a rejected `LZ4F_compressBegin()` still replaced the preferences of the frame in progress.

Fixed by validating the effective `blockSizeID` before any modification of `@cctx`, as you suggested. That also made the later check unnecessary, so the whole library change is now one line:

```c
RETURN_ERROR_IF(dstCapacity < maxFHSize, dstMaxSize_tooSmall);
if (preferencesPtr == NULL) preferencesPtr = &prefNull;
FORWARD_IF_ERROR( LZ4F_getBlockSize(preferencesPtr->frameInfo.blockSizeID) ); /* validate before modifying @cctx */
cctx->prefs = *preferencesPtr;
```

`LZ4F_getBlockSize()` maps `0` to the default internally, so this validates the effective ID. `cctx->maxBlockSize = LZ4F_getBlockSize(...)` further down is now untouched from `dev`.

Two notes from working through it:

**A related variant.** When the frame's `blockMode` differs as well — header written with `LZ4F_blockIndependent`, rejected preferences defaulting back to `LZ4F_blockLinked` — the result isn't just a missing checksum: the frame fails to decode at all (`ERROR_decompressionFailed`). Same root cause, same fix.

**On the test.** "Content checksum is sufficient" needed one more thing. With only `contentChecksumFlag` differing, the test still passed against the broken head: the missing checksum makes `LZ4F_decompress()` return a hint of `4`, not an error, and `CHECK()` only tests `LZ4F_isError()`. The real gap was that the test never asserted the frame was fully consumed, so I added `if (cErr != 0) goto _output_error;`. With that, it fails on `a06d8a1c` and passes with the fix.

Still out of scope, but worth flagging: `LZ4F_compressBegin()` remains non-atomic on failure for the `allocation_failed` and `parameter_invalid` returns, which also sit after the preferences assignment. Those require OOM or a >2 GB dictionary to reach. Making the function fully atomic on failure looks like a larger change than belongs here.

I also dropped a `LZ4F_compressFrame` sub-test I had added. Mutation testing showed it cannot fail unless the rejection sub-test above it fails first, since `LZ4F_compressFrame()` routes through `LZ4F_compressBegin()` — so it added 24 lines and no detection. The two commits are squashed and the PR is now +72/-0.

Verified on the new head: `frametest -T30s` and `fuzzer -T30s` on a clean build; C90 and C99 with `-Werror -Wpedantic`; `-Wc++-compat -Wall -Wextra -Werror` on `lz4frame.c`.

fallenmi (approved):
Revalidated exact head `1dc669d45adb774ad5e4ccc8f754358d63ba4a42` against base `0774d05537f9762f838f7ab541b7765f1a729cb5` and the previously reviewed `a06d8a1c1b083ccde31eb5a1c43af53f4788f9fe`.

The state-preservation issue is fixed. With an active 64 KB linked, content-checksummed frame, a rejected second begin using `blockSizeID=3` and otherwise-default preferences now leaves the context intact: update/end/decode consumes all 6,580 frame bytes, reproduces all 262,144 source bytes, and returns `hint=0`. The same oracle on `a06d8a1` consumes the full frame and reproduces the data but returns `hint=4`; base still accepts the invalid ID.

Invalid IDs 1, 2, 3, 8, and 9 are rejected on this head, valid IDs 0 and 4 through 7 remain accepted, and the matching-preferences preservation case passed 30/30. The new regression test fails against the old implementation and passes here.

Also passed locally: the official `c_standards_c90` and `c_standards_c99` targets, `frametest -T20s`, the external oracle under ASan+UBSan, and diff check.

The two live reds are baseline/toolchain failures, not introduced by this diff: the previous `a06d8a1` runs fail identically because `windows-latest` contains VS18 while the unchanged workflow requests Visual Studio 17 2022, and macOS 26 rejects the unchanged `examples/frameCompress.c:446` integer-to-double conversion under `-Werror`. The other 53 checks pass.

No remaining blocker from my review.

Review disclosure: OpenAI Codex assisted this rereview; I reproduced the commands and results locally against the exact commits stated.

PR #1795, opened 2026-08-25
examples: cast division denominators to double (fixes macOS -Werror build)

Alb3e3:
The `macOS` job has been red on every PR for a while. It is not caused by the PRs it fails on — it is a pre-existing break in `examples/`, surfaced by newer clang.

## What breaks

The `Run comprehensive tests` step builds with:

```
CFLAGS="-O3 -Werror -Wconversion -Wno-sign-conversion" make -j test V=1
```

clang's `-Wconversion` implies `-Wimplicit-int-float-conversion`, and `examples/Makefile` has `test : all`, so the examples get built with those flags. In these ratio expressions the numerator is cast explicitly but the denominator is left to convert implicitly:

```
frameCompress.c:446:40: error: implicit conversion from 'const unsigned long long'
    to 'double' may lose precision [-Werror,-Wimplicit-int-float-conversion]
  446 |             (double)ret.size_out / ret.size_in * 100);
      |                                  ~ ~~~~^~~~~~~
```

**`frameCompress.c` is only the first failure.** `make` stops there, which is why the CI log shows nothing else — but `bench_functions.c` has **ten more** instances of the same pattern queued behind it, so fixing just the one line moves the error rather than clearing it. I found this the hard way after proposing the one-liner on #1785/#1786.

## The change

Cast the denominators too. This matches how `fileCompress.c:193` already writes the same computation:

```c
(double)get_file_size(lz4Filename) / (double)get_file_size(inpFilename) * 100);
```

The conversion happened anyway through the usual arithmetic conversions, so this is a no-op at runtime — purely making the intent explicit so the diagnostic is satisfied.

## Verification

Reproduced and fixed locally with `clang-18`, which emits the identical diagnostic (same file, line, column and wording) as the Xcode 26.6 runner.

Running the two macOS CI commands verbatim, both now exit 0 with **zero errors and zero warnings**:

```
CFLAGS="-Werror -O0" make default V=1
CFLAGS="-O3 -Werror -Wconversion -Wno-sign-conversion" make all V=1
```

Behaviour is unchanged — `frameCompress` output is byte-identical before and after:

```
/tmp/lz4in.txt: 200000 → 106645 bytes, 53.3%
```

## Not covered by this PR

The `Core Tests (windows-latest, Release)` job is also red, but for an unrelated reason — it is a runner-image problem, not a source one:

```
CMake Error at CMakeLists.txt:35 (project):
  Generator
    Visual Studio 17 2022
  could not find any instance of Visual Studio.
```

That needs a workflow/runner change, so I left it alone.

## AI disclosure

Prepared with AI assistance, then reviewed and verified locally.


Alb3e3:
CI confirms it: **`macOS` is green on this branch** — https://github.com/lz4/lz4/actions/runs/32903824703/job/97983335892 — so the real Xcode runner agrees with the local `clang-18` reproduction. 49 checks pass.

The one red check is the `Core Tests (windows-latest, Release)` job called out in the description, unchanged and unrelated:

```
CMake Error at CMakeLists.txt:35 (project):
  Generator
    Visual Studio 17 2022
  could not find any instance of Visual Studio.
```

That is a runner-image problem and needs a workflow change, not a source one. Happy to open a separate PR for it if you want the generator relaxed (e.g. letting CMake pick the default generator, or bumping to the VS version the current `windows-latest` image ships) — just say which direction you prefer.

Alb3e3:
Quick status on this one, since it affects more than just this PR.

`Core Tests (macos-latest, Release)` passes here. It is currently failing on every other
open PR that runs the full matrix:

| PR | author | macOS |
| --- | --- | --- |
| #1795 (this) | Alb3e3 | pass |
| #1785 | Alb3e3 | fail |
| #1786 | Alb3e3 | fail |
| #1788 | jaypatrickhoward | fail |
| #1789 | jaypatrickhoward | fail |
| #1790 | jaypatrickhoward | fail |
| #1791 | wittkung | fail |

(#1787 is docs-only, so it only triggers 2 checks and never runs the matrix.)

The cause is the implicit `uint64_t` to `double` conversion in the `examples/` division
denominators, which is an error under `-Werror` on the current macOS runner.
`examples/Makefile` has `test : all`, so the macOS job stops there no matter what the PR
itself changes. That is why it reproduces on unrelated PRs from other contributors.

One thing this PR does **not** fix: `Core Tests (windows-latest, Release)` still fails here,
and on all of the PRs above. That one is `Generator "Visual Studio 17 2022" could not find
any instance of Visual Studio`, which is a runner-image problem needing a workflow change
rather than a source change. Happy to take that as a separate PR if it would help.

PR #1798, opened 2026-09-09
cli: keep source after decompression failure

Furan666:
`--rm` is documented to remove a source after successful decompression. A recoverable trailing-data error currently sets a nonzero result and then reaches the unconditional source-removal block.

For example, an empty skippable frame followed by an unknown four-byte magic exits with status 1 while deleting the input. Gate source removal on the decompression result and add a regression that verifies the malformed input remains unchanged. Successful decompression continues to remove its source.

Validation:

- strict C90/Werror build
- baseline regression fails because the source disappears; patched regression passes
- `test-lz4-basic` under the strict build and ASan/UBSan

This change is scoped to decoder failures represented by `LZ4IO_decompressSrcFile()`'s result. Output-stream close handling is separate and unchanged.

### DIFF
--- a/programs/lz4io.c
+++ b/programs/lz4io.c
@@ -2437,7 +2437,7 @@ LZ4IO_decompressSrcFile(unsigned long long* outGenSize,
 
     /* Close input */
     fclose(finput);
-    if (prefs->removeSrcFile) {  /* --rm */
+    if (!result && prefs->removeSrcFile) {  /* --rm */
         if (remove(input_filename))
             END_PROCESS(45, "Remove error : %s: %s", input_filename, strerror(errno));
     }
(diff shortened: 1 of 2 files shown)


Furan666:
CI note for reviewers: 53 checks pass. The two reported failures are outside this two-file change:

- `Core Tests (windows-latest, Release)` stops during CMake configuration because the runner has Visual Studio 18 installed while the workflow requests the Visual Studio 17 2022 generator. The same runner failure occurred on unrelated PR #1765: https://github.com/lz4/lz4/actions/runs/28361547968/job/84017164798
- `macOS` completes the LZ4 tests, then fails while compiling unchanged `examples/frameCompress.c:446` under Xcode 26.6 with `-Werror,-Wimplicit-int-float-conversion`: https://github.com/lz4/lz4/actions/runs/34345541751/job/102446147309

The changed files are only `programs/lz4io.c` and `tests/test-lz4-basic.sh`.

